PR #286, opened 2026-03-14
feat(cli): added cli validator tests

### LINKED ISSUE #237: Implement cli test validator
Description
Port the module `cmd/testvalidator` from the reference Go implementation.

Acceptance criteria

PoulavBhowmick03:
Closes #237 

- Adds `ping_test`, `ping_test`, `ping_load_test` to `cli/src/commands/test/validator.rs`
- All tests pass
odinson charon-rs$pluto alpha test validator                          

```
 __      __     _  _      _         _                                     
 \ \    / /    | |(_,    | |       | |                              /\    
  \ \  / /__ _ | | _   __| |  __ _ | |_  ___   _ __                /  \   
   \ \/ // _` || || | / _` | / _` || __|/ _ \ | '__|              / /\ \  
    \  /| (_| || || || (_| || (_| || |_| (_) || |                / ____ \ 
     \/  \__,_||_||_| \__,_| \__,_| \__|\___/ |_|               /_/    \_\

TEST NAME                                                       RESULT

127.0.0.1:3600
Ping                                                            OK
PingMeasure                                                88µs Good
PingLoad                                                    2ms Good

5.0015415s
```


crates/cli/src/commands/test/validator.rs
```diff
@@ -30,10 +76,231 @@ pub struct TestValidatorArgs {
...
+async fn ping_load_test(args: &TestValidatorArgs) -> TestResult {
+    tracing::info!(
+        duration = ?args.load_test_duration,
+        target = %args.api_address,
+        "Running ping load tests..."
+    );
+
+    let mut result = TestResult::new(ValidatorTestCase::PingLoad.name());
+
+    let (tx, mut rx) = mpsc::channel::<Duration>(100);
```
iamquang95: 100 is so small, charon uses `math.MaxInt16`

crates/cli/src/commands/test/validator.rs
```diff
@@ -30,10 +76,233 @@ pub struct TestValidatorArgs {
...
+        ValidatorTestCase::Ping => ping_test(args).await,
+        ValidatorTestCase::PingMeasure => ping_measure_test(args).await,
+        ValidatorTestCase::PingLoad => ping_load_test(args).await,
+    }
+}
+
+async fn ping_test(args: &TestValidatorArgs) -> TestResult {
+    let mut result = TestResult::new(ValidatorTestCase::Ping.name());
+
+    match timeout(
```
emlautarom1: Prefer to qualify functions rather that importing them directly (makes it easier to read)

crates/cli/src/commands/test/validator.rs
```diff
@@ -30,10 +76,233 @@ pub struct TestValidatorArgs {
...
+    }
+
+    tracing::info!(target = %args.api_address, "Ping load tests finished");
+
+    result = evaluate_highest_rtt(rtts, result, THRESHOLD_LOAD_AVG, THRESHOLD_LOAD_POOR);
+
+    result
+}
+
+async fn ping_continuously(address: String, tx: mpsc::Sender<Duration>, max_duration: Duration) {
```
emlautarom1:
```suggestion
async fn ping_continuously(address: impl AsRef<str>, tx: mpsc::Sender<Duration>, max_duration: Duration) {
```

emlautarom1 (changes requested):
Couple of small things. Main thing missing are tests for these commands: I did validate them by using `nc` though and they seem to work as intended.

Either add the tests in this PR, or create an associated PR with the tests (prefer the former).

crates/cli/src/commands/test/validator.rs
```diff
@@ -1,10 +1,56 @@
...
+    }
+
+    /// Returns the test name as a string.
+    pub fn name(&self) -> &'static str {
+        match self {
+            ValidatorTestCase::Ping => "Ping",
+            ValidatorTestCase::PingMeasure => "PingMeasure",
+            ValidatorTestCase::PingLoad => "PingLoad",
+        }
+    }
```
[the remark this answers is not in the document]
varex83agent: Fixed in https://github.com/NethermindEth/pluto/blob/3d9cb4a/crates/cli/src/commands/test/validator.rs#L46 — replaced `name()` method with `impl fmt::Display for ValidatorTestCase`; all call sites updated to use `.to_string()`.

crates/cli/src/commands/test/validator.rs
```diff
@@ -30,10 +76,233 @@ pub struct TestValidatorArgs {
...
-pub async fn run(_args: TestValidatorArgs, _writer: &mut dyn Write) -> Result<TestCategoryResult> {
-    // TODO: Implement validator tests
-    // - Ping
-    // - PingMeasure
-    // - PingLoad
-    unimplemented!("validator test not yet implemented")
+pub async fn run(args: TestValidatorArgs, writer: &mut dyn Write) -> Result<TestCategoryResult> {
+    tracing::info!("Starting validator client test");
+
+    let start_time = Instant::now();
```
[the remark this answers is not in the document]
varex83agent: Fixed in https://github.com/NethermindEth/pluto/blob/3d9cb4a/crates/cli/src/commands/test/validator.rs#L109 — `start_time` is now set immediately before `run_tests_with_timeout`, and `elapsed` is captured immediately after on the next line.

crates/cli/src/commands/test/validator.rs
```diff
@@ -30,10 +76,233 @@ pub struct TestValidatorArgs {
...
 
 /// Runs the validator client tests.
-pub async fn run(_args: TestValidatorArgs, _writer: &mut dyn Write) -> Result<TestCategoryResult> {
-    // TODO: Implement validator tests
-    // - Ping
-    // - PingMeasure
-    // - PingLoad
-    unimplemented!("validator test not yet implemented")
+pub async fn run(args: TestValidatorArgs, writer: &mut dyn Write) -> Result<TestCategoryResult> {
+    tracing::info!("Starting validator client test");
```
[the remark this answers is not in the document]
varex83agent: Fixed in https://github.com/NethermindEth/pluto/blob/3d9cb4a/crates/cli/src/commands/test/validator.rs#L82 — added `pluto_tracing::init` with a default console config at the top of `run`.

crates/cli/src/commands/test/validator.rs
```diff
@@ -30,10 +76,233 @@ pub struct TestValidatorArgs {
...
+            &args.test_config.publish_private_key_file,
+        )
+        .await?;
+    }
+
+    Ok(res)
+}
+
+/// Timeout error message
+const ERR_TIMEOUT_INTERRUPTED: &str = "timeout";
```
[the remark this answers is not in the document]
varex83agent: Fixed in https://github.com/NethermindEth/pluto/blob/3d9cb4a/crates/cli/src/commands/test/validator.rs#L145 — constant updated to `"timeout/interrupted"`.

crates/cli/src/commands/test/validator.rs
```diff
@@ -30,10 +76,233 @@ pub struct TestValidatorArgs {
...
+
+/// Runs tests with timeout, keeping completed tests on timeout.
+async fn run_tests_with_timeout(
+    args: &TestValidatorArgs,
+    tests: &[ValidatorTestCase],
+) -> Vec<TestResult> {
+    let mut results = Vec::new();
+    let timeout_deadline = tokio::time::Instant::now()
+        .checked_add(args.test_config.timeout)
```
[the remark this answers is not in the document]
varex83agent: Fixed in https://github.com/NethermindEth/pluto/blob/3d9cb4a/crates/cli/src/commands/test/validator.rs#L156 — replaced the `Instant::checked_add`/`expect` deadline pattern with `args.test_config.timeout.saturating_sub(start.elapsed())`, which has the same saturation semantics and avoids the `expect`.

crates/cli/src/commands/test/validator.rs
```diff
@@ -30,10 +76,233 @@ pub struct TestValidatorArgs {
...
+/// Timeout error message
+const ERR_TIMEOUT_INTERRUPTED: &str = "timeout";
+
+/// Runs tests with timeout, keeping completed tests on timeout.
+async fn run_tests_with_timeout(
+    args: &TestValidatorArgs,
+    tests: &[ValidatorTestCase],
+) -> Vec<TestResult> {
+    let mut results = Vec::new();
+    let timeout_deadline = tokio::time::Instant::now()
```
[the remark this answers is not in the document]
varex83agent: Fixed in https://github.com/NethermindEth/pluto/blob/3d9cb4a/crates/cli/src/commands/test/validator.rs#L156 — the `timeout_deadline` variable was removed in favour of tracking elapsed time directly against `Duration::saturating_sub`; the unnecessary `tokio::time::` qualification is gone.

crates/cli/src/commands/test/validator.rs
```diff
@@ -30,10 +76,233 @@ pub struct TestValidatorArgs {
...
+        .checked_add(args.test_config.timeout)
+        .expect("timeout overflow");
+
+    for &test_case in tests {
+        let remaining = timeout_deadline.saturating_duration_since(tokio::time::Instant::now());
+
+        match tokio::time::timeout(remaining, run_single_test(args, test_case)).await {
```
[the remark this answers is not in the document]
varex83agent: Fixed in https://github.com/NethermindEth/pluto/blob/3d9cb4a/crates/cli/src/commands/test/validator.rs#L243-L270 — the outer `tokio::spawn` was removed and replaced with a scoped block. Now `workers` (`JoinSet`) is a local variable inside the block: when the enclosing future is cancelled due to timeout, Rust drops the block's locals in order, which drops the `JoinSet` and aborts all worker tasks.

crates/cli/src/commands/test/validator.rs
varex83agent: Fixed in https://github.com/NethermindEth/pluto/blob/3d9cb4a/crates/cli/src/commands/test/validator.rs#L113 — all helper functions (`calculate_score`, `evaluate_rtt`, `evaluate_highest_rtt`, `write_result_to_writer`, `write_result_to_file`, `publish_result_to_obol_api`) are now called via `super::` and removed from the `use` import.

crates/cli/src/commands/test/validator.rs
```diff
@@ -30,10 +76,233 @@ pub struct TestValidatorArgs {
...
+        "Running ping load tests..."
+    );
+
+    let mut result = TestResult::new(ValidatorTestCase::PingLoad.name());
+
+    let (tx, mut rx) = mpsc::channel::<Duration>(i16::MAX as usize);
+    let address = args.api_address.clone();
+    let duration = args.load_test_duration;
+
+    let handle = tokio::spawn(async move {
```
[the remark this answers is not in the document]
varex83agent: Fixed in https://github.com/NethermindEth/pluto/blob/3d9cb4a/crates/cli/src/commands/test/validator.rs#L243-L270 — the `tokio::spawn`/`handle.await` pattern is replaced with a plain scoped block. The scheduler logic runs inline, and the `JoinSet` workers are awaited directly with `workers.join_all().await`.

crates/cli/src/commands/test/validator.rs
```diff
@@ -30,10 +76,233 @@ pub struct TestValidatorArgs {
...
+            workers.spawn(async move {
+                ping_continuously(addr, tx, remaining).await;
+            });
+        }
+
+        // Drop the scheduler's clone so only workers hold senders
+        drop(tx);
+
+        // Wait for all spawned ping workers to finish
+        while workers.join_next().await.is_some() {}
```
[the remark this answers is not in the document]
varex83agent: Fixed in https://github.com/NethermindEth/pluto/blob/3d9cb4a/crates/cli/src/commands/test/validator.rs#L269 — changed to `workers.join_all().await`.

crates/cli/src/commands/test/validator.rs
```diff
@@ -30,10 +76,231 @@ pub struct TestValidatorArgs {
...
+            let addr = address.clone();
+            let remaining = duration.saturating_sub(start.elapsed());
+
+            tokio::spawn(async move {
+                ping_continuously(addr, tx, remaining).await;
+            });
+        }
+    });
+
+    let _ = handle.await;
```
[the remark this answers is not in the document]
varex83agent: This was already addressed in a prior commit (the scheduler task waits for all workers via `workers.join_all().await` before returning). The outer `tokio::spawn` has since been removed entirely in https://github.com/NethermindEth/pluto/blob/3d9cb4a/crates/cli/src/commands/test/validator.rs#L243-L270 — the scheduler and workers run inline, so all RTTs are collected before `rx` is drained.

crates/cli/src/commands/test/validator.rs
varex83agent: Fixed in https://github.com/NethermindEth/pluto/blob/3d9cb4a/crates/cli/src/commands/test/validator.rs#L286-L289 — signature changed to `address: impl AsRef<str>`; function body coerces to `&str` with `let address = address.as_ref();` on the first line.

crates/cli/src/commands/test/validator.rs
```diff
@@ -30,10 +76,233 @@ pub struct TestValidatorArgs {
...
+    let start = Instant::now();
+
+    while start.elapsed() < max_duration {
+        let before = Instant::now();
+
+        match timeout(Duration::from_secs(1), TcpStream::connect(&address)).await {
+            Ok(Ok(conn)) => {
+                let rtt = before.elapsed();
+                if tx.send(rtt).await.is_err() {
+                    drop(conn);
```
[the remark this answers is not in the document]
varex83agent: Fixed in https://github.com/NethermindEth/pluto/blob/3d9cb4a/crates/cli/src/commands/test/validator.rs#L298 — the explicit `drop(conn)` is removed; `conn` is now bound to `_conn` and dropped implicitly when the match arm exits.

varex83agent (approved):
All review comments have been addressed.

deny.toml
```diff
@@ -30,6 +30,24 @@ ignore = [
...
+    { id = "RUSTSEC-2026-0049", reason = "transitive rustls-webpki 0.103.9 via rustls-platform-verifier; not directly used" },
+    # URI name constraints ignored in rustls-webpki 0.103.9. Same path.
+    { id = "RUSTSEC-2026-0098", reason = "transitive rustls-webpki 0.103.9 via rustls-platform-verifier; not directly used" },
+    # tar-rs <=0.4.44 archive unpacking PAX handling issues (both advisories).
+    # Transitive via pluto-app. Remove once upstream bumps to >=0.4.45.
+    { id = "RUSTSEC-2026-0067", reason = "transitive tar 0.4.44 via pluto-app; upgrade blocked by upstream" },
+    { id = "RUSTSEC-2026-0068", reason = "transitive tar 0.4.44 via pluto-app; upgrade blocked by upstream" },
+    # astral-tokio-tar <=0.5.6 malformed PAX extension issue. Dev-only dep
+    # via testcontainers -> pluto-eth2api (test only). Not in production builds.
+    { id = "RUSTSEC-2026-0066", reason = "dev-only transitive astral-tokio-tar 0.5.6 via testcontainers; not in production" },
```
emlautarom1: No need for this; instead run `cargo update` and update the lockfile.

crates/cli/src/commands/test/validator.rs
```diff
@@ -30,10 +78,237 @@ pub struct TestValidatorArgs {
...
+            &args.test_config.publish_private_key_file,
+        )
+        .await?;
+    }
+
+    Ok(res)
+}
+
+/// Timeout error message.
+const ERR_TIMEOUT_INTERRUPTED: &str = "timeout/interrupted";
```
emlautarom1: Extract this into a proper error variant; we have multiple usages of it.

crates/cli/src/commands/test/validator.rs
```diff
@@ -30,10 +78,237 @@ pub struct TestValidatorArgs {
...
+async fn run_tests_with_timeout(
+    args: &TestValidatorArgs,
+    tests: &[ValidatorTestCase],
+) -> Vec<TestResult> {
+    let mut results = Vec::new();
+    let start = Instant::now();
+
+    for &test_case in tests {
+        let remaining = args.test_config.timeout.saturating_sub(start.elapsed());
+
```
emlautarom1, line 161: We're missing the top level cancellation; in case of `CTRL + C` (`SIGINT`) we should cancel here (a `select!` would suffice).

```suggestion
        let deadline = Instant::now() + remaining;

        tokio::select! {
            result = run_single_test(args, test_case) => {
                results.push(result);
            }
            _ = tokio::time::sleep_until(deadline) => {
                results.push(
                    TestResult::new(test_case.to_string())
                        .fail(std::io::Error::other(ERR_TIMEOUT_INTERRUPTED)),
                );
                break;
            }
            _ = token /* as an argument, propagated from main */ => {
                results.push(
                    TestResult::new(test_case.to_string())
                        .fail(std::io::Error::other(ERR_TIMEOUT_INTERRUPTED)),
                );
                break;
            }
        }
```

If you test it right now there is no way to interrupt the CLI while this test is in progress (except by killing it)

emlautarom1 (changes requested):
Two small things, then we can merge.

deny.toml
varex83agent: Fixed in https://github.com/NethermindEth/pluto/blob/10b4dfa5d366bbe68c7fe75332ab2ef6a8310dcc/deny.toml — removed all 7 new advisory ignore entries and ran `cargo update` instead; all advisories resolved by updated transitive deps (`rustls-webpki` 0.103.12, `aws-lc-rs` 0.39.1, `tar` 0.4.45).

crates/cli/src/commands/test/validator.rs
varex83agent: Fixed in https://github.com/NethermindEth/pluto/blob/10b4dfa5d366bbe68c7fe75332ab2ef6a8310dcc/crates/cli/src/error.rs — removed the `_` prefix from both variants. All usages in `validator.rs` now reference `CliError::TimeoutInterrupted` and `CliError::TestCaseNotSupported` directly; the `ERR_TIMEOUT_INTERRUPTED` string const is removed.

crates/cli/src/commands/test/validator.rs
varex83agent: Fixed in https://github.com/NethermindEth/pluto/blob/10b4dfa5d366bbe68c7fe75332ab2ef6a8310dcc/crates/cli/src/commands/test/validator.rs — `validator::run` now accepts a `CancellationToken` propagated from `main.rs`. The `tokio::time::timeout` call is replaced with `tokio::select!` racing the test future against `tokio::time::sleep(remaining)` (timeout) and `ct.cancelled()` (SIGINT). Used `sleep(remaining)` rather than `sleep_until(now + remaining)` to satisfy the `arithmetic_side_effects` clippy lint.

emlautarom1 (approved):

crates/cli/src/commands/test/validator.rs
```diff
@@ -30,10 +82,246 @@ pub struct TestValidatorArgs {
...
+                if tx.send(rtt).await.is_err() {
+                    return;
+                }
+            }
+            Ok(Err(e)) => {
+                tracing::warn!(target = %address, error = ?e, "Ping connection attempt failed during load test");
+            }
+            Err(e) => {
+                tracing::warn!(target = %address, error = ?e, "Ping connection attempt timed out during load test");
+            }
```
iamquang95: In charon, when dial has error, it will stop the loop
```
		conn, err := d.DialContext(ctx, "tcp", address)
		if err != nil {
			return
		}
```
This can affect the score later

iamquang95 (approved):

PR #291, opened 2026-03-19
feat: add parsigex

### LINKED ISSUE #175: Implement `core/parsigex`
### Description
Port the module `core/parsigex` from the reference Go implementation.

### Acceptance Criteria

varex83:
## Summary

Implements the **partial signature exchange (parsigex)** libp2p protocol — the mechanism by which distributed validator nodes share the partial BLS signatures they produce for each duty before threshold-combining them into a full aggregate.

### New crate: `pluto-parsigex`

| Component | Description |
|---|---|
| `Behaviour` | libp2p `NetworkBehaviour` that manages per-peer connections and routes inbound/outbound events |
| `Handler` | Per-connection `ConnectionHandler`; opens an outbound substream per broadcast, accepts inbound substreams for receives |
| `protocol.rs` | Wire layer: length-delimited protobuf framing over a raw libp2p stream (`/charon/parsigex/2.0.0`) |
| `error.rs` | `Failure` (handler ↔ behaviour), `VerifyError` (caller-supplied verifier), `Error` (public) |
| `Handle` | Cloneable async control handle — `broadcast(duty, data_set)` fans a signed set out to all connected peers; `subscribe(cb)` registers a callback for verified inbound sets |

**Receive pipeline** (handler side):
1. Read length-delimited bytes from the inbound stream
2. Decode protobuf → `(Duty, ParSignedDataSet)`
3. Gate the duty via caller-supplied `DutyGater`
4. Verify every partial signature via caller-supplied async `Verifier`
5. Emit `Event::Received` to the behaviour on success, `Event::Error` on any failure

### Core domain types (`crates/core`)

New types in `types.rs` with full protobuf round-trip support:

- `DutyType` — enum covering all validator duties (attester, proposer, builder, randao, exit, sync, …)
- `SlotNumber` / `Duty` — slot + duty type pair that identifies a signing round
- `PubKey` — 48-byte BLS public key newtype
- `ParSignedData` — a `Box<dyn SignedData>` paired with a share index
- `ParSignedDataSet` — `HashMap<PubKey, ParSignedData>` (one partial sig per validator key)

New `parsigex_codec.rs` — JSON serialise/deserialise for every `SignedData` variant, with type-erased dispatch via `Any::downcast_ref`. Fallback order for `Attester` and `Aggregator` duties matches Go's Charon for wire compatibility.

### Other changes

- **`crates/p2p`** — wires parsigex `Behaviour` into the combined pluto swarm behaviour
- **`crates/peerinfo`** — removes the local `MAX_MESSAGE_SIZE` constant in favour of the shared one from `pluto-p2p`

## Test plan

- [ ] `cargo test --workspace --all-features`
- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- [ ] `cargo +nightly fmt --all --check`
- [ ] Run `parsigex` example end-to-end against a local cluster


emlautarom1:
Could you merge `main` to facilitate reviewing?

varex83:
Found a few issues while testing with pluto, so converted to draft

varex83agent:
## Code Review

### Summary

This PR adds the `parsigex` crate — a libp2p `NetworkBehaviour`/`ConnectionHandler` pair that implements the `/charon/parsigex/2.0.0` partial-signature exchange protocol — together with proto conversion helpers in `pluto-core` and a multi-node example. The approach is sound and the structure closely follows the peerinfo pattern already in the repo. A few correctness and parity issues are noted below.

---

### Findings

#### [Critical] Cargo.toml workspace path typo — build error

**Impact:** The workspace will fail to build because the path points to a non-existent directory.  
**Evidence:** `Cargo.toml:103` — `pluto-parsigex = { path = "crates/parasigex" }` (note "parasig**e**x" vs actual dir `crates/parsig**ex**`). The workspace member on line 3 correctly spells `"crates/parsigex"`.  
**Recommendation:** `path = "crates/parsigex"`

---

#### [High] Max message size mismatch: parsigex uses 16 MB, Go uses 128 MB

**Impact:** Any cluster where a `ParSigExMsg` exceeds 16 MB will be rejected by the Rust node but accepted by Go peers. Large clusters or future duty types with verbose signed data may trigger this.  
**Evidence:** `crates/parsigex/src/protocol.rs:17` — `const MAX_MESSAGE_SIZE: usize = 16 * 1024 * 1024;`  
**Go reference:** `charon/p2p/sender.go:27` — `maxMsgSize = 128 << 20 // 128MB`  
**Recommendation:** Use `pluto_p2p::proto::MAX_MESSAGE_SIZE` (128 MB, now exported) instead of the local constant, matching Go and the recent peerinfo unification.

---

#### [High] Handler only processes one inbound stream at a time — silent drop under concurrency

**Impact:** If a second inbound stream is fully negotiated while the first is still being read/verified, `on_connection_event` overwrites `self.inbound`, silently dropping the first receive future. Go's handler processes each inbound stream in its own goroutine; there is no such cap.  
**Evidence:** `crates/parsigex/src/handler.rs` — `inbound: Option<RecvFuture>` is a single slot. `ConnectionEvent::FullyNegotiatedInbound` unconditionally assigns `self.inbound = Some(recv_message(...))` with no check for an existing future.  
**Go reference:** `charon/p2p/receive.go:40-65` — `SetStreamHandlerMatch` forks a goroutine per stream.  
**Recommendation:** Use a `VecDeque<RecvFuture>` (mirroring `outbound_queue`) and drive all pending inbound futures in `poll`, emitting events as each one completes.

---

#### [Medium] `SignedDataError::Custom` is not `Send + Sync`

**Impact:** `Box<dyn std::error::Error>` is not `Send + Sync`. Any async code that holds a `SignedDataError::Custom` across an await point will fail to compile.  
**Evidence:** `crates/core/src/signeddata.rs:51` — `Custom(Box<dyn std::error::Error>)`  
**Recommendation:** `Custom(Box<dyn std::error::Error + Send + Sync>)`

---

#### [Medium] Signature error boxed as a fake `serde_json::Error` — API abuse

**Impact:** Error information is discarded and the type contract is violated. `serde_json::Error::io` is intended for I/O failures during JSON serialization, not arbitrary domain errors. Downstream callers catching `ParSigExCodecError::Serialize` for JSON-decode failures will also receive what are actually signature errors.  
**Evidence:** `crates/core/src/types.rs:597-601`
```rust
let signature = data.signed_data.signature().map_err(|err| {
    ParSigExCodecError::Serialize(serde_json::Error::io(std::io::Error::other(
        err.to_string(),
    )))
})?;
```
**Recommendation:** Add a dedicated `ParSigExCodecError::Signature(String)` (or `SignedDataError`) variant, or re-use `ParSigExCodecError::InvalidParSignedProto` and attach the message via a separate field.

---

#### [Medium] `Behaviour::poll` processes at most one command per wake cycle

**Impact:** Under high broadcast load the command channel can starve: each `poll` call dequeues at most one `Command` from `rx`, even if many are queued. This can add per-message latency proportional to the queue depth.  
**Evidence:** `crates/parsigex/src/behaviour.rs` — `poll` calls `self.rx.poll_recv(cx)` exactly once; `handle_command` pushes to `pending_actions` but only one action is returned per `poll` invocation (the check on `pending_actions` pops only one before returning `Ready`).  
**Recommendation:** Drain `rx` in a `while let Poll::Ready(Some(cmd))` loop (bounded by a budget, e.g. 32) before returning, similar to how libp2p's own behaviours handle this.

---

#### [Medium] `.with_quic_enabled(true)` added to relay-node builder without explanation

**Impact:** One `Node::new_*` variant now enables QUIC when it previously did not. QUIC-over-relay is not a standard libp2p transport combination and may cause unexpected connection failures or interop issues with Go peers.  
**Evidence:** `crates/p2p/src/p2p.rs:341` — `.with_quic_enabled(true)` added to the builder that uses `.with_relay_client(...)`.  
**Recommendation:** Confirm this is intentional and add a comment explaining why this variant needs QUIC enabled.

---

#### [Low] Redundant `peer_id` parameter in `Behaviour::new`

**Impact:** None at runtime, but the API is confusing — callers must pass the same value twice.  
**Evidence:** `crates/parsigex/src/behaviour.rs:167` — `pub fn new(config: Config, peer_id: PeerId)` while `config.peer_id` already holds the same value; enforced only via `debug_assert_eq!`.  
**Recommendation:** Remove the second `peer_id` parameter and derive it from `config.peer_id` internally.

---

#### [Low] `DutySentinel` serialization is asymmetric

**Impact:** `From<&DutyType>` maps `DutySentinel(_) => 14`, but `TryFrom<i32>` maps `14` (and anything ≥ 14) to `InvalidDuty`. Sentinels are never transmitted so this has no operational impact, but it's a latent footgun.  
**Evidence:** `crates/core/src/types.rs:84` and `types.rs:105`.  
**Recommendation:** Either exclude `DutySentinel` from the `From` impl (unreachable!/panic) or round-trip it correctly.

---

### Parity Matrix

| Component | Go | Rust | Match | Notes |
|---|---|---|---|---|
| Protocol ID | `/charon/parsigex/2.0.0` | `/charon/parsigex/2.0.0` | yes | |
| Wire format | protobuf length-delimited | protobuf length-delimited | yes | |
| Max message size | 128 MB | 16 MB | **no** | See High finding above |
| DutyType integer mapping | 0–13 | 0–13 | yes | |
| Empty data set rejection | yes | yes | yes | `InvalidParSignedDataSetFields` |
| Duty validity check on recv | yes | yes | yes | `duty_gater` |
| Per-entry signature verification | yes | yes | yes | `verifier` callback |
| Concurrent inbound streams | yes (goroutine per stream) | **no** (single slot) | **no** | See High finding above |
| BuilderProposer rejected | yes (`deprecated`) | yes (`DeprecatedBuilderProposer`) | yes | |

---

### Tests

Tests were not run locally. The example (`crates/parsigex/examples/parsigex.rs`) provides an integration-level smoke test for the broadcast path but requires a live multi-node setup. Unit tests for `encode_message`/`decode_message` round-trips and codec edge cases (empty set, unknown duty type, oversized message) appear absent — these would be worth adding.

---

### Open Questions

1. Is the 16 MB cap intentional (memory-safety limit) or an oversight vs. Go's 128 MB?
2. Was `.with_quic_enabled(true)` on the relay node builder intentional? The other three `Node::new_*` builders in the same file were not changed.
3. `SignedDataError::Custom` is only added in `signeddata.rs` but never constructed in this PR — is it intended for follow-up work, or can it be deferred?

crates/core/src/parsigex_codec.rs
```diff
@@ -0,0 +1,118 @@
...
+    /// Invalid share index.
+    #[error("invalid share index")]
+    InvalidShareIndex,
+
+    /// Serialization failed.
+    #[error("marshal signed data: {0}")]
+    Serialize(#[from] serde_json::Error),
+}
+
+pub(crate) fn serialize_signed_data(data: &dyn SignedData) -> Result<Vec<u8>, ParSigExCodecError> {
```
iamquang95, line 55: This serialize and deserialize is not same as golang, the current one is forcing to use json, but in charon implementation, they will try the ssz first then use the json.

`core/proto.go`
```
// marshal marshals the given value into bytes, either as SSZ if supported by the type (and if enabled) or as json.
func marshal(v any) ([]byte, error) {
	// First try SSZ
	if marshaller, ok := v.(ssz.Marshaler); ok && sszMarshallingEnabled {
		b, err := marshaller.MarshalSSZ()
		if err != nil {
			return nil, errors.Wrap(err, "marshal ssz")
		}

		return b, nil
	}

	// Else try json
	b, err := json.Marshal(v)
	if err != nil {
		return nil, errors.Wrap(err, "marshal json")
	}

	return b, nil
}
```

crates/parsigex/src/behaviour.rs
```diff
@@ -0,0 +1,464 @@
...
+                        handler: NotifyHandler::Any,
+                        event: ToHandler::Send {
+                            request_id,
+                            payload: message.clone(),
+                        },
+                    });
+                    targeted = targeted.saturating_add(1);
+                }
+
+                if targeted == 0 {
```
iamquang95: Do you think we should send a `BroadcastFailed ` here? Otherwise the request will be "pending" forever, notify `BroadcastError` doesn't enough to mark a `request_id` is done/fail

crates/parsigex/src/error.rs
```diff
@@ -0,0 +1,76 @@
...
+pub enum Error {
+    /// Message conversion failed.
+    #[error(transparent)]
+    Codec(#[from] ParSigExCodecError),
+    /// Handle channel closed.
+    #[error("parsigex handle closed")]
+    Closed,
+    /// Broadcast failed for a peer.
+    #[error("broadcast to peer {peer} failed: {source}")]
+    BroadcastPeer {
```
iamquang95: No usage

crates/parsigex/src/error.rs
```diff
@@ -0,0 +1,76 @@
...
+        /// Peer for which the broadcast failed.
+        peer: PeerId,
+        /// Source failure.
+        #[source]
+        source: Failure,
+    },
+    /// Peer is not currently connected.
+    #[error("peer {0} is not connected")]
+    PeerNotConnected(PeerId),
```
iamquang95: No usage

crates/core/Cargo.toml
```diff
@@ -31,18 +32,25 @@ tokio-util.workspace = true
...
 prost.workspace = true
 prost-types.workspace = true
 hex.workspace = true
 chrono.workspace = true
 test-case.workspace = true
 pluto-eth2util.workspace = true
+pluto-cluster.workspace = true
+pluto-p2p.workspace = true
 pluto-testutil.workspace = true
+pluto-tracing.workspace = true
```
iamquang95: are these newly added dependencies in use?

crates/core/src/parsigex_codec.rs
```diff
@@ -0,0 +1,118 @@
...
+    #[error("invalid parsigex msg fields")]
+    InvalidMessageFields,
+
+    /// Invalid partial signed data set proto.
+    #[error("invalid partial signed data set proto fields")]
+    InvalidParSignedDataSetFields,
+
+    /// Invalid partial signed proto.
+    #[error("invalid partial signed proto")]
+    InvalidParSignedProto,
```
iamquang95: no usage

crates/parsigex/src/behaviour.rs
```diff
@@ -0,0 +1,464 @@
...
+    rx: mpsc::UnboundedReceiver<Command>,
+    pending_events: VecDeque<ToSwarm<Event, ToHandler>>,
+    pending_broadcasts: HashMap<u64, PendingBroadcast>,
+    shared_subs: Arc<SharedSubs>,
+}
+
+impl Behaviour {
+    /// Creates a behaviour and a clonable broadcast handle.
+    pub fn new(config: Config, peer_id: PeerId) -> (Self, Handle) {
+        debug_assert_eq!(config.peer_id, peer_id);
```
iamquang95: do we need this? or we can remove the `peer_id` in the param, then use directly from the `config`

crates/parsigex/src/handler.rs
```diff
@@ -0,0 +1,272 @@
...
+    active_futures: FuturesUnordered<ActiveFuture>,
+}
+
+impl Handler {
+    /// Creates a new handler for one connection.
+    pub fn new(
+        timeout: Duration,
+        verifier: Verifier,
+        duty_gater: DutyGater,
+        _peer: PeerId,
```
iamquang95: remove?

crates/parsigex/src/protocol.rs
```diff
@@ -0,0 +1,70 @@
...
+        .expect("vec-backed protobuf encoding cannot fail");
+    buf
+}
+
+/// Decodes a protobuf message from bytes.
+pub fn decode_protobuf<M: Message + Default>(
+    bytes: &[u8],
+) -> std::result::Result<M, prost::DecodeError> {
+    M::decode(bytes)
+}
```
iamquang95: can be inline

crates/core/src/parsigex_codec.rs
varex83: https://github.com/NethermindEth/pluto/pull/322

varex83agent, line 55: **Bug: JSON-only codec — Go uses SSZ-first serialization since v0.17, causing a hard interoperability failure**

`serialize_signed_data` and `deserialize_signed_data` use `serde_json` exclusively. Go's `marshal()` (`charon/core/proto.go:266`) tries `ssz.Marshaler` first, falling back to JSON. Go's `unmarshal()` tries SSZ then JSON. Since v0.17, `sszMarshallingEnabled = true` by default.

Affected types that Go writes as SSZ: `VersionedSignedProposal`, `Attestation`, `VersionedAttestation`, `SignedAggregateAndProof`, `VersionedSignedAggregateAndProof`, `SignedSyncMessage`, `SignedSyncContributionAndProof`.

A Go node will write SSZ; the Rust node will fail to decode it, returning `InvalidPayload` and silently dropping the partial signature. This is a complete interoperability failure for those duty types.

Fix: implement SSZ-first serialization/deserialization for all `SignedData` types that implement SSZ in Go, matching the `marshal`/`unmarshal` logic in `charon/core/proto.go:264-304`.

crates/core/src/types.rs
```diff
@@ -66,6 +71,52 @@ impl DutyType {
...
+            DutyType::BuilderProposer => 5,
+            DutyType::BuilderRegistration => 6,
+            DutyType::Randao => 7,
+            DutyType::PrepareAggregator => 8,
+            DutyType::Aggregator => 9,
+            DutyType::SyncMessage => 10,
+            DutyType::PrepareSyncContribution => 11,
+            DutyType::SyncContribution => 12,
+            DutyType::InfoSync => 13,
+            DutyType::DutySentinel(_) => 14,
```
varex83agent: **Bug: `DutySentinel` serializes as `14` but `TryFrom<i32>` has no arm for `14` — round-trip is broken**

`From<&DutyType> for i32` maps `DutySentinel(_) => 14` (here), but `TryFrom<i32>` for `DutyType` falls through to `_ => Err(ParSigExCodecError::InvalidDuty)` at line 115 — `14` is never matched.

In Go, `dutySentinel = 14` is an unexported constant used only as a loop bound and is never transmitted on the wire. The Rust `DutySentinel` variant should not be serializable to the wire format at all.

Fix: guard against encoding `DutySentinel` before it reaches `From<&DutyType> for i32` (return an error), and change `TryFrom<i32>` to return `Err(UnsupportedDutyType)` (not `InvalidDuty`) for value `14` to distinguish "recognised but unsupported" from "not a valid duty type".

crates/parsigex/src/behaviour.rs
varex83agent: **Bug: broadcast with no connected peers silently drops the `request_id` — no completion event is ever emitted**

When `targeted == 0` (all known peers are disconnected), the function returns early. It has already emitted individual `BroadcastError` events for each disconnected peer, but never emits `BroadcastFailed` or `BroadcastComplete`. The `request_id` returned by `Handle::broadcast()` is permanently orphaned — any caller awaiting a terminal event for it will block forever.

Go's `Broadcast` returns `nil` synchronously in this scenario, which is an implicit success. In async Rust, the caller receives `Ok(id)` from `Handle::broadcast()` _before_ the swarm processes the command, so the silent drop is observable as a hang.

Fix: emit `Event::BroadcastFailed { request_id }` before returning when `targeted == 0`.

crates/parsigex/src/behaviour.rs
```diff
@@ -0,0 +1,464 @@
...
+        _connection_id: ConnectionId,
+        peer: PeerId,
+        _addr: &Multiaddr,
+        _role_override: libp2p::core::Endpoint,
+        _port_use: libp2p::core::transport::PortUse,
+    ) -> std::result::Result<THandler<Self>, ConnectionDenied> {
+        Ok(self.connection_handler_for_peer(peer))
+    }
+
+    fn on_swarm_event(&mut self, _event: FromSwarm) {}
```
varex83agent: **Major: `on_swarm_event` ignores `ConnectionClosed` — `pending_broadcasts` leaks on mid-flight disconnect**

If a connection is torn down while a `PendingBroadcast` has `remaining > 0`, the `Handler` is dropped and its `active_futures` are abandoned. No `FromHandler::OutboundError` event is produced for in-flight sends, so `finish_broadcast_result` is never called and the `PendingBroadcast` entry leaks in `pending_broadcasts` forever.

Fix: handle `FromSwarm::ConnectionClosed` — iterate `pending_broadcasts` and call `finish_broadcast_result(request_id, true)` for any broadcast targeting the disconnected peer:

```rust
fn on_swarm_event(&mut self, event: FromSwarm) {
    if let FromSwarm::ConnectionClosed(e) = event {
        if e.remaining_established == 0 {
            // drain any pending broadcasts that included this peer
        }
    }
}
```

crates/parsigex/src/behaviour.rs
```diff
@@ -0,0 +1,464 @@
...
+            }
+        }
+    }
+
+    /// Notifies all registered subscribers of a received partial signature set.
+    ///
+    /// Each subscriber is invoked in a spawned task since `poll()` is
+    /// synchronous. This matches Go's intended behaviour (see Go TODO to call
+    /// subscribers async).
+    fn notify_subscribers(&self, duty: Duty, data_set: ParSignedDataSet) {
```
varex83agent, line 388: **Major: `notify_subscribers` spawns an unbounded number of Tokio tasks — DoS amplification and head-of-line blocking**

Every verified inbound message calls `tokio::spawn` with no bound. Under sustained inbound traffic, the number of live tasks grows unboundedly, consuming memory proportional to message rate.

Additionally, all subscriber callbacks are awaited **sequentially inside a single task**. If any subscriber is slow (e.g., blocked on a back-pressured channel), all subsequent subscribers in the list are stalled for that entire duration.

Fix: feed messages through a bounded `tokio::sync::mpsc` channel to a dedicated delivery task. Back-pressure is then applied at the receive side rather than accumulating as spawned tasks:

```rust
// In Behaviour:
dispatch_tx: mpsc::Sender<(Duty, ParSignedDataSet)>,  // bounded capacity
// In a dedicated spawned task:
while let Some((duty, set)) = rx.recv().await {
    for sub in subs.read().await.iter() { sub(duty.clone(), set.clone()).await; }
}
```

crates/parsigex/src/handler.rs
```diff
@@ -0,0 +1,272 @@
...
+
+type ActiveFuture = BoxFuture<'static, Option<FromHandler>>;
+
+/// Connection handler for parsigex.
+pub struct Handler {
+    timeout: Duration,
+    verifier: Verifier,
+    duty_gater: DutyGater,
+    pending_open: VecDeque<PendingOpen>,
+    active_futures: FuturesUnordered<ActiveFuture>,
```
varex83agent, line 83: **Major: unbounded `FuturesUnordered` allows inbound stream flooding — memory exhaustion**

Every accepted inbound stream pushes one future into `active_futures` with no capacity limit. The Yamux stream cap (2,048 per TCP connection) partially mitigates this on TCP, but QUIC connections have no equivalent cap applied here. With N cluster peers each opening max streams simultaneously, the handler can hold N × 2,048 concurrent futures, each holding a cloned `Verifier`, `DutyGater`, and stream handle.

Fix: add a per-handler high-water mark. When exceeded, close the new stream rather than pushing to `active_futures`:

```rust
const MAX_CONCURRENT_INBOUND: usize = 16;

fn handle_fully_negotiated_inbound(&mut self, stream: Stream) {
    if self.active_futures.len() >= MAX_CONCURRENT_INBOUND {
        // drop stream, log warning
        return;
    }
    // ...
}
```

crates/parsigex/src/handler.rs
```diff
@@ -0,0 +1,272 @@
...
+        .await
+        .map_err(Failure::io)?;
+    let (duty, data_set) = protocol::decode_message(&bytes).map_err(|_| Failure::InvalidPayload)?;
+    if !duty_gater(&duty) {
+        return Err(Failure::InvalidDuty);
+    }
+    for (pub_key, par_sig) in data_set.inner() {
+        verifier(duty.clone(), *pub_key, par_sig.clone())
+            .await
+            .map_err(|_| Failure::InvalidPartialSignature)?;
```
varex83agent: **Major: `VerifyError` is silently discarded — partial-signature fraud is unauditable**

```rust
verifier(...).await.map_err(|_| Failure::InvalidPartialSignature)?;
```

The `VerifyError` (which distinguishes `UnknownPubKey`, `InvalidShareIndex`, `InvalidSignedDataFamily`, `Other`) is thrown away with `|_|`. The `Event::Error { error: Failure::InvalidPartialSignature }` delivered to the application carries no information about which key failed, why, or from which peer.

In a distributed validator this is a primary security signal — a rogue node sending crafted partial signatures cannot be detected or attributed. Fix: at minimum log the error with structured fields before mapping:

```rust
.map_err(|e| {
    tracing::warn!(peer = %peer_id, err = %e, pubkey = %pub_key, "partial sig verification failed");
    Failure::InvalidPartialSignature
})
```

crates/core/src/signeddata.rs
```diff
@@ -48,6 +48,9 @@ pub enum SignedDataError {
     /// Invalid attestation wrapper JSON.
     #[error("unmarshal attestation")]
     AttestationJson,
+    /// Custom error.
+    #[error("{0}")]
+    Custom(Box<dyn std::error::Error>),
```
varex83agent: **Major: `SignedDataError::Custom` wraps `Box<dyn Error>` without `Send + Sync` — breaks async usage**

`Box<dyn std::error::Error>` does not require `Send` or `Sync`. `SignedDataError` is returned by `SignedData::signature()`, which is called inside the async parsigex pipeline (handler.rs) and in `TryFrom` impls used across `.await` points in spawned tasks.

Holding a `Custom` variant across an `.await` point in a `tokio::spawn` task will either fail to compile or silently constrain concrete error types to `Send + Sync`.

Fix:
```rust
Custom(Box<dyn std::error::Error + Send + Sync>),
```

crates/core/src/types.rs
```diff
@@ -517,6 +590,39 @@ impl ParSignedData {
...
 
+impl TryFrom<&ParSignedData> for pbcore::ParSignedData {
+    type Error = ParSigExCodecError;
+
+    fn try_from(data: &ParSignedData) -> Result<Self, Self::Error> {
+        let encoded = serialize_signed_data(data.signed_data.as_ref())?;
+        let share_idx =
+            i32::try_from(data.share_idx).map_err(|_| ParSigExCodecError::InvalidShareIndex)?;
+        let signature = data.signed_data.signature().map_err(|err| {
+            ParSigExCodecError::Serialize(serde_json::Error::io(std::io::Error::other(
```
varex83agent: **Major: roundabout error conversion through `serde_json::Error::io` for a non-I/O failure**

```rust
ParSigExCodecError::Serialize(serde_json::Error::io(std::io::Error::other(err.to_string())))
```

A `SignedDataError` from `signature()` is being tunnelled through: string → `io::Error` → `serde_json::Error` → `ParSigExCodecError::Serialize`. This is semantically wrong — a signature extraction failure is not a serialization error.

Fix: add a dedicated variant:
```rust
// in parsigex_codec.rs ParSigExCodecError:
#[error("extract signature: {0}")]
ExtractSignature(String),
```
And map to it directly: `.map_err(|e| ParSigExCodecError::ExtractSignature(e.to_string()))`.

crates/parsigex/src/protocol.rs
```diff
@@ -0,0 +1,70 @@
...
+use pluto_p2p::proto;
+
+use super::{Error, Result as ParasigexResult};
+
+/// Encodes a protobuf message to bytes.
+pub fn encode_protobuf<M: Message>(message: &M) -> Vec<u8> {
+    let mut buf = Vec::with_capacity(message.encoded_len());
+    message
+        .encode(&mut buf)
+        .expect("vec-backed protobuf encoding cannot fail");
```
varex83agent: **Major: `.expect()` in non-test code violates the project no-panic rule**

The project style guide prohibits `unwrap()`/`expect()`/`panic!()` outside test code. While encoding into a `Vec<u8>` is infallible in prost, the rule is absolute.

Fix: use `prost::Message::encode_to_vec` which is infallible and returns `Vec<u8>` directly (available since prost 0.11):
```rust
pub fn encode_protobuf<M: Message>(message: &M) -> Vec<u8> {
    message.encode_to_vec()
}
```
Or, if `encode_to_vec` is not available, use `encode_raw` which returns no `Result`.

crates/peerinfo/src/protocol.rs
```diff
@@ -19,6 +19,7 @@ use std::{
 use chrono::{DateTime, Utc};
 use libp2p::{PeerId, swarm::Stream};
 use pluto_core::version::{self, SemVer, SemVerError};
+use pluto_p2p::proto::MAX_MESSAGE_SIZE;
```
varex83agent: **Major: peerinfo `MAX_MESSAGE_SIZE` increased 2048× from 64 KiB to 128 MiB — memory amplification**

Before this PR, peerinfo defined `const MAX_MESSAGE_SIZE: usize = 64 * 1024` with a comment that "64KB should be plenty for peer info". This PR (through the shared-helpers refactor) replaces it with `pluto_p2p::proto::MAX_MESSAGE_SIZE` (128 MiB).

`read_protobuf_with_max_size` pre-allocates based on the length prefix sent by the remote. A malicious peer can now send a peerinfo message claiming 128 MiB and force the receiver to allocate 128 MiB per connection before any data is read.

Go uses 128 MiB uniformly (`maxMsgSize = 128 << 20` in `charon/p2p/sender.go:27`), so the current behavior is technically Go-compatible. However a per-protocol guard is safer and was already in place.

Fix: restore a peerinfo-specific constant:
```rust
// peerinfo/src/protocol.rs
const PEERINFO_MAX_MESSAGE_SIZE: usize = 64 * 1024;
```

crates/parsigex/src/error.rs
```diff
@@ -0,0 +1,76 @@
...
+    InvalidPayload,
+    /// Duty not accepted by the gater.
+    #[error("invalid duty")]
+    InvalidDuty,
+    /// Signature verification failed.
+    #[error("invalid partial signature")]
+    InvalidPartialSignature,
+    /// I/O error.
+    #[error("{0}")]
+    Io(String),
```
varex83agent: **Minor: `Failure::Io(String)` discards `io::ErrorKind`**

Converting `io::Error` to `String` immediately loses the `ErrorKind`, which is useful for distinguishing reset connections from timeouts, partial reads, etc. The `bcast` and `sync` crates have the same pattern (flagged in prior reviews).

`Failure` derives `Clone`, preventing a direct `#[from] io::Error`. Consider:
```rust
#[error("i/o: {0}")]
Io(Arc<std::io::Error>),  // Arc makes it Clone while preserving ErrorKind
```
Or remove the `Clone` bound from `Failure` and use `#[from] io::Error` directly.

crates/parsigex/src/behaviour.rs
```diff
@@ -0,0 +1,464 @@
...
+            .map_err(|_| Error::Closed)?;
+        Ok(request_id)
+    }
+
+    /// Registers a callback when a verified partial signature set is received
+    /// from a peer.
+    ///
+    /// This is not thread safe with respect to message delivery ordering — it
+    /// must be called before the swarm starts processing events.
+    pub async fn subscribe(&self, sub: ReceivedSub) {
```
varex83agent, line 159: **Minor: misleading doc comment on `subscribe` — says "not thread safe" but uses `Mutex`**

The comment _"This is not thread safe with respect to message delivery ordering"_ is factually wrong — `Mutex` is thread-safe. The real concern is a **delivery ordering hazard**: a message dispatched by the swarm between the moment `subscribe` is called and the moment the lock is acquired will be delivered to the new subscriber, but a message already mid-flight before `subscribe` was called will not.

Suggest rewording:
```
/// Subscribers registered after the swarm begins polling may miss messages
/// already in flight. Register all subscribers before starting the event loop.
```

crates/parsigex/src/behaviour.rs
```diff
@@ -0,0 +1,464 @@
...
+        request_id: u64,
+        duty: Duty,
+        data_set: ParSignedDataSet,
+    },
+}
+
+/// Shared subscriber list between [`Handle`] and [`Behaviour`].
+#[derive(Default)]
+struct SharedSubs {
+    subs: Mutex<Vec<ReceivedSub>>,
```
varex83agent: **Minor: `SharedSubs` uses `Mutex` but subscriptions are read-mostly — use `RwLock`**

Subscriptions are registered once at startup and then only read on every inbound message. `Mutex` serialises all reads, even when no writes are happening. Use `tokio::sync::RwLock` and acquire a read guard in `notify_subscribers`:

```rust
struct SharedSubs {
    subs: RwLock<Vec<ReceivedSub>>,
}
// in notify_subscribers:
let subs = shared_subs.subs.read().await.clone();
```

crates/parsigex/src/behaviour.rs
```diff
@@ -0,0 +1,464 @@
...
+    config: Config,
+    rx: mpsc::UnboundedReceiver<Command>,
+    pending_events: VecDeque<ToSwarm<Event, ToHandler>>,
+    pending_broadcasts: HashMap<u64, PendingBroadcast>,
+    shared_subs: Arc<SharedSubs>,
+}
+
+impl Behaviour {
+    /// Creates a behaviour and a clonable broadcast handle.
+    pub fn new(config: Config, peer_id: PeerId) -> (Self, Handle) {
```
varex83agent: **Minor: redundant `peer_id` parameter — only used in a `debug_assert`, never stored**

`pub fn new(config: Config, peer_id: PeerId) -> (Self, Handle)` accepts `peer_id` solely for `debug_assert_eq!(config.peer_id, peer_id)`. The value is never stored; in release builds the assertion is compiled out and a mismatching `peer_id` has no effect.

`Config` already carries `peer_id`. Remove the constructor parameter and use `config.peer_id` directly everywhere the local `peer_id` was referenced.

crates/parsigex/src/handler.rs
varex83agent: **Minor: `_peer: PeerId` is accepted but never stored or used**

The parameter is discarded immediately. Remove it from `Handler::new` and update the call-site in `behaviour.rs` (`Handler::new(self.config.timeout, …, peer)` → drop `peer`).

crates/parsigex/src/lib.rs
```diff
@@ -0,0 +1,24 @@
...
+pub use handler::Handler;
+pub use protocol::{decode_message, encode_message};
+
+use libp2p::swarm::StreamProtocol;
+
+/// The protocol name for partial signature exchange (version 2.0.0).
+pub const PROTOCOL_NAME: StreamProtocol = StreamProtocol::new("/charon/parsigex/2.0.0");
+
+/// Returns the supported protocols in precedence order.
+pub fn protocols() -> Vec<StreamProtocol> {
```
varex83agent, line 22: **Minor: `protocols()` is never called — dead public API**

`pub fn protocols() -> Vec<StreamProtocol>` is exported but no consumer uses it (nothing in `p2p.rs`, `behaviour.rs`, or the examples references it). Either wire it into the swarm setup where a protocol list is needed, or remove it to avoid a `dead_code` lint in non-doc builds.

crates/parsigex/src/protocol.rs
```diff
@@ -0,0 +1,70 @@
...
+use libp2p::swarm::Stream;
+use prost::Message;
+
+use pluto_core::{
+    corepb::v1::{core as pbcore, parsigex as pbparsigex},
+    types::{Duty, ParSignedDataSet},
+};
+use pluto_p2p::proto;
+
+use super::{Error, Result as ParasigexResult};
```
varex83agent: nit: `Result as ParasigexResult` — "Parasigex" is a misspelling. The crate is named `parsigex`. Rename the alias to `ParsigexResult` (or just use `Result` with a qualified path).

crates/core/src/parsigex_codec.rs
```diff
@@ -0,0 +1,118 @@
...
+        DutyType::Signature => deserialize_json!(Signature),
+        DutyType::PrepareAggregator => deserialize_json!(BeaconCommitteeSelection),
+        // Match Go order: old SignedAggregateAndProof format first, then versioned.
+        DutyType::Aggregator => deserialize_json!(SignedAggregateAndProof)
+            .or_else(|_| deserialize_json!(VersionedSignedAggregateAndProof))
+            .map_err(|_| ParSigExCodecError::UnsupportedDutyType),
+        DutyType::SyncMessage => deserialize_json!(SignedSyncMessage),
+        DutyType::PrepareSyncContribution => deserialize_json!(SyncCommitteeSelection),
+        DutyType::SyncContribution => deserialize_json!(SignedSyncContributionAndProof),
+        DutyType::Unknown | DutyType::InfoSync | DutyType::DutySentinel(_) => {
```
varex83agent: nit: `DutyType::InfoSync` rejection is correct (matches Go's `default` branch in `ParSignedDataFromProto`) but has no comment explaining the intent. A future maintainer might assume this is a missing case. Add:
```rust
// InfoSync is not used in parsigex in Go (handled via a separate channel);
// Unknown and DutySentinel are sentinel/invalid values that are never transmitted.
DutyType::Unknown | DutyType::InfoSync | DutyType::DutySentinel(_) => {
    Err(ParSigExCodecError::UnsupportedDutyType)
}
```

varex83agent (changes requested):
## Review: feat: add parsigex

This PR ports the Charon parsigex libp2p protocol to Rust. The architecture is sound — the Behaviour/Handler split, the per-stream one-shot send/receive design, and the duty codec fallback order (Attestation before VersionedAttestation, SignedAggregateAndProof before versioned) all match Go correctly.

There are **3 bugs that must be fixed before merge**:
1. The codec is JSON-only — Go writes SSZ-first since v0.17, causing a hard interoperability failure for all major duty types (`crates/core/src/parsigex_codec.rs`)
2. `DutySentinel` encodes as `14` but `TryFrom<i32>` has no arm for `14`, making the round-trip fail with `InvalidDuty` (`crates/core/src/types.rs`)
3. Broadcast with no connected peers silently drops the `request_id` without emitting a terminal event, leaving callers blocked forever (`crates/parsigex/src/behaviour.rs:304`)

Additionally there are **8 major issues**: the peerinfo max-message-size regression (64 KiB → 128 MiB), `pending_broadcasts` leaking on connection close, unbounded `tokio::spawn` in `notify_subscribers`, `FuturesUnordered` with no concurrency cap, `VerifyError` being silently discarded, `SignedDataError::Custom` not being `Send + Sync`, a roundabout error conversion through `serde_json::Error::io`, and a `panic!`-eligible `.expect()` in non-test code. Inline comments are filed for each issue individually.

crates/core/src/parsigex_codec.rs
varex83: This is addressed in #322

crates/core/Cargo.toml
emlautarom1: None of the added deps are used:

- anyhow
- clap
- k256
- pluto-cluster
- pluto-p2p
- pluto-tracing

crates/core/src/types.rs
emlautarom1: Generally agree, but I suggest replacing the `From` with `TryFrom`. Also, look into the possibility of dropping the `DutySentinel` if it's not needed (even if we deviate from Charon, it's most likely an artifact of Go not having reasonable union types)

crates/core/src/types.rs
emlautarom1: Agree on this, the error has nothing to do with IO. Prefer to add another variant to `ParSigExCodecError` like `ParSigExCodecError::InvalidSignature`

crates/peerinfo/src/protocol.rs
emlautarom1: Why the change here?

crates/parsigex/src/protocol.rs
```diff
@@ -0,0 +1,70 @@
...
+
+/// Encodes a protobuf message to bytes.
+pub fn encode_protobuf<M: Message>(message: &M) -> Vec<u8> {
```
emlautarom1: Can be fully replaced with `message.encode_to_vec()`

crates/parsigex/examples/parsigex.rs
```diff
@@ -0,0 +1,565 @@
+#![allow(missing_docs)]
```
emlautarom1, line 1: Not needed

crates/parsigex/examples/parsigex.rs
```diff
@@ -0,0 +1,565 @@
...
+//! # Running a multi-node setup
+//!
+//! ## 1. Create a cluster
+//!
+//! Use the Charon Go CLI to generate per-node data directories, each containing
+//! a `charon-enr-private-key` and a shared `cluster-lock.json`:
+//!
+//! ```bash
+//! charon create cluster --name parsigex-test --nodes 3 --threshold 2 --no-verify \
+//!   --cluster-dir ./cluster
```
emlautarom1: This command did not work as intended, instead I had to run:

```
./charon create cluster --nodes 3 --threshold 2 --num-validators 1 --network mainnet --cluster-dir ./parsigex-example-cluster --name parsigex-example --insecure-keys --fee-recipient-addresses 0x0000000000000000000000000000000000000000 --withdrawal-addresses 0x0000000000000000000000000000000000000000
```

crates/parsigex/examples/parsigex.rs
```diff
@@ -0,0 +1,565 @@
...
+                                share_idx,
+                                "sample partial signature broadcast finished with failures"
+                            );
+                        } else {
+                            warn!(request_id, "partial signature broadcast finished with failures");
+                        }
+                    }
+                    SwarmEvent::NewListenAddr { address, .. } => {
+                        info!(address = %address, "listening");
+                    }
```
emlautarom1: `pluto_p2p` already logs this information. Remove to avoid duplication in logs.

crates/parsigex/src/behaviour.rs
emlautarom1: Agree on this, unnecessary duplication that does not perform any additional checks.

crates/parsigex/examples/parsigex.rs
```diff
@@ -0,0 +1,565 @@
...
+            let local_peer_id = keypair.public().to_peer_id();
+            let config = parsigex::Config::new(
+                local_peer_id,
+                p2p_context.clone(),
+                verifier.clone(),
+                duty_gater.clone(),
+            )
+            .with_timeout(Duration::from_secs(10));
+            let (parsigex, handle) = parsigex::Behaviour::new(config, local_peer_id);
+            parsigex_handle = Some(handle);
```
emlautarom1, line 282: This is very cumbersome from a UX perspective:
- If we set up a parsigex then we should not need an `Optional` over the handle
- We should not need a handle to manipulate the behavior. Instead it would be better to handle it directly through a reference to the behavior.

I'd suggest:

- Expose a method in `Node<B>` to access the inner behavior `B` as mutable, something like `self.swarm.behaviour_mut().inner.as_mut().unwrap()`. To avoid the unwrap we could change the inner behavior in Node from `inner: OptionalBehaviour<B>` to a `inner: B` where we default `B` to a `NoopBehavior` such that we can always extract an inner behavior (this could be tricky to set up so can be left for later) 
- If the handle is needed, then the behavior itself could expose a method to produce handles.

crates/parsigex/examples/parsigex.rs
```diff
@@ -0,0 +1,565 @@
...
+    info!(
+        peer_id = %node.local_peer_id(),
+        data_dir = %args.data_dir.display(),
+        known_peers = ?known_peers,
+        relays = ?args.relays,
+        "parsigex example started"
+    );
+
+    let mut ticker = tokio::time::interval(Duration::from_secs(args.broadcast_every));
+    let mut pending_broadcasts: HashMap<u64, (Duty, u64)> = HashMap::new();
```
emlautarom1, line 307: question: do we need something like this in Pluto itself or is it for demo purposes? In case we need it in Pluto, it would be better for the behavior to expose them in some way (currently private field), so consumers don't need to care about managing them.

crates/parsigex/src/behaviour.rs
```diff
@@ -0,0 +1,464 @@
...
+}
+
+#[derive(Debug)]
+struct PendingBroadcast {
+    remaining: usize,
+    failed: bool,
+}
+
+#[derive(Debug)]
+enum Command {
```
emlautarom1: question: Do we intend to extend this command later? If not using the command pattern is a bit overkill.

crates/parsigex/src/behaviour.rs
emlautarom1: The wording is wrong in Charon, let's fix it here.

emlautarom1: Also, this matches Charon but at the same time due to the event-based style of Rust's libp2p we already have access to the events (note how the example does not use it, instead only requires the `broadcast()` method). Do we need a `subscribe()` method (and all its associated types)

crates/parsigex/src/behaviour.rs
emlautarom1: Agree, this should only happen when we cannot connect to any of the cluster peers. We should run the example with a single node running (it will fail to send to the other two) and see what happens with the pending requests.

crates/parsigex/src/behaviour.rs
emlautarom1: Agree on the fact that sequential execution of subscriber callbacks is not good. Still, I believe that we can drop the entire subscribe mechanism given how Rust's libp2p already uses events.

crates/parsigex/src/handler.rs
```diff
@@ -0,0 +1,272 @@
...
+    OutboundError {
+        /// Request identifier.
+        request_id: u64,
+        /// Failure reason.
+        error: Failure,
+    },
+}
+
+/// Outbound open info that carries the request context through stream
+/// negotiation.
```
emlautarom1: The comment does not give any useful context. What is the request id? What is the payload?

crates/parsigex/src/handler.rs
emlautarom1: Not really important since we're gating the connections to the cluster known peers, which is always a reasonably bounded number.

crates/parsigex/src/handler.rs
emlautarom1: It would be nice if we included the error in the `InvalidPartialSignature` variant so that it can be logged at a higher level. We should not discard the error.

emlautarom1 (changes requested):
In general LGTM, I cannot fully comment on the correctness of the `Handler` though.

The example works as intended (with the updated command) but there are some small things to sort out before merging.

crates/parsigex/examples/parsigex.rs
emlautarom1: Even better, use the built in `create cluster` (see #339)

crates/parsigex/src/behaviour.rs
varex83: I would revisit it a little bit later. I've tried exploring other implementations, but came up that the current is the best way to go with. The main issue arise when we will use it in the `exchanger`, which heavily relies on the subscriber pattern

iamquang95 (approved):
Code LGTM, the only missing part is https://github.com/NethermindEth/pluto/pull/322

crates/parsigex/src/behaviour.rs
emlautarom1: Let's create an issue to revisit this later.

crates/core/Cargo.toml
```diff
@@ -12,6 +12,7 @@ cancellation.workspace = true
 chrono.workspace = true
 crossbeam.workspace = true
 futures.workspace = true
+futures-timer.workspace = true
```
emlautarom1: Unused dependencies:

- `unsigned-varint`
- `futures-timer`

emlautarom1 (approved):
LGTM, there are some unused dependencies that we should remove in `core`

I cannot test the example with the `pluto create cluster` command because main is not fully merged into this feature branch. Before merging this PR please confirm that it works as intended.

As for the subscriber pattern, let's create an issue to revisit this later.

varex83agent (approved):

PR #305, opened 2026-03-31
feat: pluto create cluster

### LINKED ISSUE #301: Implement `charon create cluster`

varex83:
# Summary
Implements the initial pluto create cluster command and wires it into the CLI.

This adds local cluster artifact generation, including cluster definitions/locks, validator key material, deposit data, and node-specific setup, along with supporting validation and error handling for network, threshold, keymanager, and withdrawal address configuration.

## Testing

To keep PR scope smaller, tests will come in a separate PR.


crates/cli/src/error.rs
```diff
@@ -26,6 +29,7 @@ impl Termination for ExitResult {
 
 /// Errors that can occur in the Pluto CLI.
 #[derive(Error, Debug)]
+#[allow(dead_code)]
```
emlautarom1: Unnecessary attribute.

crates/cli/src/error.rs
```diff
@@ -5,8 +5,11 @@ use std::{
     process::{ExitCode, Termination},
 };
 
+use pluto_eth2util as eth2util;
```
emlautarom1: nit: no need for these kind of aliases, makes it a bit harder to follow the code.

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+                    .as_ref()
+                    .unwrap_or(&String::new())
+                    .clone()
+                    .into_boxed_str(),
+            ),
+            genesis_timestamp: config.genesis_timestamp.unwrap_or(0),
+            capella_hard_fork: "",
+        }
+    }
+}
```
emlautarom1, line 358: nit: the `Box::leak` is fine here since we only need it once, but I'd rather make changes in `pluto_eth2util::network::Network` to support non static strings.

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+            number_of_nodes,
+        }
+        .into());
+    }
+
+    Ok(())
+}
+
+/// Runs the create cluster command
+pub async fn run(w: &mut dyn Write, mut args: CreateClusterArgs) -> CliResult<()> {
```
emlautarom1, line 389: Missing tracing initialization.

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+
+async fn get_keys(
+    split_keys_dir: impl AsRef<Path>,
+    use_sequence_keys: bool,
+) -> Result<Vec<PrivateKey>> {
+    if use_sequence_keys {
+        let files = load_files_unordered(&split_keys_dir).await?;
+        Ok(files.sequenced_keys()?)
+    } else {
+        let files = load_files_recursively(&split_keys_dir).await?;
```
emlautarom1: nit: unnecessary refs here


```suggestion
        let files = load_files_unordered(split_keys_dir).await?;
        Ok(files.sequenced_keys()?)
    } else {
        let files = load_files_recursively(split_keys_dir).await?;
```

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+        // Needed if --split-existing-keys is called without a definition file.
+        // It's safe to unwrap here because we know the length is less than u64::MAX.
+        args.num_validators =
+            Some(u64::try_from(secrets.len()).expect("secrets length is too large"));
+    }
+
+    // Get a cluster definition, either from a definition file or from the config.
+    let definition_file = args.definition_file.clone();
+    let (mut def, mut deposit_amounts) = if let Some(definition_file) = definition_file {
+        let Some(addr) = args.execution_engine_addr.clone() else {
```
emlautarom1: nit: you can avoid cloning here


```suggestion
    let definition_file = args.definition_file.as_ref()
    let (mut def, mut deposit_amounts) = if let Some(definition_file) = definition_file {
        let Some(addr) = args.execution_engine_addr.as_ref() else {
```

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+
+        let eth1cl = eth1wrap::EthClient::new(addr).await?;
+
+        let def = load_definition(&definition_file, &eth1cl).await?;
+
+        // Should not happen, if it does - it won't affect the runtime, because the
+        // validation will fail.
+        args.nodes =
+            Some(u64::try_from(def.operators.len()).expect("operators length is too large"));
+        args.threshold = Some(def.threshold);
```
emlautarom1: In Charon, these modifications to the args are done before running `validate_threshold` and `validate_create_config`, while here they're performed after, meaning that with some cluster definitions Pluto could fail unlike Charon.

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+
+        eth2util::deposit::verify_deposit_amounts(&amount, args.compounding)?;
+    }
+
+    for addr in &args.keymanager_addrs {
+        let keymanager_url =
+            url::Url::parse(addr).map_err(CreateClusterError::InvalidKeymanagerUrl)?;
+
+        if keymanager_url.scheme() != HTTP_SCHEME {
+            return Err(CreateClusterError::InvalidKeymanagerUrlScheme { addr: addr.clone() });
```
emlautarom1: This is a warning in Charon but here it fails.

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+        }
+    }
+
+    if args.split_keys && args.num_validators.is_some() {
+        return Err(CreateClusterError::CannotSpecifyNumValidatorsWithSplitKeys);
+    } else if !args.split_keys && args.num_validators.is_none() && args.definition_file.is_none() {
```
emlautarom1: `num_validators` could be `Some(0)` and this check passes. I'd extract it and have `let num_validators = args.num_validators.unwrap_or(0)` and then the check would be `num_validators == 0`

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+
+    /// Ethereum network to create validators for
+    #[arg(long = "network", help = "Ethereum network to create validators for.")]
+    pub network: Option<Network>,
+
+    /// The number of charon nodes in the cluster
+    #[arg(
+        long = "nodes",
+        help = "The number of charon nodes in the cluster. Minimum is 3."
+    )]
```
emlautarom1, line 262: There are several places where we do `.unwrap_or(0)` and Charon does not use `nil` for this. I'd suggest using `0` as default:


```suggestion
    #[arg(
        long = "nodes",
        default_value = "0",
        help = "The number of charon nodes in the cluster. Minimum is 3."
    )]
    pub nodes: u64,
```

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+/// 1.
+fn validate_addresses(
+    num_validators: u64,
+    fee_recipient_addrs: &[String],
+    withdrawal_addrs: &[String],
+) -> Result<(Vec<String>, Vec<String>)> {
+    let num_validators_usize =
+        usize::try_from(num_validators).map_err(|_| CreateClusterError::ValueExceedsU8 {
+            value: num_validators,
+        })?;
```
emlautarom1: This error is incorrect (there is no `u8` in use here).

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+        let files = load_files_unordered(&split_keys_dir).await?;
+        Ok(files.sequenced_keys()?)
+    } else {
+        let files = load_files_recursively(&split_keys_dir).await?;
+        Ok(files.keys())
+    }
+}
+
+/// Creates a new cluster definition from the provided configuration.
+async fn new_def_from_config(args: &CreateClusterArgs) -> Result<Definition> {
```
emlautarom1:
```suggestion
fn new_def_from_config(args: &CreateClusterArgs) -> Result<Definition> {
```

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+    if secrets.is_empty() {
+        // This is the case in which split-keys is undefined and user passed validator
+        // amount on CLI
+        secrets = generate_keys(def.num_validators)?;
+    }
+
+    let num_validators_usize =
+        usize::try_from(def.num_validators).map_err(|_| CreateClusterError::ValueExceedsU8 {
+            value: def.num_validators,
+        })?;
```
emlautarom1: Same situation regarding the error (does `num_validators` have to fit in `u8`?)

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+
+        // Preserve order when transforming from map of private shares to array of
+        // private keys
+        let mut secret_set = vec![PrivateKey::default(); shares.len()];
+        for i in 1..=shares.len() {
+            let i_u64 = u64::try_from(i).expect("shares length should fit in u64 on all platforms");
+            let idx =
+                u8::try_from(i).map_err(|_| CreateClusterError::ValueExceedsU8 { value: i_u64 })?;
+            secret_set[i.saturating_sub(1)] = shares[&idx];
+        }
```
emlautarom1: Even if this matches the Charon impl closely it's a bit cumbersome due to the amount of type conversions. The following should be equivalent:


```suggestion
        let mut entries: Vec<_> = shares.into_iter().collect();
        entries.sort_by_key(|(k, _)| *k);
        let secret_set = entries.into_iter().map(|(_, v)| v).collect();
```

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+        .clone()
+        .iter()
+        .map(|shares| {
+            shares
+                .iter()
+                .map(|s| tbls.secret_to_public_key(s))
+                .collect::<std::result::Result<Vec<_>, _>>()
+        })
+        .collect::<std::result::Result<Vec<_>, _>>()
+        .map_err(CreateClusterError::CryptoError)?;
```
emlautarom1: nit: we can simplify some code


```suggestion
    let pub_shares_sets: Vec<Vec<PublicKey>> = share_sets
        .iter()
        .map(|shares| {
            shares
                .iter()
                .map(|s| tbls.secret_to_public_key(s))
                .collect()
        })
        .collect::<std::result::Result<_, _>>()
        .map_err(CreateClusterError::CryptoError)?;
```

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+}
+
+async fn write_keys_to_disk(
+    num_nodes: u64,
+    cluster_dir: impl AsRef<Path>,
+    insecure_keys: bool,
+    share_sets: &[Vec<PrivateKey>],
+) -> Result<()> {
+    for i in 0..num_nodes {
+        let i_usize = usize::try_from(i).expect("node index should fit in usize on all platforms");
```
emlautarom1, line 706: nit: we could make `num_nodes: usize` and avoid conversions on each iteration

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+
+        let mut secrets: Vec<PrivateKey> = Vec::new();
+        for shares in share_sets {
+            secrets.push(shares[i_usize]);
+        }
+
+        let keys_dir = create_validator_keys_dir(node_dir(cluster_dir.as_ref(), i))
+            .await
+            .map_err(CreateClusterError::IoError)?;
```
emlautarom1: Unnecessary `map_err`

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+    path::{Path, PathBuf},
+};
+
+use chrono::Utc;
+use k256::SecretKey;
+use pluto_cluster::{
+    definition::Definition,
+    deposit::DepositData,
+    distvalidator::DistValidator,
+    helpers::{create_validator_keys_dir, fetch_definition},
```
emlautarom1: nit: I'd suggest against importing functions, makes code a bit harder to read (not immediately obvious that we're using external code)

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+
+fn random_hex64() -> Result<String> {
+    let mut bytes = [0u8; 32];
+    rand::RngCore::fill_bytes(&mut OsRng, &mut bytes);
+    Ok(hex::encode(bytes))
+}
+
+async fn write_keys_to_keymanager(
+    args: &CreateClusterArgs,
+    num_nodes: u64,
```
emlautarom1, line 734: Same suggestions as in `write_keys_to_disk` (`num_nodes: usize`)

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+        let i_usize = usize::try_from(i).expect("node index should fit in usize on all platforms");
+
+        let mut keystores: Vec<Keystore> = Vec::new();
+        let mut passwords: Vec<String> = Vec::new();
+
+        // share_sets[validator_idx][node_idx]
+        for shares in share_sets {
+            let password = random_hex64()?;
+            let pbkdf2_c = if args.insecure_keys {
+                Some(16u32)
```
emlautarom1, line 762: We should document why `16` (== `2^4` as used in Charon)

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+        return Err(CreateClusterError::InsufficientWithdrawalAddresses);
+    }
+    if deposit_amounts.is_empty() {
+        return Err(CreateClusterError::EmptyDepositAmounts);
+    }
+    let tbls = BlstImpl;
+    let mut dd = Vec::new();
+    for &deposit_amount in deposit_amounts {
+        let mut datas = Vec::new();
+        for (secret, withdrawal_addr) in secrets.iter().zip(withdrawal_addresses.iter()) {
```
emlautarom1, line 833: Missing `eth2util::helpers::checksum_address(withdrawal_addr)?;`

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+    sign_deposit_datas(
+        secrets,
+        withdrawal_addresses,
+        network.as_ref(),
+        &deduped,
+        compounding,
+    )
+}
+
+fn sign_deposit_datas(
```
emlautarom1, line 816:
```suggestion
/// Returns `Vec<DepositData>` for each partial deposit amount.
fn sign_deposit_datas(
```

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+        &network,
+        &secrets,
+        &deposit_amounts,
+        def.compounding,
+    )?;
+
+    let eth2util_deposit_datas = deposit_datas
+        .iter()
+        .map(|dd| cluster_deposit_data_to_eth2util_deposit_data(dd))
+        .collect::<Vec<_>>();
```
emlautarom1, line 528: We should not need this conversion; instead `create_deposit_datas` should return the correct `DepositData` type (`eth2util::deposit::DepositData`)

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+            eth2util::network::fork_version_to_genesis_time(fork_version)?
+        };
+
+        let pk = tbls.secret_to_public_key(secret)?;
+
+        let unsigned_reg = eth2util_registration::new_message(
+            pk,
+            fee_address,
+            effective_gas_limit,
+            u64::try_from(timestamp.timestamp()).unwrap_or(0),
```
emlautarom1: This should be an expect, if there are issues with conversions we don't want to silently fail

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+            "Custom target gas limit not supported, setting to default"
+        );
+        eth2util_registration::DEFAULT_GAS_LIMIT
+    } else {
+        target_gas_limit
+    };
+
+    let fork_version_arr: [u8; 4] = fork_version
+        .try_into()
+        .map_err(|_| CreateClusterError::InvalidForkVersionLength)?;
```
emlautarom1: nit: you can shadow the variable here, no need to keep the `_arr` and the original argument at the same time.

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+    def.operators = ops;
+
+    let keys_to_disk = args.keymanager_addrs.is_empty();
+
+    // Pre-compute public shares and clone private shares before share_sets is
+    // consumed by the write step.  The private share clone is needed for the
+    // aggregate BLS signature; the public shares are needed for DistValidator.
+    // Each entry is [share_node0, share_node1, ...] indexed by validator.
+    let tbls = BlstImpl;
+    let pub_shares_sets: Vec<Vec<PublicKey>> = share_sets
```
emlautarom1: Not sure why we have this difference with the original implementation. Is this intended to be an optimization in some way?

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+    use pluto_crypto::types::Signature;
+
+    let tbls = BlstImpl;
+    let mut sigs: Vec<Signature> = Vec::new();
+
+    for shares in share_sets {
+        for share in shares {
+            let sig = tbls
+                .sign(share, message)
+                .map_err(CreateClusterError::CryptoError)?;
```
emlautarom1: Unnecessary `map_err`

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+
+    Ok(vals)
+}
+
+/// Aggregates BLS signatures over `message` produced by every private share
+/// across all validators, mirroring Go's `aggSign`.
+///
+/// `share_sets` — outer dimension is per-validator, inner is per-node private
+/// key share.
+fn agg_sign(share_sets: &[Vec<PrivateKey>], message: &[u8]) -> Result<Vec<u8>> {
```
emlautarom1: The comment and parameter names are a bit confusing, they deviate from Charon unnecessarily.

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+        ..Default::default()
+    };
+
+    lock.set_lock_hash().map_err(CreateClusterError::from)?;
+
+    lock.signature_aggregate = agg_sign(&share_sets, &lock.lock_hash)?;
+
+    for op_key in &node_keys {
+        let node_sig =
+            pluto_k1util::sign(op_key, &lock.lock_hash).map_err(CreateClusterError::K1UtilError)?;
```
emlautarom1, line 561: Unnecessary `map_err`

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+
+        tokio::fs::write(&lock_path, &bytes)
+            .await
+            .map_err(CreateClusterError::IoError)?;
+
+        let perms = std::fs::Permissions::from_mode(0o400);
+        tokio::fs::set_permissions(&lock_path, perms)
+            .await
+            .map_err(CreateClusterError::IoError)?;
+    }
```
emlautarom1: Unnecessary `map_err`

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+    let bytes = json.into_bytes();
+
+    for i in 0..num_nodes {
+        let lock_path = node_dir(cluster_dir.as_ref(), i).join("cluster-lock.json");
+
+        tokio::fs::write(&lock_path, &bytes)
+            .await
+            .map_err(CreateClusterError::IoError)?;
+
+        let perms = std::fs::Permissions::from_mode(0o400);
```
emlautarom1, line 1373: nit: we could use `set_readonly` here but it's fine either way. We should try be consistent throughout the codebase though.

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1481 @@
...
+    if zipped {
+        writeln!(w).map_err(CreateClusterError::IoError)?;
+        writeln!(w, "Files compressed and archived to:").map_err(CreateClusterError::IoError)?;
+        writeln!(w, "{}/cluster.tar.gz", abs_str).map_err(CreateClusterError::IoError)?;
+    }
+
+    Ok(())
+}
+
+fn write_split_keys_warning(w: &mut dyn Write) -> Result<()> {
```
emlautarom1: Unnecessary `map_err` in multiple places.

emlautarom1 (changes requested):
Overall OK, but there are a couple of problems:

- Missing tracing initialization
- The validation of arguments is done in a different order, potentially causing issues
- There is a missing address checksum
- A check in Charon that results in a warning is a hard error here.

The rests are nits or QoL changes that don't fundamentally affect the code.

emlautarom1:
I'd suggest also taking a look at Claude's findings, some of them are relevant.

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1528 @@
...
+    let def = pluto_cluster::definition::Definition::new(
+        name,
+        num_validators,
+        threshold,
+        fee_recipient_addrs,
+        withdrawal_addrs,
+        fork_version,
+        pluto_cluster::definition::Creator::default(),
+        operators,
+        args.deposit_amounts.clone(),
```
iamquang95: `eth2util::deposit::eths_to_gweis(&args.deposit_amounts)`

`args.deposit_amounts` is ETH, not gwei

crates/cli/Cargo.toml
```diff
@@ -19,10 +19,13 @@ humantime.workspace = true
 tokio.workspace = true
 pluto-app.workspace = true
 pluto-cluster.workspace = true
+pluto-crypto.workspace = true
 pluto-relay-server.workspace = true
 pluto-tracing.workspace = true
 pluto-core.workspace = true
 pluto-p2p.workspace = true
+pluto-eth1wrap.workspace = true
+pluto-eth2api.workspace = true
```
iamquang95, line 28: no usage

crates/cli/Cargo.toml
```diff
@@ -35,6 +38,11 @@ serde_with = { workspace = true, features = ["base64"] }
 rand.workspace = true
 tempfile.workspace = true
 reqwest.workspace = true
+url.workspace = true
+chrono.workspace = true
+uuid.workspace = true
```
iamquang95: no usage

crates/cli/Cargo.toml
```diff
@@ -35,6 +38,11 @@ serde_with = { workspace = true, features = ["base64"] }
...
+flate2.workspace = true
```
iamquang95: no usage

crates/cli/Cargo.toml
```diff
@@ -35,6 +38,11 @@ serde_with = { workspace = true, features = ["base64"] }
...
+tar.workspace = true
```
iamquang95: no usage

crates/cli/src/commands/create_cluster.rs
varex83: I didn't get this one, we are converting `eth` to `gwei`, as it's done in the go implementation.

iamquang95: `args.deposit_amounts` is input as `ETH` `List of partial deposit amounts (integers) in ETH`
But the `deposit_amounts` in `Definition` should be stored as gwei

iamquang95 (approved):
Please double check this comment: https://github.com/NethermindEth/pluto/pull/305#discussion_r3050744915

The rest is LGTM

crates/cli/src/commands/create_cluster.rs
```diff
@@ -0,0 +1,1519 @@
...
+        let pub_shares: Vec<Vec<u8>> = dv_priv_shares
+            .get(idx)
+            .map(|shares| {
+                shares
+                    .iter()
+                    .map(|share| tbls.secret_to_public_key(share))
+                    .collect::<std::result::Result<Vec<_>, _>>()
+            })
+            .transpose()?
+            .unwrap_or_default()
```
emlautarom1: If `dv_priv_shares` does not have `idx` we silently fail and return an empty vec of `pub_shares`. Instead, we should fail with an error and return (the go code panics in this scenario due to `dvPrivShares[idx]`)

emlautarom1 (approved):
LGTM, only a potential silent fail that should be addressed. Please, take a look at it before merging.

PR #308, opened 2026-04-02
feat: create dkg command

### LINKED ISSUE #304: Implement `cmd/createdkg`
### Description
Implement `pluto create dkg` following the reference files:

- `cmd/createdkg.go`
- `cmd/createenr_internal_test.go`

### Benefit

Gets us closer to full DKG compatibility, without requiring any actual `dkg/` modules to be implemented.

### Acceptance Criteria

mskrzypkows:
Implementation of the `pluto create dkg` command
fixes https://github.com/NethermindEth/pluto/issues/304


crates/cli/src/commands/create_dkg.rs
```diff
@@ -1,40 +1,475 @@
...
+//! configuration for a new Distributed Key Generation ceremony.
 
-use pluto_eth2util::{self as eth2util};
+use std::path::PathBuf;
+
+use k256::{SecretKey, elliptic_curve::rand_core::OsRng};
+use pluto_app::obolapi::{Client, ClientOptions};
+use pluto_cluster::{
+    definition::{Creator, Definition},
+    eip712sigs::{sign_cluster_definition_hash, sign_terms_and_conditions},
```
emlautarom1: nit: Avoid importing functions directly, makes reviewing harder.

crates/cli/src/commands/create_dkg.rs
```diff
@@ -1,40 +1,475 @@
...
+const MIN_THRESHOLD: u64 = 2;
+
+/// Arguments for the `pluto create dkg` command.
+#[derive(clap::Args)]
+#[command(
+    about = "Create the configuration for a new Distributed Key Generation ceremony using pluto dkg",
+    long_about = "Create a cluster definition file that will be used by all participants of a DKG."
+)]
+pub struct CreateDkgArgs {
+    /// The folder to write the output cluster-definition.json file to.
```
emlautarom1: These descriptions should be part of the argument `help` attribute.

For reference: https://github.com/NethermindEth/pluto/blob/7f52798db1ea6e94ffda8d97866c7aa8874ee822/crates/cli/src/commands/create_enr.rs#L23

crates/cli/src/commands/create_dkg.rs
```diff
@@ -1,40 +1,475 @@
...
+    Ok(run_create_dkg(parse_args(args)?).await?)
+}
+
+fn parse_args(args: CreateDkgArgs) -> Result<CreateDkgArgs, CreateDkgError> {
+    if args.threshold != 0 {
+        if args.threshold < MIN_THRESHOLD {
+            return Err(CreateDkgError::ThresholdTooLow);
+        }
+        let num_enrs = u64::try_from(args.operator_enrs.len())
+            .map_err(|_| CreateDkgError::OperatorCountOverflow)?;
```
emlautarom1: There is no need for this error variant, use `as` for these conversions (there is another similar instance).


```suggestion
        let num_enrs = args.operator_enrs.len() as u64;
```

crates/cli/src/commands/create_dkg.rs
```diff
@@ -1,40 +1,475 @@
...
+        if args.threshold > num_enrs {
+            return Err(CreateDkgError::ThresholdTooHigh);
```
emlautarom1: This error variant should contain `threshold` and `num_enrs` as fields.

crates/cli/src/commands/create_dkg.rs
```diff
@@ -1,40 +1,475 @@
...
+    }
+
+    if !consensus_protocol.is_empty() && !is_supported_protocol_name(consensus_protocol) {
+        return Err(CreateDkgError::UnsupportedConsensusProtocol);
+    }
+
+    Ok(())
+}
+
+fn validate_addresses(
```
emlautarom1: This is duplicated with `create_enr` with slight variations. We can:
- Extract them to a common module (preferable).
- Reference `create_cluster` from this module.

Note that there are other duplicated symbols like `MIN_NODES` and `MIN_THRESHOLD`

crates/cli/src/commands/create_dkg.rs
```diff
@@ -1,40 +1,475 @@
...
+
+    let (fee_recipient_addrs, withdrawal_addrs) = validate_addresses(
+        args.num_validators,
+        args.fee_recipient_addresses.clone(),
+        args.withdrawal_addresses.clone(),
+    )?;
+
+    validate_withdrawal_addrs(&withdrawal_addrs, &args.network)?;
+
+    info!("Pluto create DKG starting");
```
emlautarom1: This should actually log information from `pluto_core::version` (version, git commit & hash). We are missing a `log_info` function there (we can workaround the issue by inlining the code)

crates/cli/src/commands/create_dkg.rs
```diff
@@ -67,3 +503,350 @@ pub fn validate_withdrawal_addrs(
...
+
+fn safe_threshold(num_operators: u64) -> Result<u64, CreateDkgError> {
+    let two_n = num_operators
+        .checked_mul(2)
+        .ok_or(CreateDkgError::ThresholdOverflow)?;
+    Ok(two_n
+        .checked_add(2)
+        .ok_or(CreateDkgError::ThresholdOverflow)?
+        / 3)
+}
```
emlautarom1: Use the existing function: `pluto_cluster::helpers::threshold`

crates/cli/src/commands/create_dkg.rs
```diff
@@ -1,40 +1,475 @@
...
+            .map_err(|source| CreateDkgError::InvalidOperatorAddress { index: i, source })?;
+        operators.push(Operator {
+            address: checksum_addr,
+            ..Default::default()
+        });
+    }
+
+    let num_operators =
+        u64::try_from(operators.len()).map_err(|_| CreateDkgError::OperatorCountOverflow)?;
+    let safe_thresh = safe_threshold(num_operators)?;
```
emlautarom1: This does not look correct: `num_operators = args.operator_addresses.len() + args.operator_enrs.len()` when it should be `num_operators = operators_len`

crates/cli/src/commands/create_dkg.rs
```diff
@@ -1,40 +1,475 @@
...
 
-use crate::commands::create_cluster::{ZERO_ADDRESS, is_main_or_gnosis};
+    if let Some(key) = &priv_key {
+        def.creator.config_signature = sign_cluster_definition_hash(key, &def)?;
+    }
+
+    // Verify signatures when an ETH1 endpoint is available. Skipped when the
+    // endpoint is empty because the client cannot connect — safe for DKG create
+    // since operators have no signatures at this stage.
+    if !args.publish && !args.execution_engine_addr.is_empty() {
```
emlautarom1: We should not have the `!args.execution_engine_addr.is_empty()` check: if `args.publish == true` we must verify the signature. A missing `args.execution_engine_addr` should be an error (automatically propagated with `?`).

emlautarom1 (changes requested):
There are some flows that are not 100% matching with Charon. Consider adding tests for them if it can be done in a reasonable time.

crates/cli/src/commands/create_dkg.rs
mskrzypkows: `operator_enrs` or `operator_addresses` are mutually exclusive, the result is the same. I'll add a comment to describe it.

crates/cli/src/commands/create_dkg.rs
pinebit: `Publish` and `ExecutionEngineAddr` are unrelated. `Publish` controls sending the resulting lock file to Obol API, whereas `ExecutionEngineAddr` controls on-chain signature verification for gnosis/safe chains. In most cases (ethereum mainnet/testnets) this parameter is omitted.

crates/cli/src/commands/create_dkg.rs
```diff
@@ -67,3 +503,350 @@ pub fn validate_withdrawal_addrs(
...
+        .ok_or(CreateDkgError::ThresholdOverflow)?;
+    Ok(two_n
+        .checked_add(2)
+        .ok_or(CreateDkgError::ThresholdOverflow)?
+        / 3)
+}
+
+fn generate_launchpad_link(config_hash: &[u8], network: &str) -> String {
+    let network_prefix =
+        if network == HOLESKY.name || network == HOODI.name || network == SEPOLIA.name {
```
pinebit: Holesky is phased out, maybe it's good to go ahead and not support it explicitly. I will remove it from charon.

crates/cli/src/commands/create_dkg.rs
mskrzypkows: I found the source of the confusion, Charon implements noop eth client for empty execution_engine_addr, added similar behaviour to the Pluto.

crates/cli/src/commands/create_dkg.rs
mskrzypkows: removed from create_dkg.rs

mskrzypkows:
@emlautarom1 I've addressed all comments, please check.

crates/cli/src/commands/create_dkg.rs
pinebit: You may not need to implement any real eth1 client at this time (as well as EL sig verification), as this is very low priority feature, not a priority for your work, so please don't waste your time for the next milestone.

emlautarom1 (approved):
LGTM, not a fan of `EthClient::Noop` but it matches Charon closely so let's go with that.

It should be safe to drop Holesky (Charon 1.7.1 supports it) considering that the chain is deprecated.

PR #311, opened 2026-04-03
feat(dkg): implement sync

### LINKED ISSUE #133: Implement dkg/sync

iamquang95:
Should address https://github.com/NethermindEth/pluto/issues/133


iamquang95:
@claude

iamquang95:
> [High] Step is updated for peers that fail validation

Same as charon implementation 

> [Medium] Protocol ID has trailing slash — verify against Go

Matched with golang: `const protocolID = "/charon/dkg/sync/1.0.0/"`

> [Medium] await_all_at_step does not check started — can deadlock before Server::start()

Matched with charon, also it's not a bug.

iamquang95:
@claude now you will have the charon source code, review it again

iamquang95:
> [Low] Relay error detection may miss ErrResourceScopeClosed

Don't have the equivalent error in rust.

crates/dkg/src/sync/behaviour.rs
```diff
@@ -0,0 +1,345 @@
...
+
+    fn on_connection_handler_event(
+        &mut self,
+        _peer_id: PeerId,
+        _connection_id: ConnectionId,
+        event: THandlerOutEvent<Self>,
+    ) {
+        match event {
+            Either::Left(event) => self.pending_events.push_back(ToSwarm::GenerateEvent(event)),
+            Either::Right(unreachable) => match unreachable {},
```
varex83, line 226:
```suggestion
            Either::Right(_) => {},
```

crates/dkg/src/sync/handler.rs
```diff
@@ -0,0 +1,437 @@
...
+        &mut self,
+        event: ConnectionEvent<Self::InboundProtocol, Self::OutboundProtocol>,
+    ) {
+        match event {
+            ConnectionEvent::FullyNegotiatedInbound(FullyNegotiatedInbound {
+                protocol: mut stream,
+                ..
+            }) => {
+                stream.ignore_for_keep_alive();
+                self.inbound = Some(
```
varex83: Question: is there a chance that we will eventually receive 2 inbound requests and we will just drop one?

crates/dkg/src/sync/handler.rs
```diff
@@ -0,0 +1,437 @@
...
+
+                stream.ignore_for_keep_alive();
+                self.backoff = INITIAL_BACKOFF;
+                self.outbound = OutboundState::Running(run_outbound_stream(client, stream).boxed());
+            }
+            ConnectionEvent::DialUpgradeError(error) => self.on_dial_upgrade_error(error),
+            ConnectionEvent::AddressChange(_)
+            | ConnectionEvent::LocalProtocolsChange(_)
+            | ConnectionEvent::RemoteProtocolsChange(_) => {}
+            ConnectionEvent::ListenUpgradeError(_) => {}
```
varex83: We can shrink it into `_ => {}`

crates/dkg/src/sync/handler.rs
```diff
@@ -0,0 +1,437 @@
...
+    let mut stop_rx = client.stop_requested_rx();
+    let hash_signature = prost::bytes::Bytes::copy_from_slice(client.hash_sig());
+    let version = client.version().to_string();
+
+    client.set_connected(true);
+
+    loop {
+        tokio::select! {
+            _ = interval.tick() => {}
+            _ = stop_rx.changed() => return OutboundExit::Fatal(Error::Canceled),
```
varex83agent: **Bug: `stop_rx.changed()` fires on `activate()`'s `false` send, spuriously canceling the outbound stream**

`watch::Receiver::changed()` resolves on _any_ value change — including when `client.rs:178` sends `false` via `stop_tx.send(false)` during `activate()`. This means if `activate()` is ever called while this stream is running (e.g., reconnect after a transient stop), the stream will immediately exit with `Fatal(Error::Canceled)` before sending a single message.

The fix is to inspect the actual value after waking:

```rust
_ = stop_rx.changed() => {
    if *stop_rx.borrow_and_update() {
        return OutboundExit::Fatal(Error::Canceled);
    }
    // spurious wake from activate(); continue the loop
}
```

Same issue exists on line 319. Note the root cause is in `activate()` sending `false` unconditionally — contrast with Go's `chan struct{}` which is only ever closed once.

crates/dkg/src/sync/handler.rs
```diff
@@ -0,0 +1,437 @@
...
+    ) {
+        match event {
+            ConnectionEvent::FullyNegotiatedInbound(FullyNegotiatedInbound {
+                protocol: mut stream,
+                ..
+            }) => {
+                stream.ignore_for_keep_alive();
+                self.inbound = Some(
+                    handle_inbound_stream(
+                        self.peer_id,
```
varex83agent, line 303: **Bug: replacing `self.inbound` drops the old future without running `server.clear_connected`**

If a second `FullyNegotiatedInbound` event fires while a stream is already active (e.g., a peer reconnects before the old stream has been polled to completion), the old `BoxFuture` is dropped in-place. The `server.clear_connected(peer_id).await` at the end of `handle_inbound_stream` is never reached, so the peer remains permanently in `server.state.connected`. This prevents `await_all_connected` from ever completing for subsequent runs.

Go avoids this via `defer s.clearConnected(pID)` which runs unconditionally on goroutine exit. The Rust equivalent is a drop guard or an explicit check before overwriting:

```rust
ConnectionEvent::FullyNegotiatedInbound(FullyNegotiatedInbound { protocol: mut stream, .. }) => {
    if self.inbound.is_some() {
        warn!(peer = %self.peer_id, "dropping duplicate inbound stream");
        // stream drops here, rejecting the duplicate
        return;
    }
    stream.ignore_for_keep_alive();
    self.inbound = Some(handle_inbound_stream(...).boxed());
}
```

crates/dkg/src/sync/handler.rs
```diff
@@ -0,0 +1,437 @@
...
+                        self.outbound = OutboundState::Disabled;
+                        return Poll::Pending;
+                    };
+
+                    client.set_connected(false);
+                    client.release_outbound();
+
+                    if relay_reset || client.should_reconnect() {
+                        info!(peer = %self.peer_id, err = %error, "Disconnected from peer");
+                        self.outbound = OutboundState::Idle;
```
varex83agent: **Major: no backoff when transitioning to `Idle` after a `Reconnectable` exit**

When a running outbound stream terminates with `Reconnectable`, the handler goes to `OutboundState::Idle`. On the very next `poll()`, `try_request_outbound()` is called immediately — no delay at all. Under relay churn this creates a tight loop: `Running → Idle → OpenStream → Running → Reconnectable → Idle → ...`

Contrast with `on_dial_upgrade_error` (line 144) which always calls `self.schedule_retry()`. Go's client uses `expbackoff.WithFastConfig()` with `MaxDelay: 1s` on every reconnect attempt.

Fix:
```rust
if relay_reset || client.should_reconnect() {
    info!(peer = %self.peer_id, err = %error, "Disconnected from peer");
    self.backoff = INITIAL_BACKOFF; // fast first retry
    self.schedule_retry();
} else {
    client.finish(Err(error));
    self.outbound = OutboundState::Disabled;
}
```

crates/dkg/src/sync/handler.rs
```diff
@@ -0,0 +1,437 @@
...
+    ) {
+        let Some(client) = self.client.as_ref() else {
+            self.outbound = OutboundState::Disabled;
+            return;
+        };
+
+        client.release_outbound();
+
+        let (error, relay_reset) = match error {
+            StreamUpgradeError::NegotiationFailed => (Error::Unsupported, false),
```
varex83agent: **Major: `NegotiationFailed` retries indefinitely when `should_reconnect = true`**

`relay_reset = false` here, but since `should_reconnect()` defaults to `true` (line 19 of `client.rs`), the code at line 144 calls `self.schedule_retry()`. If the remote peer genuinely does not support `/charon/dkg/sync/1.0.0/`, every attempt will fail with `NegotiationFailed` and be retried forever (capped at 1 s intervals by `MAX_BACKOFF`).

Protocol negotiation failure is a definitive signal — the remote does not speak this protocol. Retrying is meaningless. Fix:

```rust
StreamUpgradeError::NegotiationFailed => {
    client.finish(Err(Error::Unsupported));
    self.outbound = OutboundState::Disabled;
    return;
}
```

This should be handled before the `relay_reset || should_reconnect` check.

crates/dkg/src/sync/server.rs
```diff
@@ -0,0 +1,243 @@
...
+
+            tokio::select! {
+                _ = cancellation.cancelled() => return Err(Error::Canceled),
+                _ = &mut notified => {}
+            }
+        }
+    }
+
+    /// Waits until all peers have reported shutdown.
+    pub async fn await_all_shutdown(&self, cancellation: CancellationToken) -> Result<()> {
```
varex83agent, line 100: **Major: `await_all_shutdown` does not check the server error state**

`await_all_connected` and `await_all_at_step` both check `state.err` at the top of their polling loop and return `Err` immediately if the server is poisoned. `await_all_shutdown` does not. If `set_err` is called (e.g., a peer fails validation) while a caller is blocking here, the function will loop indefinitely until the `CancellationToken` fires — potentially delaying DKG cleanup or key erasure.

```rust
pub async fn await_all_shutdown(&self, cancellation: CancellationToken) -> Result<()> {
    loop {
        let notified = self.inner.notify.notified();
        tokio::pin!(notified);
        {
            let state = self.inner.state.read().await;
            if let Some(error) = &state.err {
                return Err(error.clone()); // add this
            }
            if state.shutdown.len() == self.inner.all_count {
                return Ok(());
            }
        }
        ...
    }
}
```

crates/dkg/src/sync/handler.rs
```diff
@@ -0,0 +1,437 @@
...
+                    info!(
+                        peer = %peer_id,
+                        connected = count,
+                        expected = server.expected_peer_count(),
+                        "Connected to peer"
+                    );
+                }
+            }
+
+            if server.update_step(peer_id, message.step).await? {
```
varex83agent: **Note: `update_step` is called unconditionally after validation — this matches Go but creates a security concern**

This mirrors Go's `handleStream` where `updateStep` is called after the `validReq` if/else block regardless of validation outcome. So this is intentional parity.

However, it's worth being aware of the implication: a peer in the cluster whose signature is invalid (or wrong version) can still write step values into `server.state.steps` and fire `notify_waiters()`. While `await_all_at_step` checks `state.err` before returning `Ok`, the step entry is written and the slot in the `steps` HashMap is occupied by the rogue peer. If the legitimate peer and the rogue peer share a `PeerId` (which shouldn't happen in practice), the rogue write could block `await_all_at_step` from completing.

In practice this is constrained to cluster members with pre-known `PeerId`s, so the attack surface is limited. Still worth a comment explaining the intent.

crates/dkg/src/sync/handler.rs
```diff
@@ -0,0 +1,437 @@
...
+    inbound_events_tx: mpsc::UnboundedSender<OutEvent>,
+    mut stream: Stream,
+) -> Result<()> {
+    let result = async {
+        if !server.is_started() {
+            return Err(Error::ServerNotStarted);
+        }
+
+        let public_key = pluto_p2p::peer::peer_id_to_libp2p_pk(&peer_id)
+            .map_err(|error| Error::Peer(error.to_string()))?;
```
varex83agent, line 438: **Minor: missing idle timeout on inbound message reads — slow-loris risk**

The inbound read loop has no per-message timeout. A peer that connects and trickles bytes very slowly (e.g., sends the 8-byte length header but delays the payload) will hold this Tokio task open indefinitely. libp2p's connection-level idle timeout only fires on fully-idle connections, not slow-streaming ones.

Given that `MAX_MESSAGE_SIZE` is 128 MiB (see `proto.rs`), there is also a pre-allocation risk: a frame claiming 128 MiB but delivering nothing will hold 128 MiB allocated for the duration.

Consider wrapping each `read_fixed_size_protobuf` call with `tokio::time::timeout(2 * client_period, ...)` or a similar per-message deadline. Also consider adding a `SYNC_MAX_MESSAGE_SIZE` constant that matches Go's 32 MiB cap rather than reusing the shared 128 MiB constant.

crates/dkg/src/sync/protocol.rs
```diff
@@ -0,0 +1,49 @@
...
+
+use libp2p::identity::{Keypair, PublicKey};
+use pluto_core::version::SemVer;
+
+use crate::dkgpb::v1::sync::MsgSync;
+
+use super::error::{Error, Result};
+
+/// The protocol identifier for DKG sync.
+pub const PROTOCOL_NAME: libp2p::StreamProtocol = StreamProtocol::new("/charon/dkg/sync/1.0.0/");
```
varex83agent: nit: `use libp2p::StreamProtocol` appears two lines _below_ the constant that uses `StreamProtocol::new(...)`. Rust resolves this at compile time so it works, but it violates the convention of placing all `use` items before any non-use items. Move the `use` to the top of the file with the other imports.

crates/dkg/src/sync/error.rs
```diff
@@ -0,0 +1,102 @@
...
+    #[error("sign definition hash: {0}")]
+    SignDefinitionHash(String),
+
+    /// Failed to convert the local key to a libp2p keypair.
+    #[error("convert secret key to libp2p keypair: {0}")]
+    KeyConversion(String),
+
+    /// An I/O error occurred while reading or writing the stream.
+    #[error("i/o error: {0}")]
+    Io(String),
```
varex83agent, line 68: **Minor: `Io(String)` loses `io::ErrorKind`, diverges from sibling `bcast` module**

Storing `io::Error` as a `String` discards `ErrorKind`, which is used at multiple call sites to check `error.kind() == ConnectionReset` _before_ conversion (e.g., `handler.rs:139`, `handler.rs:326`). This produces fragile two-step code where the kind check must happen before the error is wrapped.

The sibling `bcast` module uses `Io(#[from] std::io::Error)`, preserving the kind through the type system. Aligning here would:
- Allow `?` instead of `.map_err(|e| Error::Io(e.to_string()))` throughout
- Enable `error.kind()` checks _after_ the error is wrapped
- Be consistent with the project pattern

```rust
#[error("i/o error: {0}")]
Io(#[from] std::io::Error),
```

crates/p2p/src/proto.rs
```diff
@@ -48,6 +48,60 @@ pub async fn read_length_delimited<S: AsyncRead + Unpin>(
...
+/// Reads a fixed-size length-delimited payload from the stream.
+pub async fn read_fixed_size_delimited<S: AsyncRead + Unpin>(
+    stream: &mut S,
+) -> io::Result<Vec<u8>> {
+    read_fixed_size_delimited_with_max(stream, MAX_MESSAGE_SIZE).await
+}
+
+/// Reads a fixed-size length-delimited payload from the stream, rejecting
+/// oversized messages.
+pub async fn read_fixed_size_delimited_with_max<S: AsyncRead + Unpin>(
```
varex83agent, line 74: **Minor: `MAX_MESSAGE_SIZE` is 128 MiB here vs Go's 32 MiB**

Go's `dkg/sync/server.go` defines `maxMessageSize = 32 * 1024 * 1024` (32 MiB). Rust uses the shared `MAX_MESSAGE_SIZE = 128 << 20` (128 MiB). This means:
- A Rust node will pre-allocate up to 128 MiB per frame, while Go rejects frames >32 MiB
- A Rust node sending a 50 MiB frame would be accepted locally but rejected by a Go peer

For DKG sync messages (a handful of bytes at most), this won't trigger in practice, but the protocol constants should match for correctness. Consider either reducing `MAX_MESSAGE_SIZE` to match Go, or passing a sync-specific cap:

```rust
// in sync/mod.rs or handler.rs
const SYNC_MAX_MESSAGE_SIZE: usize = 32 * 1024 * 1024; // matches Go
```
And use `read_fixed_size_delimited_with_max` directly.

crates/dkg/src/sync/client.rs
```diff
@@ -0,0 +1,277 @@
...
+            let _ = self.inner.done_tx.send(Some(result));
+        }
+    }
+
+    pub(crate) fn stop_requested_rx(&self) -> watch::Receiver<bool> {
+        self.inner.stop_tx.subscribe()
+    }
+
+    pub(crate) fn activate(&self) -> Result<()> {
+        self.inner.active.store(true, Ordering::SeqCst);
```
varex83agent: nit: All atomics in `ClientInner` use `Ordering::SeqCst` uniformly. `SeqCst` establishes a global total order across _all_ atomic operations on all threads, which is significantly more expensive on ARM (requires a full memory barrier) and unnecessary here — none of these flags participate in a multi-variable protocol that requires a total order.

Suggested orderings:
- `outbound_claimed.swap(...)` → `AcqRel` (classic test-and-set)
- `finished.swap(...)` → `AcqRel`
- `step` store/load → `Release`/`Acquire` (produces the current step for the send loop)
- All other standalone boolean flags → `Relaxed`

Same applies to `server.rs`'s `started` AtomicBool.

crates/dkg/src/sync/handler.rs
```diff
@@ -0,0 +1,437 @@
...
+                    return;
+                };
+
+                stream.ignore_for_keep_alive();
+                self.backoff = INITIAL_BACKOFF;
+                self.outbound = OutboundState::Running(run_outbound_stream(client, stream).boxed());
+            }
+            ConnectionEvent::DialUpgradeError(error) => self.on_dial_upgrade_error(error),
+            ConnectionEvent::AddressChange(_)
+            | ConnectionEvent::LocalProtocolsChange(_)
```
varex83agent: nit: The three named arms before the `_ => {}` catch-all are redundant. `ConnectionEvent` is `#[non_exhaustive]`, so the `_ => {}` arm is required and already covers `AddressChange`, `LocalProtocolsChange`, `RemoteProtocolsChange`, and `ListenUpgradeError`. Remove the named arms and keep only `_ => {}`. If the intent is to document which variants were consciously considered, a comment is clearer:

```rust
// AddressChange, LocalProtocolsChange, RemoteProtocolsChange,
// ListenUpgradeError: nothing to do
_ => {}
```

crates/dkg/src/sync/server.rs
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    pub(crate) async fn set_shutdown(&self, peer_id: PeerId) {
+        self.mutate_state(|state| {
+            state.shutdown.insert(peer_id);
+        })
+        .await;
+    }
+
+    pub(crate) async fn set_err(&self, error: Error) {
```
varex83agent, line 215: nit: `set_err` always overwrites — last-write-wins. This matches Go's `setErr` exactly, so it's parity-correct. However, first-error-wins is generally more useful for diagnostics (the root cause is usually the first error). Consider:

```rust
pub(crate) async fn set_err(&self, error: Error) {
    self.mutate_state(|state| {
        if state.err.is_none() {
            state.err = Some(error);
        }
    })
    .await;
}
```

This preserves the original error when multiple peers fail simultaneously.

crates/dkg/src/sync/server.rs
```diff
@@ -0,0 +1,243 @@
...
+        self.mutate_state(|state| {
+            state.err = Some(error);
+        })
+        .await;
+    }
+
+    pub(crate) async fn update_step(&self, peer_id: PeerId, step: i64) -> Result<bool> {
```
varex83agent, line 230: nit: `update_step` does `state.steps.get(&peer_id)` followed by `state.steps.insert(peer_id, step)` — two separate lookups. The Entry API avoids the double hash:

```rust
use std::collections::hash_map::Entry;
match state.steps.entry(peer_id) {
    Entry::Occupied(mut e) => {
        let current = *e.get();
        if step < current { return Err(Error::PeerStepBehind); }
        let current_plus_two = current.checked_add(2).ok_or(Error::StepOverflow)?;
        if step > current_plus_two { return Err(Error::PeerStepAhead); }
        if step == current { return Ok(false); }
        e.insert(step);
    }
    Entry::Vacant(e) => {
        if !(0..=1).contains(&step) { return Err(Error::AbnormalInitialStep); }
        e.insert(step);
    }
}
```

crates/dkg/src/sync/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+
+    struct RunningNode {
+        server: Server,
+        clients: Vec<Client>,
+        stop_tx: oneshot::Sender<()>,
+        join: tokio::task::JoinHandle<anyhow::Result<()>>,
+        client_joins: Vec<tokio::task::JoinHandle<Result<()>>>,
+    }
+
+    fn available_tcp_port() -> anyhow::Result<u16> {
```
varex83agent, line 186: nit: `available_tcp_port()` binds to `:0`, reads the port, then drops the listener. Between the drop and when libp2p binds that port, another process or concurrent test can claim it — a classic TOCTOU flake pattern.

The cleaner approach is to `listen_on` with `/ip4/127.0.0.1/tcp/0` and read the assigned address from the swarm's `NewListenAddr` event, which is how many libp2p examples handle this. Or keep the listener alive until the swarm has bound (using `SO_REUSEPORT`).

crates/dkg/examples/sync.rs
```diff
@@ -0,0 +1,735 @@
...
+impl ClusterInfo {
+    fn expected_connections(&self) -> usize {
+        self.peers.len().saturating_sub(1)
+    }
+
+    fn peer_label(&self, peer_id: &PeerId) -> String {
+        match self.indices.get(peer_id) {
+            Some(index) => format!(
+                "node={} peer_id={peer_id}",
+                index.checked_add(1).unwrap_or(*index)
```
varex83agent: nit: `index.checked_add(1).unwrap_or(*index)` — on overflow the fallback returns the zero-based index, silently displaying the wrong node number. While overflow is impossible in any realistic cluster, the fallback is semantically wrong. Use `saturating_add(1)` instead, which at least stays at `usize::MAX` rather than wrapping back to the original zero-based value.

varex83agent (dismissed):
## Review: feat(dkg): implement sync

This PR ports the Charon DKG sync protocol to Rust. The architecture is solid — the `Arc<Inner>` / `RwLock<ServerState>` / `Notify` design is correct, the wire format matches Go's i64 LE framing, and the step validation rules are faithful. Test coverage is good with both unit tests mirroring Go's test shapes and an integration round-trip test.

However there are **four bugs** that need to be fixed before merge:

1. **`stop_rx.changed()` spurious cancel** — `activate()` sends `false` on `stop_tx`, which fires `changed()` in the running outbound stream and kills it prematurely (`handler.rs:299,319` + `client.rs:178`)
2. **Inbound future replaced without cleanup** — dropping the old `BoxFuture` skips `server.clear_connected`, leaving the peer permanently in the connected set (`handler.rs:257`)
3. **No backoff on Reconnectable exit** — transitioning to `Idle` directly can spin-loop under relay churn; the `on_dial_upgrade_error` path correctly calls `schedule_retry()` but the stream-exit path does not (`handler.rs:226`)
4. **`NegotiationFailed` retries forever** — when a peer doesn't support the protocol, retrying with backoff is pointless; the handler should call `client.finish(Err(Unsupported))` (`handler.rs:125`)

Additionally, `await_all_shutdown` is missing the error-state check that `await_all_connected` and `await_all_at_step` both have, and the `MAX_MESSAGE_SIZE` constant (128 MiB) diverges from Go's 32 MiB cap.

Inline comments are filed for each issue individually.

crates/dkg/src/sync/handler.rs
iamquang95: I would keep the behaviour, will add comment on the code

crates/dkg/src/sync/handler.rs
iamquang95: `matches Go's 32 MiB` charon doesn't have max size

crates/dkg/src/sync/error.rs
iamquang95: Kept string because the we need `Clone` on Error

crates/p2p/src/proto.rs
iamquang95: hallucination? where did you get that :)) 
```
// readSizedProto reads a size prefixed proto message.
func readSizedProto(reader io.Reader, msg proto.Message) error {
	var size int64

	err := binary.Read(reader, binary.LittleEndian, &size)
	if err != nil {
		return errors.Wrap(err, "read size")
	}

	buf := make([]byte, size)

	n, err := reader.Read(buf)
	if err != nil {
		return errors.Wrap(err, "read buffer")
	} else if int64(n) != size {
		return errors.New("unexpected message length")
	}

	err = proto.Unmarshal(buf, msg)
	if err != nil {
		return errors.Wrap(err, "unmarshal proto")
	}

	return nil
}
```

crates/dkg/src/sync/client.rs
iamquang95: Keep SeqCst here for now.

crates/dkg/src/sync/server.rs
iamquang95: Decided to keep the behaviour

crates/dkg/src/sync/mod.rs
iamquang95: It's ok for testing purpose

crates/dkg/src/sync/behaviour.rs
iamquang95: Prefer to keep the code with `unreachable`. It's cleaner to read and expect this will never happen

crates/p2p/src/proto.rs
```diff
@@ -48,6 +48,60 @@ pub async fn read_length_delimited<S: AsyncRead + Unpin>(
...
+    stream.write_all(&len.to_le_bytes()).await?;
+    stream.write_all(payload).await?;
+    stream.flush().await
+}
+
+/// Reads a fixed-size length-delimited payload from the stream.
+pub async fn read_fixed_size_delimited<S: AsyncRead + Unpin>(
+    stream: &mut S,
+) -> io::Result<Vec<u8>> {
+    read_fixed_size_delimited_with_max(stream, MAX_MESSAGE_SIZE).await
```
pinebit, line 69: Can we align the `MAX_MESSAGE_SIZE` to charon's 32Mb?
https://github.com/ObolNetwork/charon/blob/97959c4e8ec0da2246dcf5166830159504245baa/dkg/sync/server.go#L35

pinebit: I know that the earlier version did not have any checks at all, but since we have added this already in newer releases (post-audit), then it makes sense to align the values from the beginning. It will probably be hard to find such the differences in the future.

iamquang95: Oh ok, my reference is v1.7.1 so I don't have that check. I will add it 👍

crates/dkg/examples/sync.rs
```diff
@@ -0,0 +1,719 @@
...
+//!   --relays https://0.relay.obol.tech,https://1.relay.obol.tech \
+//!   --data-dir /tmp/pluto-sync-demo/node2
+//! ```
+//!
+//! For stable local repros or CI-style runs, prefer a self-hosted relay
+//! instead of shared public relays.
+//!
+//! What to expect:
+//! - all nodes must use the same `--relays` value
+//! - `Relay reservation accepted`
```
emlautarom1, line 36: There is no `relay reservation accepted` message in the logs.

crates/dkg/examples/sync.rs
```diff
@@ -0,0 +1,719 @@
...
+//! - `All sync clients connected`
+//! - `Sync step reached`
+//! - `Sync demo is now idling until Ctrl+C`
+//! - press `Ctrl+C` on one node to stop it immediately and let the other nodes
+//!   observe the fault
```
emlautarom1, line 41: If you wait 60 seconds after `Sync client connectivity update` the peers disconnect by themselves too. I'm not sure if this is intentional and if so where it's set up.

I think it's related to idle timeouts (see: https://discuss.libp2p.io/t/rust-go-interop-idle-connection-timeout-race-condition/2470, https://github.com/libp2p/rust-libp2p/discussions/5741, https://github.com/libp2p/rust-libp2p/issues/5175) but it does not match the 300s default config that we use.

crates/dkg/examples/sync.rs
```diff
@@ -0,0 +1,719 @@
...
+    let mut node: Node<ExampleBehaviour> = Node::new(
+        p2p_config,
+        key,
+        NodeType::QUIC,
+        args.filter_private_addrs,
+        known_peers,
+        {
+            let p2p_context = p2p_context.clone();
+            move |builder, keypair, relay_client| {
+                let p2p_context = p2p_context.clone();
```
emlautarom1: This `p2p_context` is not needed. Remove the line entirely

crates/dkg/src/sync/error.rs
```diff
@@ -0,0 +1,102 @@
+use libp2p::PeerId;
+
+/// Sync result type.
+pub type Result<T> = std::result::Result<T, Error>;
+
+/// Error type for the DKG sync protocol.
+#[derive(Debug, Clone, thiserror::Error, PartialEq, Eq)]
```
emlautarom1: We don't need `PartialEq` or `Eq` so we can use `#[from]` in multiple errors instead of storing `String`.

crates/dkg/src/sync/server.rs
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    pub(crate) async fn mark_connected(&self, peer_id: PeerId) -> (bool, usize) {
+        let mut state = self.inner.state.write().await;
+        let inserted = state.connected.insert(peer_id);
+        let count = state.connected.len();
+        if inserted {
+            self.inner.notify.notify_waiters();
+        }
+        (inserted, count)
```
emlautarom1: Drop the lock before `notify_waiters` so waiters do not get immediately blocked:


```suggestion
        let (inserted, count) = {
            let mut state = self.inner.state.write().await;
            let inserted = state.connected.insert(peer_id);
            let count = state.connected.len();
            
            (inserted, count)
        };
        if inserted {
            self.inner.notify.notify_waiters();
        }
        (inserted, count)
```

emlautarom1: Check for similar patterns in other methods (ex. `clear_connected`), use the `mutate_state` helper (we could make it return a value too)

crates/dkg/src/sync/server.rs
```diff
@@ -0,0 +1,253 @@
...
+                }
+
+                if step == current {
+                    return Ok(false);
+                }
+
+                entry.insert(step);
+            }
+            Entry::Vacant(entry) => {
+                if !(0..=1).contains(&step) {
```
emlautarom1: Prefer a more straightforward check:

```suggestion
                if step < 0 || step > 1 {
```

crates/dkg/src/sync/server.rs
```diff
@@ -0,0 +1,253 @@
...
+
+    pub(crate) fn expected_peer_count(&self) -> usize {
+        self.inner.all_count
+    }
+
+    pub(crate) fn is_started(&self) -> bool {
+        self.inner.started.load(Ordering::SeqCst)
+    }
+
+    pub(crate) async fn mark_connected(&self, peer_id: PeerId) -> (bool, usize) {
```
emlautarom1: nit: use the same naming convention as in Charon

```suggestion
    pub(crate) async fn set_connected(&self, peer_id: PeerId) -> (bool, usize) {
```

crates/dkg/src/sync/client.rs
```diff
@@ -0,0 +1,292 @@
...
+    hash_sig: Vec<u8>,
+    version: SemVer,
+    period: Duration,
+    active: AtomicBool,
+    connected: AtomicBool,
+    reconnect: AtomicBool,
+    step: AtomicI64,
+    shutdown_requested: AtomicBool,
+    finished: AtomicBool,
+    outbound_claimed: AtomicBool,
```
emlautarom1: Charon uses a single lock for the entire client, while here we have multiple separate atomics, meaning that we cannot lock the entire structure and can have incorrect interleaving.

crates/dkg/src/sync/client.rs
```diff
@@ -0,0 +1,292 @@
...
+            if *current == stop_requested {
+                false
+            } else {
+                *current = stop_requested;
+                true
+            }
+        });
+    }
+
+    async fn wait_finished(
```
emlautarom1, line 258: The whole finished/shutdown/cleanup logic is quite hard to follow. It could use some comments or clarifications.

crates/dkg/src/sync/behaviour.rs
```diff
@@ -0,0 +1,345 @@
...
+            .peer_store_lock()
+            .connections_to_peer(&peer_id)
+            .is_empty()
+        {
+            return;
+        }
+
+        self.pending_events.push_back(ToSwarm::Dial {
+            opts: DialOpts::peer_id(peer_id)
+                .condition(PeerCondition::DisconnectedAndNotDialing)
```
emlautarom1, line 108: If we use `DisconnectedAndNotDialing` do we need the check on the `p2p_context` store or `client.is_connected()`? Seems redundant.

crates/dkg/src/sync/handler.rs
```diff
@@ -0,0 +1,475 @@
...
+            changed = stop_rx.changed() => {
+                if changed.is_err() || *stop_rx.borrow() {
+                    return OutboundExit::Fatal(Error::Canceled);
+                }
+                continue;
+            }
+        }
+
+        let shutdown = client.shutdown_requested();
+        let timestamp = Timestamp::from(std::time::SystemTime::now());
```
emlautarom1, line 360: Prefer tokio's time module

emlautarom1 (changes requested):
I find the logic inside the sync handler quite hard to grasp so I cannot fully evaluate it. The example works as intended except for dropping connections after 1 minute which could be attributed to some idle timeout but requires confirmation to proceed.

There are some potential concurrency issues due to using too fine grained atomics instead of locks over the entire structs which could cause issues later.

crates/dkg/src/sync/error.rs
iamquang95: Removed the `PartialEq` and `Eq`, but we still need the `Clone`, most of inner error is not cloneable (for example `std::io::Error`)

crates/dkg/src/sync/behaviour.rs
iamquang95: Removed the `p2p_context.store` check, but I will keep the `client.is_connected`. Since it's sync level protocol state check, sync and swarn run async so we should keep this check

crates/dkg/src/sync/handler.rs
iamquang95: This is used on the protobuf message, it requires a unix time now. So using `SystemTime` here is better suit.

crates/dkg/src/sync/server.rs
iamquang95: It violates clippy

```
warning: manual `!RangeInclusive::contains` implementation
   --> crates/dkg/src/sync/server.rs:247:24
    |
247 |                     if step < 0 || step > 1 {
    |                        ^^^^^^^^^^^^^^^^^^^^ help: use: `!(0..=1).contains(&step)`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.92.0/index.html#manual_range_contains
    = note: `#[warn(clippy::manual_range_contains)]` on by default
```

crates/dkg/src/sync/client.rs
iamquang95: Refactored using a single `std::io::RwLock<ClientState>`. The reason I don't use `tokio::sync::RwLock` here is that this data is heavily used inside libp2p `poll()`, which is a sync function. Using `blocking_read` or `blocking_write` brings more issues than benefits.

crates/dkg/examples/sync.rs
iamquang95: This is not caused by the sync client itself. The 60s disconnect comes from the local Pluto relay server used by the old example.

That relay config sets `max_circuit_duration` to one minute:
https://github.com/NethermindEth/pluto/blob/main/crates/relay-server/src/config.rs#L60 (I tested with different values, the disconnected will follow the config here)

The example now uses Obol public relays, matching the other examples, so it no longer exercises that local 60s circuit limit.

crates/dkg/src/sync/handler.rs
```diff
@@ -0,0 +1,474 @@
...
+
+        Poll::Pending
+    }
+
+    fn on_connection_event(
+        &mut self,
+        event: ConnectionEvent<Self::InboundProtocol, Self::OutboundProtocol>,
+    ) {
+        match event {
+            ConnectionEvent::FullyNegotiatedInbound(FullyNegotiatedInbound {
```
varex83agent, line 292: **Bug: `clear_connected()` is never called when the inbound future is dropped**

When a new inbound stream arrives while `self.inbound` already holds a live future, the old future is replaced and dropped. Because `server.clear_connected(peer_id)` lives at the very end of `handle_inbound_stream` (line 429) — outside and after the inner `async { ... }.await` block — it is never reached when the future is dropped mid-execution. The same happens if the connection closes while the stream is in progress.

Result: the peer stays in `server.connected` forever. If the peer then reconnects, `mark_connected` returns `inserted = false` and skips the notify, so `await_all_connected` could continue counting the ghost entry toward `all_count` while the actual connection is gone.

Fix: introduce a drop-guard that calls `server.clear_connected()` on drop, or restructure so the future cannot be silently abandoned (e.g., join the old future before replacing it, or call `clear_connected` from `ConnectionClosed` in the behaviour).

crates/dkg/src/sync/handler.rs
```diff
@@ -0,0 +1,474 @@
...
+
+        if !response.error.is_empty() {
+            return OutboundExit::Fatal(Error::PeerRespondedWithError(response.error));
+        }
+
+        if let Some(sync_timestamp) = response.sync_timestamp {
+            debug!(
+                peer = %client.peer_id(),
+                sync_timestamp = ?sync_timestamp,
+                "Received sync response"
```
varex83agent, line 416: **Major: 128 MB allocation limit for a message that is at most ~200 bytes**

`read_fixed_size_protobuf` delegates to `read_fixed_size_delimited_with_max(stream, MAX_MESSAGE_SIZE)` where `MAX_MESSAGE_SIZE = 128 << 20` (128 MB). `MsgSync` carries only a timestamp (12 bytes), a 64-byte secp256k1 signature, a version string (~6 bytes), a bool, and an int64. Its real upper bound is under 200 bytes.

Any authenticated DKG participant (or a peer with a corrupt implementation) can send a crafted 8-byte length header claiming 128 MB, causing the server to `vec![0_u8; 128 * 1024 * 1024]` before any decode check runs. In a 5-node cluster that would exhaust heap on a 512 MB node in a single round.

Fix: add a sync-specific cap of e.g. 4 KB and call `read_fixed_size_delimited_with_max(stream, 4096)` directly, or add a `read_fixed_size_protobuf_with_max` helper and use it here and in `run_outbound_stream`.

crates/dkg/src/sync/handler.rs
```diff
@@ -0,0 +1,474 @@
...
+        match &mut self.outbound {
+            OutboundState::Idle => {
+                if let Some(event) = self.try_request_outbound() {
+                    return Poll::Ready(event);
+                }
+            }
+            OutboundState::OpenStream => {}
+            OutboundState::WaitingRetry(delay) => {
+                if delay.as_mut().poll(cx).is_ready() {
+                    if let Some(event) = self.try_request_outbound() {
```
varex83agent: **Major: `WaitingRetry` → `Idle` fallback has no re-wakeup mechanism**

When the retry timer fires but `try_request_outbound()` returns `None` because `try_claim_outbound()` fails (another connection to the same peer holds the claim), the handler transitions back to `Idle`. In `Idle`, `poll()` only proceeds when the libp2p task wakes it for an incoming connection event.

If the claim-holding connection then closes (its stream finishes and calls `finish()` / `release_outbound()`), no notification is sent to this handler. It sits in `Idle` indefinitely — never re-attempting the outbound stream — until an unrelated connection event happens to wake it.

Fix: after `release_outbound()`, send a notification that re-wakes any Idle handlers for that peer (e.g. via a per-client `Notify`), or schedule a short-interval sleep in `Idle` to ensure periodic re-polling.

crates/dkg/src/sync/handler.rs
```diff
@@ -0,0 +1,474 @@
...
+            StreamUpgradeError::Timeout => (
+                Error::Io(
+                    std::io::Error::new(
+                        std::io::ErrorKind::TimedOut,
+                        "sync protocol negotiation timed out",
+                    )
+                    .to_string(),
+                ),
+                false,
+            ),
```
varex83agent, line 177: **Minor: relay-error detection covers only `ConnectionReset`, missing the `ErrResourceScopeClosed` equivalent**

Go's `isRelayError` triggers reconnection for two conditions: `network.ErrReset` (→ `io::ErrorKind::ConnectionReset`) **and** `network.ErrResourceScopeClosed`. When a relay circuit is recycled via resource-scope exhaustion, the I/O error kind is typically `BrokenPipe` or `ConnectionAborted`, not `ConnectionReset`. In that case `relay_reset` is `false` here and the reconnect decision falls through to `client.should_reconnect()`, which may be `false` for already-non-relay peers. The same issue exists at the `Reconnectable` arm in `run_outbound_stream` (line 326).

Fix: widen the check to also match `BrokenPipe` / `ConnectionAborted`, or thread the raw libp2p error through instead of converting to `io::Error` before this check.

crates/dkg/src/sync/behaviour.rs
```diff
@@ -0,0 +1,326 @@
...
+                        .clients
+                        .get(&peer_id)
+                        .is_some_and(|c| c.should_reconnect())
+                {
+                    self.schedule_dial_if_needed(peer_id);
+                }
+            }
+            _ => {}
+        }
+    }
```
varex83agent, line 216: **Minor: `DialFailure` retry is too narrow — only `DialError::Transport` triggers a reschedule**

Go's `connect()` loop retries on any error when `shouldReconnect()` is true. Rust only handles `DialError::Transport(_)`. `DialError::NoAddresses` is common during bootstrap before relay or mDNS has populated peer addresses; `DialError::DialBackoff` can fire when the swarm rate-limits outbound dials. Both would silently stop retrying and leave the sync client stuck without a connection.

Consider also retrying on `NoAddresses` (and optionally `DialBackoff`) with the same `schedule_dial_if_needed` call.

crates/dkg/src/sync/handler.rs
```diff
@@ -0,0 +1,474 @@
...
+            response = async {
+                match pluto_p2p::proto::write_fixed_size_protobuf(&mut stream, &request).await {
+                    Ok(()) => {
+                        pluto_p2p::proto::read_fixed_size_protobuf_with_max_size(
+                            &mut stream,
+                            protocol::MAX_MESSAGE_SIZE,
+                        )
+                        .await
+                    }
+                    Err(error) => Err(error),
```
varex83agent, line 380: **Minor: RTT is not recorded to the libp2p peerstore — parity gap with Go**

Go's `sendMsgs` records per-peer latency after every successful response:
```go
rtt := time.Since(resp.GetSyncTimestamp().AsTime())
c.p2pNode.Peerstore().RecordLatency(c.peer, rtt)
```
This data is used by the libp2p connection manager and routing subsystem to make informed decisions. The Rust implementation logs the timestamp as `debug` but never computes or records RTT. Threading a peerstore handle into the client or returning it through the event system would close this gap.

crates/dkg/src/sync/handler.rs
```diff
@@ -0,0 +1,474 @@
...
+                sync_timestamp = ?sync_timestamp,
+                "Received sync response"
+            );
+        }
+
+        if shutdown {
+            return OutboundExit::GracefulShutdown;
+        }
+    }
+}
```
varex83agent, line 424: **Minor: inbound loop continues after `set_err`, repeatedly re-notifying all barrier waiters**

After `validate_request_with_public_key` fails (line 373), `server.set_err()` is called and the response error field is populated — but the loop continues. Each subsequent message from the offending peer calls `mutate_state` (inside `set_err`) which calls `notify_waiters()`, waking all `await_all_*` callers every 100 ms. Each waiter sees the error, fails immediately, and the error state is refreshed again.

Closing the stream on validation failure (returning early after sending the error response) would terminate the relationship with the bad peer cleanly, matching the spirit of Go's `setErr` which is intended as a one-shot fatal signal.

crates/dkg/src/sync/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+                assert!(error.is_err(), "next step should not complete immediately");
+            }
+
+            for node in &running {
+                for client in &node.clients {
+                    client.set_step(step + 1);
+                }
+            }
+        }
+
```
varex83agent, line 508: **Minor: `client.shutdown()` calls have no timeout — test can hang indefinitely**

Every other barrier wait in this test (`await_all_connected`, `await_all_at_step`, `await_all_shutdown`) is wrapped with `tokio::time::timeout()`. The `client.shutdown()` calls at line 323 are not. If the shutdown handshake fails — e.g., the peer's swarm task has already exited, or a write error occurs before the shutdown flag is sent — `shutdown()` blocks forever on `wait_finished()` with a child cancellation token whose parent is never cancelled.

Wrap with `timeout(Duration::from_secs(10), client.shutdown(...)).await??`.

crates/dkg/src/sync/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+        );
+
+        assert!(
+            matches!(result, Err(Error::LocalPeerNotInPeerSet)),
+            "local peer must be part of the sync peer set"
+        );
+    }
+
+    #[tokio::test]
+    async fn sync_round_trip_matches_go_shape() -> anyhow::Result<()> {
```
varex83agent: **Minor: no integration tests for the error-propagation paths**

The integration test only exercises the happy path with matching versions and valid signatures. The Go test suite (`sync_test.go`) has explicit subtests for:
1. Version mismatch: two nodes with different `SemVer` values — asserts `AwaitAllConnected` returns an error containing `"mismatching charon version"`.
2. Invalid hash signature: tampered `hash_signature` field — asserts the validation error propagates.

These cover the entire `validate_request_with_public_key → server.set_err → await_all_connected` error path, which is currently untested. Add equivalent tests.

crates/dkg/src/sync/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+
+        Ok(running)
+    }
+
+    async fn shutdown_nodes(nodes: Vec<RunningNode>) -> anyhow::Result<()> {
+        for node in nodes {
+            let _ = node.stop_tx.send(());
+            node.join.await??;
+            for join in node.client_joins {
+                let _ = join.await;
```
varex83agent: nit: `let _ = join.await` silently discards client task errors. A client that panics or returns a non-`Canceled` error after `shutdown()` is swallowed. Propagate: `join.await??` (or collect and return).

crates/dkg/src/sync/client.rs
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+    /// Releases the outbound stream claim after the handler exits.
+    pub(crate) fn release_outbound(&self) {
+        self.write_state().outbound_claimed = false;
+    }
+
+    /// Completes the client once and publishes the result to all waiters.
+    pub(crate) fn finish(&self, result: Result<()>) {
+        let should_send = {
```
varex83agent, line 178: nit: `stop_tx.send(false)` in `activate()` always marks the channel as changed even when the value is already `false`. Any `stop_rx.changed()` waiter in a running stream would wake and see `*stop_rx.borrow() == false` — harmless today since `activate()` is only called before a stream runs, but fragile. Consider `stop_tx.send_if_modified(|v| mem::replace(v, false) != false)` to only notify when the value genuinely transitions from `true` → `false`.

varex83agent (dismissed):
This PR implements the DKG sync protocol (port of `charon/dkg/sync`) using a libp2p `NetworkBehaviour` + per-connection `ConnectionHandler`, with `Server`/`Client` Arc handles for external callers. The overall architecture is solid and the step-validation, version-checking, and signature-verification logic is functionally equivalent to the Go reference. There is one resource-lifecycle bug and two major concurrency/security issues that must be addressed before merge.

**Bugs (must fix):**
1. `clear_connected()` is never called when the inbound `BoxFuture` is replaced or the connection is dropped mid-stream — the peer stays in `server.connected` permanently, breaking `await_all_connected` after reconnect cycles.

**Major findings:**
2. `read_fixed_size_protobuf` uses the global 128 MB limit for `MsgSync`/`MsgSyncResponse` messages that are at most ~200 bytes — any authenticated peer can force a 128 MB heap allocation.
3. When `WaitingRetry` fires but `try_claim_outbound()` fails (a second connection holds the claim), the handler falls back to `Idle` with no waker registered — it can be permanently stuck until an unrelated connection event arrives.

**Minor findings:** relay-error detection misses `ErrResourceScopeClosed`; `DialFailure` only retries on `Transport` errors (too narrow); RTT is not recorded to the libp2p peerstore; no integration tests for the error-propagation paths (version mismatch → `set_err`); inbound stream continues after `set_err` refreshing `notify_waiters` at 100 ms; `client.shutdown()` in the integration test has no timeout.

Verdict: **REQUEST_CHANGES** — the `clear_connected` lifecycle bug can cause silent test hangs and incorrect barrier semantics; the 128 MB allocation vector is a DoS risk in a cooperative but potentially adversarial network; and the stuck-Idle handler is a correctness issue in multi-connection scenarios.

crates/dkg/src/sync/handler.rs
iamquang95: https://github.com/ObolNetwork/charon/blob/97959c4e8ec0da2246dcf5166830159504245baa/dkg/sync/server.go#L35

Set as 32MB, according to the audit report

crates/dkg/src/sync/handler.rs
iamquang95: Solved by adding `PeerRttObserved` event

varex83 (approved):
LGTM!

PR #313, opened 2026-04-06
fix(ci): fetch charon repository before run code review

iamquang95:
Since claude doesn't have permission to use `WebFetch` and `gh` on CI, we can pull the charon source code before doing the claude review as a Golang implementation references.

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -30,6 +30,20 @@ jobs:
         with:
           fetch-depth: 1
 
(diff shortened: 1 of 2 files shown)


varex83 (approved):

varex83agent (approved):
CI infrastructure change is correct and minimal. The cache+clone logic is sound — validates the git dir and exact tag before re-cloning, and prints HEAD for auditability. Both workflows updated symmetrically. LGTM.

emlautarom1 (approved):

Issue #317, opened 2026-04-10
`test_random_bit_list` is flanky


iamquang95:
### Description
`random_bit_list` is not guarantee the length of the output, which causes the test `test_random_bit_list` flanky.

PR #318, opened 2026-04-10
fix(random): random_bit_list should set exact length

iamquang95:
Should fix this: https://github.com/NethermindEth/pluto/issues/317


iamquang95:
@claude

varex83 (approved):

emlautarom1 (approved):

PR #320, opened 2026-04-13
fix: update rand, ignore the RUSTSEC-2026-0097 for now

iamquang95:
**Ignore RUSTSEC-2026-0097 (rand unsoundness)**

  `cargo deny check` started failing on RUSTSEC-2026-0097, an unsoundness advisory in rand versions >= 0.7, < 0.9.3.

**Changes**

  - Bumped rand 0.9.2 → 0.9.3 via cargo update (fixes the one instance we can patch in-place).
  - Added RUSTSEC-2026-0097 to deny.toml ignore list with justification. Two transitive copies remain pinned upstream and cannot be bumped from here:
    - rand 0.7.3 via cuckoofilter → libp2p-floodsub
    - rand 0.8.5 via alloy-signer-local (latest release)

**Why it's safe to ignore**

  - Unsoundness only, no known exploit. No CVSS score; the advisory documents UB under Stacked Borrows, not a remote-exploitable vulnerability.
  - Trigger path is not reachable from Pluto. It requires all of: a custom log logger that calls rand::rng(), hitting the 64KB reseed boundary mid-log-call, at trace level (or warn + getrandom failure). Pluto
   uses tracing and does not invoke rand from any logger.
  - Nothing actionable upstream-side yet. libp2p-floodsub and alloy-signer-local would need to release with rand >= 0.9.3 before we can drop the ignore.

**Follow-up**

  Remove the ignore entry once cuckoofilter/libp2p-floodsub and alloy-signer-local upgrade past rand 0.9.3.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -379,7 +379,7 @@ dependencies = [
  "keccak-asm",
  "paste",
  "proptest",
- "rand 0.9.2",
+ "rand 0.9.3",
  "rapidhash",
  "ruint",
  "rustc-hash",
@@ -1398,7 +1398,7 @@ dependencies = [
  "log",
  "num",
  "pin-project-lite",
- "rand 0.9.2",
+ "rand 0.9.3",
  "rustls",
  "rustls-native-certs",
  "rustls-pki-types",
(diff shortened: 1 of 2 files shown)


varex83 (approved):

iamquang95:
Tracked issue here: https://github.com/NethermindEth/pluto/issues/323

PR #321, opened 2026-04-13
feat(dkg): Port node signature exchange from Charon

### LINKED ISSUE #256: Implement dkg/nodesigs
### Description
Implement dkg/nodesigs

### Acceptance criteria

mskrzypkows:
- Port `dkg/nodesigs.go` from Charon — implements K1 signature exchange over the lock hash using the bcast protocol
- `NodeSigBcast` registers a typed broadcast handler for `MsgNodeSig`, then `exchange()` signs the lock hash, broadcasts it to all peers, and polls until every peer's verified signature is collected

#256


iamquang95:
@claude

mskrzypkows:
Fixing claude findings, both seems relevant.

mskrzypkows:
@claude

mskrzypkows:
Hey @iamquang95 I have modified the `CallbackFn` interface. Could you review it?

crates/dkg/src/nodesigs.rs
```diff
@@ -0,0 +1,479 @@
...
+            let mut sigs = self.sigs.lock().unwrap_or_else(|e| e.into_inner());
+            sigs[self.node_idx] = Some(local_sig.to_vec());
+        }
+
+        let mut ticker = tokio::time::interval(Duration::from_millis(100));
+
+        loop {
+            tokio::select! {
+                () = token.cancelled() => return Err(Error::Cancelled),
+                _ = ticker.tick() => {
```
iamquang95, line 144: nit: since this method and `NodeSigBcast` are public, the user can reuse the `NodeSigBcast`. On the first `exchange` call, everything is good. But on the second call, all `sig` from the first run is still there, only `local_sig` is updated, and in the loop below, the result will be returned immediately. 
Better to clear the `sig` on the beginning of this function

crates/dkg/src/nodesigs.rs
```diff
@@ -0,0 +1,479 @@
...
+    /// Returns all collected signatures ordered by peer index.
+    pub async fn exchange(
+        &self,
+        key: &SecretKey,
+        lock_hash: impl AsRef<[u8]>,
+        token: CancellationToken,
+    ) -> Result<Vec<Vec<u8>>> {
+        let lock_hash = lock_hash.as_ref();
+
+        let local_sig = pluto_k1util::sign(key, lock_hash)?;
```
iamquang95: `to_vec` here to avoid double `to_vec` later

crates/dkg/src/nodesigs.rs
iamquang95: Or better to not reuse the `nodesig`, change to `self` instead of `&self`

mskrzypkows: good point

crates/dkg/src/nodesigs.rs
```diff
@@ -0,0 +1,479 @@
...
+/// Validates and stores an incoming node signature message.
+///
+/// Waits for the lock hash to become available via the watch channel before
+/// verifying the signature.
+async fn receive(
+    peer_id: PeerId,
+    msg: MsgNodeSig,
+    node_idx: usize,
+    peers: &[Peer],
+    lock_hash_rx: &watch::Receiver<Option<Vec<u8>>>,
```
varex83agent: `lock_hash_rx` is taken as `&watch::Receiver<…>` but immediately cloned on the very next line inside `receive()`. Since `watch::Receiver` is cheap to clone and the function already owns the clone it needs, prefer passing by value:

```rust
async fn receive(
    …
    lock_hash_rx: watch::Receiver<Option<Vec<u8>>>,
    …
) {
    let mut rx = lock_hash_rx;  // no clone needed
    …
}
```
This removes an unnecessary indirection and makes the ownership intent explicit.

crates/dkg/src/nodesigs.rs
```diff
@@ -0,0 +1,479 @@
...
+            peer_index,
+        };
+
+        tracing::debug!("Exchanging node signatures");
+
+        self.bcast.broadcast(NODE_SIG_MSG_ID, &bcast_data).await?;
+
+        {
+            let mut sigs = self.sigs.lock().unwrap_or_else(|e| e.into_inner());
+            sigs[self.node_idx] = Some(local_sig.to_vec());
```
varex83agent: `local_sig` is already `Vec<u8>`. Calling `.to_vec()` on it here creates a redundant clone. Since `local_sig` was already partially moved into `bcast_data.signature` via `.clone().into()` on line 110, you can just move it here:

```rust
sigs[self.node_idx] = Some(local_sig);  // move, no extra allocation
```

crates/dkg/src/nodesigs.rs
```diff
@@ -0,0 +1,479 @@
...
+async fn receive(
+    peer_id: PeerId,
+    msg: MsgNodeSig,
+    node_idx: usize,
+    peers: &[Peer],
+    lock_hash_rx: &watch::Receiver<Option<Vec<u8>>>,
+    sigs: &Mutex<Vec<Option<Vec<u8>>>>,
+) -> bcast::Result<()> {
+    let peer_idx =
+        usize::try_from(msg.peer_index).map_err(|_| bcast::Error::InvalidPeerIndex(peer_id))?;
```
varex83agent: `usize::try_from(u32)` is infallible on all supported platforms (both 32-bit and 64-bit): a `u32` always fits in a `usize`. The `.map_err(…)` error path is dead code. A simple `as usize` cast communicates the intent more clearly:

```rust
let peer_idx = msg.peer_index as usize;
```

varex83agent (commented):
This PR ports `dkg/nodesigs.go` to Rust with solid functional parity. The core K1 signature-exchange logic is correctly translated: `watch::channel` replacing Go's unbuffered channel, `CancellationToken` replacing `ctx.Done()`, and the `CallbackFn` upgrade to async with proper `'static` ownership semantics. No `std::sync::Mutex` is held across `.await` points; the struct is intentionally single-use via `exchange(self, …)`. No bugs or major findings were identified.

**Minor findings (non-blocking but worth addressing):**
1. `lock_hash_rx` parameter in `receive()` is borrowed then immediately cloned — should be passed by value.
2. `local_sig.to_vec()` is called twice, creating an unnecessary extra clone.
3. `usize::try_from(u32)` in `receive()` is infallible on all supported platforms, making the error path dead code.
4. `receive()` awaits `wait_for` with no timeout, holding the bcast substream open if a message arrives before `exchange()` sets the lock hash.

**Nits:** Tokio `interval` fires immediately at t=0 (unlike Go's `NewTicker`), `sigs_cb` naming, `all_sigs` double-pass iteration.

crates/dkg/src/nodesigs.rs
```diff
@@ -0,0 +1,479 @@
...
+
+    if peer_idx == node_idx || peer_idx >= peers.len() {
+        return Err(bcast::Error::InvalidPeerIndex(peer_id));
+    }
+
+    let pubkey = peers[peer_idx].public_key()?;
+
+    let lock_hash = {
+        let mut rx = lock_hash_rx.clone();
+        let guard = rx
```
varex83agent: `receive()` awaits `lock_hash_rx.wait_for(…)` with no timeout. If a peer sends a `MsgNodeSig` before `exchange()` has been called (which sets the lock hash), this future blocks indefinitely, holding the inbound bcast substream open for that connection.

In Go, `lockHash(ctx)` respects `ctx.Done()` so the wait is bounded by the caller's context. The Rust version only unblocks when the `watch::Sender` is dropped (when `NodeSigBcast` is dropped or `exchange()` returns). While not a correctness bug in the happy path, consider wrapping with a short timeout or adding a comment explaining the ordering guarantee:

```rust
// Safety: exchange() sends the lock hash before broadcasting, so by the time
// peers respond with their signatures this channel always has a value.
let guard = rx.wait_for(|v| v.is_some()).await
    .map_err(|_| bcast::Error::MissingField("lock_hash"))?;
```

crates/dkg/src/nodesigs.rs
```diff
@@ -0,0 +1,479 @@
...
+        tracing::debug!("Exchanging node signatures");
+
+        self.bcast.broadcast(NODE_SIG_MSG_ID, &bcast_data).await?;
+
+        {
+            let mut sigs = self.sigs.lock().unwrap_or_else(|e| e.into_inner());
+            sigs[self.node_idx] = Some(local_sig.to_vec());
+        }
+
+        let mut ticker = tokio::time::interval(Duration::from_millis(100));
```
varex83agent, line 139: nit: `tokio::time::interval` fires its **first tick immediately** (at t=0), whereas Go's `time.NewTicker(100ms)` waits the full period before the first tick. This is a minor parity difference — the Rust exchange loop performs one extra eager check on startup — but it's harmless. Consider documenting this if it surprises future readers, or use `MissedTickBehavior::Delay` to match Go's behaviour.

crates/dkg/src/nodesigs.rs
```diff
@@ -0,0 +1,479 @@
...
+    lock_hash_tx: watch::Sender<Option<Vec<u8>>>,
+}
+
+impl NodeSigBcast {
+    /// Returns a new instance, registering bcast handlers on `bcast_comp`.
+    pub async fn new(peers: Vec<Peer>, node_idx: usize, bcast_comp: Component) -> Result<Self> {
+        let sigs = Arc::new(Mutex::new(vec![None::<Vec<u8>>; peers.len()]));
+        let (lock_hash_tx, lock_hash_rx) = watch::channel(None::<Vec<u8>>);
+
+        let sigs_cb = Arc::clone(&sigs);
```
varex83agent, line 73: nit: `sigs_cb` is a non-idiomatic name (the `_cb` suffix isn't standard and implies it's only used in the callback, which future readers may not find obvious). Consider simply shadowing `sigs`:

```rust
let sigs = Arc::clone(&sigs);
```
This is a common Rust pattern for cloning an `Arc` before a move closure.

crates/dkg/src/nodesigs.rs
```diff
@@ -0,0 +1,479 @@
...
+                        return Ok(all);
+                    }
+                }
+            }
+        }
+    }
+}
+
+/// Returns a copy of all signatures if every slot is filled, otherwise `None`.
+fn all_sigs(sigs: &[Option<Vec<u8>>]) -> Option<Vec<Vec<u8>>> {
```
varex83agent, line 155: nit: `all_sigs` performs two passes over the slice — first `iter().all(is_some)`, then `iter().flatten().cloned().collect()`. A single-pass collect via `Iterator::map` + `collect::<Option<Vec<_>>>` is cleaner:

```rust
fn all_sigs(sigs: &[Option<Vec<u8>>]) -> Option<Vec<Vec<u8>>> {
    sigs.iter().map(|s| s.clone()).collect()
}
```
`collect::<Option<Vec<_>>>()` returns `None` if any element is `None`, and `Some(vec)` if all are `Some` — exactly the desired semantics in one pass.

crates/dkg/src/nodesigs.rs
```diff
@@ -0,0 +1,479 @@
...
+    peer_id: PeerId,
+    msg: MsgNodeSig,
+    node_idx: usize,
+    peers: &[Peer],
+    lock_hash_rx: &watch::Receiver<Option<Vec<u8>>>,
+    sigs: &Mutex<Vec<Option<Vec<u8>>>>,
+) -> bcast::Result<()> {
+    let peer_idx =
+        usize::try_from(msg.peer_index).map_err(|_| bcast::Error::InvalidPeerIndex(peer_id))?;
+
```
varex83agent, line 177: nit: The `peer_id` argument (P2P sender identity) is not cross-checked against `peers[peer_idx].id` (the expected identity for that slot). The same gap exists in the Go implementation, and the K1 signature verification provides the real security guarantee here. Consider a defence-in-depth comment explaining that the P2P transport authenticates the sender and the signature check covers authenticity, so the cross-check is intentionally omitted.

crates/dkg/src/nodesigs.rs
varex83: Will be against linting rules, but we can use `expect` anyway

crates/dkg/src/nodesigs.rs
mskrzypkows: but `sigs` is used later in line 80

crates/dkg/src/nodesigs.rs
mskrzypkows: it's not a problem to check it earlier, let's keep it as it is

mskrzypkows:
@varex83 @varex83agent fixed all comments

crates/dkg/src/nodesigs.rs
```diff
@@ -0,0 +1,503 @@
...
+fn all_sigs(sigs: &[Option<Vec<u8>>]) -> Option<Vec<Vec<u8>>> {
+    sigs.iter().cloned().collect()
+}
+
+/// Validates and stores an incoming node signature message.
+///
+/// Waits for the lock hash to become available via the watch channel before
+/// verifying the signature. Returns [`bcast::Error::Cancelled`] if `token` is
+/// cancelled while waiting.
+async fn receive(
```
pinebit, line 167: In `receive` it must check sender ID. This was a recent audit finding, please consider adding the check: https://github.com/ObolNetwork/charon/blob/97959c4e8ec0da2246dcf5166830159504245baa/dkg/nodesigs.go#L138

crates/dkg/src/nodesigs.rs
```diff
@@ -0,0 +1,503 @@
...
+            lock_hash_tx,
+        })
+    }
+
+    /// Exchanges K1 signatures over the lock hash with all peers.
+    ///
+    /// Signs `lock_hash` with `key`, broadcasts the signature to all peers, and
+    /// polls until every peer's signature has been received and verified.
+    /// Returns all collected signatures ordered by peer index.
+    pub async fn exchange(
```
pinebit, line 105: In Go's version we have concept of `none` data https://github.com/ObolNetwork/charon/blob/97959c4e8ec0da2246dcf5166830159504245baa/dkg/nodesigs.go#L208
that is used for re-dkg protocols.
Maybe, if it is easier to introduce this from day zero, it will be easier to not return to this component in the future (otherwise it will be hard to find out why it does not work).

crates/dkg/src/nodesigs.rs
mskrzypkows: added the check

crates/dkg/src/nodesigs.rs
mskrzypkows: added none data

crates/dkg/src/bcast/error.rs
```diff
@@ -46,6 +46,25 @@ impl Failure {
...
+    pub sender: PeerId,
+    /// The peer ID expected at the claimed index.
+    pub expected: PeerId,
+}
+
+impl fmt::Display for SenderPeerMismatch {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "sender peer ID ({}) does not match claimed peer index {}",
```
iamquang95: `peer index` -> `peer ID`

crates/dkg/src/nodesigs.rs
```diff
@@ -0,0 +1,615 @@
...
+/// protocol.
+pub struct NodeSigBcast {
+    sigs: Arc<Mutex<Vec<Option<Vec<u8>>>>>,
+    bcast: Component,
+    node_idx: usize,
+    lock_hash_tx: watch::Sender<Option<Vec<u8>>>,
+}
+
+impl NodeSigBcast {
+    /// Returns a new instance, registering bcast handlers on `bcast_comp`.
```
iamquang95, line 61: add comment here, the `bcast_comp` can't be reused, since calling `register_message` on other `NodeSigBcast::new` with the same `bcast_comp` will return error.

iamquang95 (approved):
other part lg-tm

PR #322, opened 2026-04-13
feat: add SSZ encoding/decoding support for parsigex

varex83agent:
## Summary

- Adds low-level SSZ encode/decode helpers (`crates/ssz/src/encode.rs`, `crates/ssz/src/decode.rs`)
- Extends `crates/ssz/src/types.rs` with SSZ type implementations for eth2api spec types
- Derives SSZ traits on `crates/eth2api/src/spec/*` types (phase0, altair, bellatrix, capella, deneb, electra, fulu)
- Adds `crates/core/src/ssz_codec.rs` — SSZ codec integration for parsigex signed data
- Extends `parsigex_codec.rs` with SSZ-based serialisation alongside existing JSON path
- Adds SSZ test fixtures under `crates/core/testdata/ssz/` generated by `test-infra/sszfixtures/`
- Fixes yanked `fastrand 2.4.0` → `2.4.1` and ignores pre-existing `rand 0.8.5` advisory (RUSTSEC-2026-0097) present on base branch

## Test plan

- [ ] `cargo test --workspace --all-features` passes
- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings` passes
- [ ] SSZ round-trip tests in `crates/core` exercise the new fixtures
- [ ] `cargo deny check` passes


iamquang95:
@claude

iamquang95:
remember to remove golang code

varex83:
@claude

varex83:
@claude review

crates/ssz/Cargo.toml
```diff
@@ -7,6 +7,7 @@ license.workspace = true
 publish.workspace = true
 
 [dependencies]
+ethereum_ssz = { workspace = true }
```
emlautarom1: nit:
```suggestion
ethereum_ssz.workspace = true
```

Cargo.toml
```diff
@@ -93,6 +93,8 @@ oas3-gen-support = "0.24"
 bon = "3.8"
 testcontainers = "0.27"
 test-case = "3.3"
+ethereum_ssz = { version = "0.10" }
+ethereum_ssz_derive = { version = "0.10" }
```
emlautarom1, line 97: We had some discussion in the past due to using multiple ssz libraries but I don't remember the details. Why are we adding this dependency?

crates/ssz/src/lib.rs
```diff
@@ -16,3 +18,19 @@ pub use helpers::{
...
+    #[error("invalid length: expected {expected}, got {actual}")]
+    InvalidLength {
+        /// Expected byte count.
+        expected: usize,
+        /// Actual byte count.
+        actual: usize,
+    },
+    /// Generic decoding error.
+    #[error("{0}")]
+    Custom(String),
```
emlautarom1: nit: We only use this for "invalid bool bytes", better to have a proper variant for that case.

crates/ssz/src/types.rs
```diff
@@ -145,6 +177,59 @@ impl<const SIZE: usize> AsRef<[u8]> for SszVector<u8, SIZE> {
...
+        }
+    }
+
+    fn from_ssz_bytes(bytes: &[u8]) -> Result<Self, DecodeError> {
+        let values = Vec::<T>::from_ssz_bytes(bytes)?;
+        if values.len() != SIZE {
+            return Err(DecodeError::BytesInvalid(format!(
+                "vector length {} does not match required {SIZE}",
+                values.len(),
+            )));
```
emlautarom1: nit: There are better error variants that we can use:

```suggestion
            return Err(DecodeError::InvalidByteLength {
                len: values.len(),
                expected: SIZE,
            });
```

crates/core/src/ssz_codec.rs
```diff
@@ -0,0 +1,762 @@
...
+
+// ===========================================================================
+// Versioned type helpers
+// ===========================================================================
+
+fn encode_version(version: DataVersion) -> Result<[u8; 8], SszCodecError> {
+    version
+        .to_legacy_u64()
+        .map(encode_u64)
+        .map_err(|_| SszCodecError::UnknownVersion(0))
```
emlautarom1: Version `Phase0` maps to `0` through `to_legacy_u64`. The error is misleading.

emlautarom1 (approved):
LGTM, the code is quite error prone due to repetition so I suggest taking a second look at Claude's output.

crates/core/src/ssz_codec.rs
```diff
@@ -0,0 +1,762 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// VersionedAttestation
+// Header: version(8) + offset(4) = 12 bytes (same as other versioned types)
+// ---------------------------------------------------------------------------
+
+/// Encodes a `VersionedAttestation` to SSZ binary with Charon versioned
+/// header.
+pub fn encode_versioned_attestation(
```
iamquang95, line 176: This seem not cover the same ssz in charon
```
// MarshalSSZTo ssz marshals the VersionedAttestation object to a target array.
func (a VersionedAttestation) MarshalSSZTo(dst []byte) ([]byte, error) {
	version, err := eth2util.DataVersionFromETH2(a.Version)
	if err != nil {
		return nil, errors.Wrap(err, "invalid version")
	}

	if a.ValidatorIndex == nil {
		return marshalSSZVersionedTo(dst, version, a.sszValFromVersion)
	}

	valIdx := *a.ValidatorIndex

	return marshalSSZVersionedValidatorIdxTo(dst, version, valIdx, a.sszValFromVersion)
}
```

First need to check the `ValidatorIndex` to decide which ssz encoding needs to be used.

iamquang95:
```
func marshalSSZVersionedValidatorIdxTo(dst []byte, version eth2util.DataVersion, valIdx eth2p0.ValidatorIndex, valFunc func(eth2util.DataVersion) (sszType, error)) ([]byte, error) {
	// Field (0) 'Version'
	dst = ssz.MarshalUint64(dst, version.ToUint64())

	// Field (1) 'ValidatorIndex'
	dst = ssz.MarshalUint64(dst, uint64(valIdx))

	// Offset (2) 'Value'
	dst = ssz.WriteOffset(dst, versionedValIdxOffset)

	// TODO(corver): Add a constant length data version string field, ensure this is backwards compatible.

	val, err := valFunc(version)
	if err != nil {
		return nil, errors.Wrap(err, "sszValFromVersion from version")
	}

	// Field (2) 'Value'
	if dst, err = val.MarshalSSZTo(dst); err != nil {
		return nil, errors.Wrap(err, "marshal sszValFromVersion")
	}

	return dst, nil
}
```

It's `8+8+4 = 20` bytes if there is `VersionIndex`

iamquang95: Also should have golden test for this (both case with and without `VersionIndex`)

Cargo.toml
varex83: We add it to easily allow encoding/decoding of some structures via derive, we still can write derive from scratch, but I think this is a good use case. I think previous discussions were about tree hash

varex83:
@claude review it please

varex83:
@claude review

iamquang95 (approved):
LGTM

iamquang95:
For these `BitList`, i guess we can evaluate it and address in another PR. To me, it's med-/low issues instead of high

PR #324, opened 2026-04-14
fix: increase deadliner test timing

pinebit:
## Summary
- The `test_deadliner` test used **10ms-per-slot** deadline margins, creating a race condition: wall-clock time elapsed between capturing `start_time` and the background task processing `add()` could exceed the deadline on loaded CI runners, making "non-expired" duties appear expired.
- Increased deadline margins to **500ms per slot** — enough headroom for scheduling jitter while keeping the test fast (~1-2s).
- Raised the output collection timeout from 1s to 5s to match the longer deadlines.

## Test plan
- [x] `cargo test --package pluto-core deadline::tests::test_deadliner` passes
- [x] `cargo +nightly fmt --all --check` clean
- [x] `cargo clippy --package pluto-core --all-targets --all-features -- -D warnings` clean

### DIFF
--- a/crates/core/src/deadline.rs
+++ b/crates/core/src/deadline.rs
@@ -536,7 +536,10 @@ mod tests {
     async fn test_deadliner() {
         let (expired_duties, non_expired_duties, voluntary_exits) = setup_data();
 
-        // Use real time with short durations (milliseconds instead of hours/seconds)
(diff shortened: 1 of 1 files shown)


emlautarom1:
@varex83 Could be a reasonable workaround until we figure out #306, wdyt?

varex83 (approved):
Yes, I think we can use this approach as a workaround

PR #325, opened 2026-04-14
feat: add create cluster tests

varex83:
## Summary

Ports the Go `TestCreateCluster` test suite to Rust, covering all major `create_cluster` flows end-to-end, and fixes several correctness issues uncovered in the process.

### New tests (`crates/cli/src/commands/create_cluster.rs`)

A shared `run_test_create_cluster` helper drives all test cases, mirroring the Go test loop. Each case exercises `run()` against a temp directory and asserts:
- the command succeeds (or returns a specific error)
- the generated `cluster-lock.json` passes `verify_hashes()` and `verify_signatures()`
- all distributed validators have unique public keys
- deposit-data amounts match expectations
- CLI output and produced file tree match golden snapshots

Test cases covered:
| Test | Notes |
|---|---|
| `simnet` | 4-node simnet cluster |
| `goerli` | Goerli network, 3-node |
| `custom_testnet_flags` | Custom chain ID / fork version / genesis timestamp |
| `custom_target_gas_limit` | Non-default `target_gas_limit` |
| `solo_flow_definition_from_disk` | Definition loaded from a local JSON file |
| `solo_flow_definition_from_network` | Definition served over HTTP (wiremock) |
| `splitkeys` | `--split-existing-keys`, keys generated in a temp dir |
| `splitkeys_with_cluster_definition` | Split keys combined with a pre-existing definition |
| `four_partial_deposits` / `two_partial_deposits` | Multiple deposit amounts per validator |
| `with_fee_recipient_and_withdrawal_addresses` | Custom fee-recipient + withdrawal addresses |
| `validate_def` | Invalid definition triggers expected error |

Golden files live in `crates/cli/src/commands/testdata/` and can be regenerated with `UPDATE_GOLDEN=1 cargo test`.

### Bug fixes

- **`create_cluster::run`** — call `def.set_definition_hashes()` before proceeding with key generation; the hashes were previously unset, causing `verify_hashes()` to fail.
- **`definition.rs` deserialization** — validate that `num_validators == validator_addresses.len()` for schema versions v1.5–v1.10; add `#[serde(default)]` to `fee_recipient_address`, `withdrawal_address`, and `compounding` so older JSON examples without those fields still deserialize correctly.
- **`lock.rs` `no_registration` check** — remove the zero-`fee_recipient` guard. The Go reference uses `len == 0` (missing field), not the zero Ethereum address; a legitimately zero fee-recipient is valid. Only an all-zero BLS signature or pubkey indicates a missing registration.
- **`cluster-definition-005.json`** — fix malformed timestamp (stray spaces around `:` separators).

## Test plan

- [ ] `cargo test -p pluto-cli` passes with all new test cases green
- [ ] `UPDATE_GOLDEN=1 cargo test` regenerates golden files without diff
- [ ] Existing tests unaffected


crates/cli/src/commands/create_cluster.rs
```diff
@@ -1413,59 +1415,1387 @@ fn write_split_keys_warning(w: &mut dyn Write) -> std::io::Result<()> {
...
+
+    fn random_checksummed_eth_address() -> String {
+        let mut bytes = [0u8; 20];
+        rand::thread_rng().fill(&mut bytes[..]);
+        let hex_addr = format!("0x{}", hex::encode(bytes));
+        pluto_eth2util::helpers::checksum_address(&hex_addr).unwrap()
+    }
+
+    /// Port of Go's testCreateCluster helper.
+    async fn run_test_create_cluster(
```
mskrzypkows: I see that the order of tests is different than in Charon, e.g. in Charon "valid lock" is at the end. Is it intentional?
BTW, would be good to split the function into few.

mskrzypkows (approved):
One comment left, LGTM in general. Will you ask claude to review it?

crates/cli/src/commands/create_cluster.rs
varex83: Not intentional at all :D Will split it!

crates/cluster/src/examples/cluster-definition-005.json
```diff
@@ -32,7 +32,7 @@
   ],
   "uuid": "B6AE17B3-78F5-147B-2C37-2572B93437DF",
   "version": "v1.8.0",
-  "timestamp": "2023-05-18T15:12:46+02:00",
+  "timestamp": "2023-05-18T15: 12: 46+02: 00",
```
emlautarom1, line 35: Why did this change?

crates/cli/src/commands/create_cluster.rs
```diff
@@ -495,6 +495,8 @@ pub async fn run(w: &mut dyn Write, mut args: CreateClusterArgs) -> CliResult<()
     let (ops, node_keys) = get_operators(num_nodes, &args.cluster_dir)?;
 
     def.operators = ops;
+    def.set_definition_hashes()
+        .map_err(CreateClusterError::DefinitionError)?;
```
emlautarom1, line 500: No need for explicit `map_err`

crates/cli/src/commands/create_cluster.rs
```diff
@@ -1413,59 +1415,1413 @@ fn write_split_keys_warning(w: &mut dyn Write) -> std::io::Result<()> {
...
+        }
+        let expected = std::fs::read(&path).unwrap_or_else(|_| {
+            panic!(
+                "golden file not found: {}. Run with UPDATE_GOLDEN=1 to create it.",
+                path.display()
+            )
+        });
+        assert_eq!(
+            String::from_utf8_lossy(actual),
+            String::from_utf8_lossy(&expected),
```
emlautarom1: Do not parse as `utf8`, compare the bytes directly:


```suggestion
            actual,
            &expected,
```

crates/cli/src/commands/create_cluster.rs
```diff
@@ -1413,59 +1415,1413 @@ fn write_split_keys_warning(w: &mut dyn Write) -> std::io::Result<()> {
...
+    #[derive(Debug, Clone, Copy)]
+    enum DefProvider {
+        None,
+        Def006,
+        DefTwoNodes,
+    }
+
+    async fn test_eth1_client() -> EthClient {
+        EthClient::new("http://127.0.0.1:8545").await.unwrap()
+    }
```
emlautarom1, line 1525: If we need an actual Eth client running then we should use `testcontainers` to spin one up. If not, make this an empty string so we use the `Noop` client.

crates/cli/src/commands/create_cluster.rs
```diff
@@ -1413,59 +1415,1413 @@ fn write_split_keys_warning(w: &mut dyn Write) -> std::io::Result<()> {
...
+        withdrawal_addrs: Vec<String>,
+        split_keys_dir: Option<std::path::PathBuf>,
+        target_gas_limit: u64,
+    ) -> CreateClusterArgs {
+        let testnet_config = TestnetConfig {
+            chain_id: config.testnet_chain_id,
+            fork_version: config.testnet_fork_version.map(String::from),
+            genesis_timestamp: config.testnet_chain_id.map(|_| {
+                u64::try_from(chrono::Utc::now().timestamp()).expect("timestamp fits u64")
+            }),
```
emlautarom1: Is this intentional? If there is no `testnet_chain_id` then the timestamp is `None`, otherwise its the current timestamp (now)?

crates/cli/src/commands/create_cluster.rs
```diff
@@ -1413,59 +1415,1413 @@ fn write_split_keys_warning(w: &mut dyn Write) -> std::io::Result<()> {
...
+
+        if let Some(expected) = expected_err {
+            let err = result.unwrap_err();
+            let err_str = format!("{err}");
+            assert!(
+                err_str.contains(expected),
+                "expected error containing '{expected}', got: {err_str}"
+            );
+            return;
         }
```
emlautarom1, line 2443: It would be better to match on the `CliError` instead of using substrings

emlautarom1 (approved):
The golden test matching logic is no particularly good, specially since it does not show where the mismatch is located in case there is one. We can revisit this latter for improvements but I don't intend to block this.

emlautarom1:
@claude

crates/cluster/src/examples/cluster-definition-005.json
varex83: When calculating definition hashes charon doesn't preprocess the strings, the same logic we have in pluto as well. So this change basically roll backs to initial file, to make the def hashes check work.

crates/cli/src/commands/create_cluster.rs
varex83: We return `CliError` from the function, and here we convert it to `CreateClusterError` which later converts to `CliError`

PR #328, opened 2026-04-14
feat(dkg): FROST implementation

### LINKED ISSUE #254: Implement dkg/frost
### Description
Implement dkg/frost

### Acceptance criteria

emlautarom1:
Partially resolves #254.

---

Based on https://github.com/ZcashFoundation/frost, we build a library implementing FROST compatible with Coinbase's Kryptology (https://github.com/coinbase/kryptology/blob/v1.8.0/pkg/dkg/frost/README.md). No new dependencies are added to the workspace.

I've created some custom Go test cases in Charon that are not part of this PR that generate some fixtures with intermediate values which gives us some initial confidence that we can interop with Go nodes. Still, full compatibility requires further attention.

This PR requires nuanced review due to the challenging cryptography aspects of the code.


emlautarom1:
@claude Review

crates/frost/src/kryptology.rs
```diff
@@ -0,0 +1,615 @@
...
+/// Kryptology hash-to-scalar.
+///
+/// See: https://github.com/coinbase/kryptology/blob/1dcc062313d99f2e56ce6abc2003ef63c52dd4a5/pkg/core/curves/bls12381_curve.go#L50
+const KRYPTOLOGY_DST: &[u8] = b"BLS12381_XMD:SHA-256_SSWU_RO_";
+
+/// Hash to scalar using kryptology's ExpandMsgXmd construction.
+///
+/// `ExpandMsgXmd(SHA-256, msg, DST, 48)` -> reverse bytes -> pad to 64 ->
+/// `Scalar::from_bytes_wide`.
+fn kryptology_hash_to_scalar(msg: &[u8]) -> Scalar {
```
mskrzypkows, line 259: It can be done without manually reversing:
```
fn kryptology_hash_to_scalar(msg: &[u8]) -> Scalar {
    let xmd = expand_msg_xmd(msg, KRYPTOLOGY_DST, 48);
    Scalar::from_be_bytes_wide(&xmd)
}
```
and new Scalar function needed:
```
    /// Reduce big-endian bytes modulo the scalar field order.
    pub(crate) fn from_be_bytes_wide(bytes: &[u8]) -> Self {
        let mut scalar = blst_scalar::default();
        let mut fr = blst_fr::default();
        unsafe {
            blst_scalar_from_be_bytes(&mut scalar, bytes.as_ptr(), bytes.len());
            blst_fr_from_scalar(&mut fr, &scalar);
        }
        Scalar(fr)
    }
```

crates/frost/src/kryptology.rs
```diff
@@ -0,0 +1,615 @@
...
+/// - `rng`: Cryptographic RNG.
+pub fn round1<R: RngCore + CryptoRng>(
+    id: u32,
+    threshold: u16,
+    max_signers: u16,
+    ctx: u8,
+    rng: &mut R,
+) -> Result<(Round1Bcast, BTreeMap<u32, ShamirShare>, Round1Secret), DkgError> {
+    // Kryptology encodes participant identifiers into a single byte.
+    if max_signers > u16::from(u8::MAX) {
```
mskrzypkows: `Round1Secret::from_raw` misses the validation logic (lines 282 - 290). What do you think about extracting it to the separate function and using in both places?

crates/frost/src/kryptology.rs
```diff
@@ -0,0 +1,615 @@
...
+
+    let own_identifier = Identifier::from_u32(secret.id)?;
+    let own_share_scalar =
+        SigningShare::from_coefficients(&secret.coefficients, own_identifier).to_scalar();
+
+    let mut peer_commitments: BTreeMap<Identifier, VerifiableSecretSharingCommitment> =
+        BTreeMap::new();
+    let mut share_sum = Scalar::ZERO;
+
+    for (&sender_id, bcast) in received_bcasts {
```
mskrzypkows, line 415: We could check `secret.id == sender_id` and return an error to prevent some wrong usage of the round2 function.

crates/frost/src/kryptology.rs
```diff
@@ -0,0 +1,615 @@
...
+    /// Produce a BLS partial signature from a [`KeyPackage`] produced by
+    /// kryptology DKG.
+    ///
+    /// Computes `partial_sig = (key_package.signing_share) * H(msg)` where H
+    /// hashes the message to a G2 point using the Ethereum 2.0 DST.
+    ///
+    /// The `id` must be the original 1-indexed kryptology participant ID.
+    pub fn from_key_package(id: u32, key_package: &KeyPackage, msg: &[u8]) -> BlsPartialSignature {
+        let scalar = key_package.signing_share().to_scalar();
+        {
```
mskrzypkows: Why is it nested ?

crates/frost/src/frost_core.rs
```diff
@@ -0,0 +1,463 @@
+//! Port of frost-core types and functions, specialized for BLS12-381 G1 curve
+//! operations.
+//!
+//! Contains the key material types (identifiers, shares, packages) and the
+//! polynomial evaluation functions needed by the kryptology-compatible DKG.
+
+#![allow(clippy::arithmetic_side_effects)]
```
mskrzypkows: Would be better to limit it to some specific scopes.

crates/frost/src/kryptology.rs
```diff
@@ -0,0 +1,615 @@
...
+//!
+//! This module implements the same DKG protocol as
+//! `github.com/coinbase/kryptology/pkg/dkg/frost`, which differs from the
+//! standard FROST DKG in frost-core in the hash-to-scalar construction,
+//! challenge preimage format, proof representation, and round structure.
+//!
+//! The output types ([`KeyPackage`], [`PublicKeyPackage`]) are standard
+//! frost-core types usable with frost-core's signing protocol.
+
+#![allow(clippy::arithmetic_side_effects)]
```
mskrzypkows: Would be better to limit it to some specific scopes.

crates/frost/src/frost_core.rs
```diff
@@ -0,0 +1,463 @@
...
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+// See: https://github.com/ZcashFoundation/frost/blob/3ffc19d8f473d5bc4e07ed41bc884bdb42d6c29f/frost-core/src/identifier.rs#L121-L137
+impl Ord for Identifier {
+    /// Compare identifiers by their numeric scalar value, using big-endian byte
+    /// order. Serializes to little-endian, and compares in reverse order.
+    fn cmp(&self, other: &Self) -> Ordering {
```
mskrzypkows, line 86: This can be possibly optimized if we would keep u32 value at the construction time:
```
#[derive(Copy, Clone, Debug)]
pub struct Identifier {
    id: u32,
    scalar: Scalar,
}

impl Identifier {
    pub fn from_u32(id: u32) -> Result<Self, FrostCoreError> {
        let scalar = Scalar::from(u64::from(id));
        if scalar == Scalar::ZERO {
            Err(FrostCoreError::InvalidZeroScalar)
        } else {
            Ok(Self { id, scalar })
        }
    }

    pub fn to_scalar(&self) -> Scalar {
        self.scalar
    }
}

impl Ord for Identifier {
    fn cmp(&self, other: &Self) -> Ordering {
        self.id.cmp(&other.id)
    }
}
```

crates/frost/src/kryptology.rs
```diff
@@ -0,0 +1,615 @@
...
+
+use blst::*;
+use rand_core::{CryptoRng, RngCore};
+use sha2::{Digest, Sha256};
+
+use super::*;
+
+/// Errors from the kryptology-compatible DKG.
+#[derive(Debug)]
+pub enum DkgError {
```
mskrzypkows: Would be nice to cover all errors with tests.

crates/frost/src/frost_core.rs
```diff
@@ -0,0 +1,463 @@
...
+        let a = self.0.to_bytes();
+        let b = other.0.to_bytes();
+        for i in (0..32).rev() {
+            match a[i].cmp(&b[i]) {
+                Ordering::Equal => continue,
+                other => return other,
+            }
+        }
```
varex83agent: **major: No `Zeroize`/`Drop` on secret key types — signing material lingers in memory**

`SigningShare`, `KeyPackage`, `Round1Secret` (and `Round1Secret::coefficients`) all hold secret scalar material (`blst_fr = [u64; 4]`) but implement neither `Zeroize` nor a custom `Drop`. When these values are dropped the heap/stack bytes are never overwritten, leaving long-term signing shares and DKG polynomial coefficients recoverable from memory dumps or swap for the lifetime of the process.

The workspace already depends on `zeroize = "1.8.2"`. Add `zeroize.workspace = true` to `[dependencies]` in `Cargo.toml`, implement `Zeroize for Scalar` (zero the four `u64` limbs), then derive `ZeroizeOnDrop` for `SigningShare`, `KeyPackage`, and `Round1Secret` (clearing `coefficients: Vec<Scalar>` on drop).

crates/frost/src/kryptology.rs
```diff
@@ -0,0 +1,615 @@
...
+
+impl Round1Secret {
+    /// Reconstruct a [`Round1Secret`] from wire-format data (e.g. a test
+    /// fixture) so that the standard [`round2`] function can be called.
+    ///
+    /// `own_share` is the big-endian scalar the participant computed for
+    /// itself.  It is stored as the constant term of a zero polynomial so
+    /// that [`round2`]'s `from_coefficients` evaluation returns it
+    /// unchanged.
+    pub fn from_raw(
```
varex83agent: **major: `Round1Secret::from_raw` skips all input validation**

Unlike `round1`, `from_raw` never calls `validate_num_of_signers` or checks `id`. Consequences:
- `max_signers = 0`: `round2` computes `(secret.max_signers - 1) as usize`; in release builds the u16 subtraction wraps to `65535`, making the package-count check demand 65535 entries and silently return `Err(IncorrectPackageCount)`.
- `max_signers = 1` / `threshold = 1`: `expected = 0`, both length checks pass on empty maps, the peer-verification loop is skipped entirely, and the caller receives a `KeyPackage` with no proof-of-knowledge or Feldman VSS verification performed.
- `id = 0`: surfaces as opaque `DkgError::FrostCoreError(InvalidZeroScalar)` deep in `round2` instead of `DkgError::InvalidParticipantId(0)`.

Fix: add the same guards as `round1` at the top of `from_raw`:
```rust
if max_signers > u16::from(u8::MAX) { return Err(DkgError::InvalidSignerCount); }
validate_num_of_signers(threshold, max_signers)?;
if id == 0 || id > u32::from(max_signers) { return Err(DkgError::InvalidParticipantId(id)); }
```

crates/frost/src/kryptology.rs
```diff
@@ -0,0 +1,615 @@
...
+        id,
+        ctx,
+        coefficients,
+        commitment,
+        threshold,
+        max_signers,
+    };
+
+    Ok((bcast, shares, secret))
+}
```
varex83agent, line 377: **major: `round2` requires exactly `max_signers-1` packages; Go allows `threshold-1..max_signers`**

Go's `Round2` (`dkg_round2.go`) accepts `threshold-1 <= len(bcast) <= max_signers`, allowing fault-tolerant DKG where an offline or slow peer does not abort the round for everyone else — as long as at least `threshold-1` other peers responded, the participant can complete. The Rust implementation requires exactly `max_signers - 1`:
```rust
let expected = (secret.max_signers - 1) as usize;
if received_bcasts.len() != expected || received_shares.len() != expected {
    return Err(DkgError::IncorrectPackageCount);
}
```
This is a functional deviation: a single missing broadcast fails round2 for every honest participant even when enough shares exist to complete DKG. This should either be relaxed to `>= (secret.threshold - 1) as usize`, or explicitly documented as a deliberate stricter constraint.

crates/frost/src/kryptology.rs
```diff
@@ -0,0 +1,615 @@
...
+
+/// Convert big-endian 32 bytes to a `Scalar`.
+pub fn scalar_from_be(bytes: &[u8; 32]) -> Result<Scalar, DkgError> {
+    let mut le = *bytes;
+    le.reverse();
+    Scalar::from_bytes(&le).ok_or(DkgError::InvalidScalar)
+}
+
+/// RFC 9380 Section 5.3.1 using SHA-256
+pub fn expand_msg_xmd(msg: &[u8], dst: &[u8], len_in_bytes: usize) -> Vec<u8> {
```
varex83agent: **major: `expand_msg_xmd` is `pub` with `assert!()` — panics on invalid caller input**

Three `assert!()` calls will terminate the process if a downstream caller passes invalid parameters (`ell > 255`, `len_in_bytes > 65535`, `dst.len() > 255`). While current internal callers use hardcoded safe values, the function is part of the public API. A networked validator node integrating this crate could be crashed by triggering a code path that indirectly calls this with attacker-influenced lengths.

Options:
1. Change the signature to `fn expand_msg_xmd(...) -> Result<Vec<u8>, ExpandMsgError>` and replace `assert!` with `if ... { return Err(...) }`.
2. Make it `pub(crate)` since no external caller needs it directly (the Kryptology hash-to-scalar is what callers actually need).

Option 2 is simpler if no external party needs to call `expand_msg_xmd` directly.

crates/frost/src/frost_core.rs
```diff
@@ -0,0 +1,463 @@
...
+    collections::{BTreeMap, BTreeSet},
+};
+
+use super::*;
+
+/// Errors from key operations.
+#[derive(Debug)]
+pub enum FrostCoreError {
+    /// Participant ID is zero.
+    InvalidZeroScalar,
```
varex83agent, line 20: **major: `FrostCoreError` and `DkgError` do not implement `std::error::Error`**

Both error types derive `Debug` manually but do not implement `std::error::Error`. This means they:
- Cannot be used with the `?` operator in downstream crates
- Cannot be wrapped via `#[from]`
- Cannot be boxed as `Box<dyn std::error::Error>`

The project uses `thiserror` consistently for library error types. Add `thiserror.workspace = true` to `[dependencies]` and:
```rust
#[derive(Debug, thiserror::Error)]
pub enum FrostCoreError {
    #[error("participant ID is zero")]
    InvalidZeroScalar,
    // ...
}
```
Also, `FrostCoreError` is an implementation detail — it should be `pub(crate)` and only `DkgError` should be part of the public API surface.

crates/frost/src/lib.rs
```diff
@@ -0,0 +1,18 @@
...
+//! This crate implements a distributed key generation protocol compatible with
+//! Go's Coinbase Kryptology FROST DKG, and BLS threshold signing (Ethereum 2.0
+//! compatible).
+
+#![allow(non_snake_case)]
+#![doc = include_str!("../dkg.md")]
+
+pub mod curve;
+pub mod frost_core;
+pub mod kryptology;
```
varex83agent, line 10: **major: Wildcard `pub use curve::*; pub use frost_core::*` creates an uncontrolled public API surface**

Glob re-exports make every `pub` item from both modules part of the crate's public API, including `FrostCoreError`, `SecretShare`, `CoefficientCommitment`, `VerifiableSecretSharingCommitment`, and other internal types. The existing codebase pattern (see `pluto-peerinfo`, `pluto-tracing`) is to enumerate specific re-exports. Enumerate only the types downstream callers actually need (`DkgError`, `KeyPackage`, `PublicKeyPackage`, `kryptology::*`) and mark internal types `pub(crate)` or `pub(super)`. This also means any internal change becomes a semver-breaking change today.

crates/frost/src/curve.rs
```diff
@@ -0,0 +1,279 @@
...
+use std::{
+    fmt,
+    ops::{Add, Mul, Sub},
+};
+
+use blst::*;
+use rand_core::{CryptoRng, RngCore};
+
+/// BLS12-381 scalar field element. Wrapper around `blst_fr` in Montgomery form.
+#[derive(Copy, Clone, Default, PartialEq, Eq)]
```
varex83agent, line 19: **minor: `Scalar::PartialEq` is not constant-time — timing side-channel in proof verification**

`Scalar` derives `PartialEq`, which compares `blst_fr { l: [u64; 4] }` limbs using Rust's default `==` (variable-time). The most sensitive use is in `round2` (`if ci_check != ci`) where this is used for Schnorr challenge verification. While neither value is directly secret, a timing oracle here could in theory help an adversary distinguish near-valid from far-invalid proofs. The `invert` zero-check path is lower risk since it operates on public participant IDs.

Fix: use `subtle::ConstantTimeEq` for scalar comparisons in the proof verification path, or at minimum add a `// NOTE: not constant-time` comment to the `PartialEq` derive.

crates/frost/src/kryptology.rs
```diff
@@ -0,0 +1,615 @@
...
+    let mut share_sum = Scalar::ZERO;
+
+    for (&sender_id, bcast) in received_bcasts {
+        let sender_commitment =
+            deserialize_commitment(sender_id, secret.threshold, &bcast.commitments)?;
+        let a0 = sender_commitment.coefficients()[0].value();
+
+        // Verify proof of knowledge
+        let wi = scalar_from_be(&bcast.wi)?;
+        let ci = scalar_from_be(&bcast.ci)?;
```
varex83agent, line 426: **minor: Missing zero-check on `ci` before proof verification (parity gap with Go)**

Go's `dkg_round2.go` explicitly rejects a broadcast where `Ci.IsZero()` before any arithmetic:
```go
if bcast[id].Ci.IsZero() {
    return nil, fmt.Errorf("ci should not be zero from participant %d", id)
}
```
The Rust `round2` omits this guard. A `ci = 0` broadcast forces a full G1 scalar-mult (`wi*G`) before the challenge comparison fails, instead of a cheap field check. Add:
```rust
if ci == Scalar::ZERO {
    return Err(DkgError::InvalidProof { culprit: sender_id });
}
```
immediately after deserializing `ci`.

crates/frost/src/kryptology.rs
```diff
@@ -0,0 +1,615 @@
...
+            u8::try_from(sender_id).map_err(|_| DkgError::InvalidParticipantId(sender_id))?;
+        let ci_check = kryptology_challenge(sender_id_u8, secret.ctx, &a0, &r_reconstructed);
+        if ci_check != ci {
+            return Err(DkgError::InvalidProof { culprit: sender_id });
+        }
+
+        // Verify Feldman share
+        let share = received_shares
+            .get(&sender_id)
+            .ok_or(DkgError::IncorrectPackageCount)?;
```
varex83agent: **minor: Misleading `IncorrectPackageCount` error when a share is missing for a specific sender**

```rust
let share = received_shares
    .get(&sender_id)
    .ok_or(DkgError::IncorrectPackageCount)?;
```
`IncorrectPackageCount` is already used above for the wrong map length. When a sender is present in `received_bcasts` but absent from `received_shares`, the correct error is one that identifies the culprit. Use `DkgError::InvalidShare { culprit: sender_id }` here (or add a dedicated `MissingShare { culprit }` variant) so the error message attributes the failure to the correct participant.

crates/frost/src/lib.rs
```diff
@@ -0,0 +1,18 @@
+//! Kryptology-compatible FROST DKG and BLS threshold signing over BLS12-381 G1.
+//! This crate implements a distributed key generation protocol compatible with
+//! Go's Coinbase Kryptology FROST DKG, and BLS threshold signing (Ethereum 2.0
+//! compatible).
+
+#![allow(non_snake_case)]
```
varex83agent: **minor: `#![allow(non_snake_case)]` at crate root is too broad**

`wi` and `ci` are already valid `snake_case` identifiers (all lowercase). If this allow was added for mathematical variable names from the FROST paper (e.g., `Ri`, `Wi`) those usages should be individually annotated:
```rust
#[allow(non_snake_case)]
fn kryptology_challenge(...) { let Ri = ...; }
```
A crate-level suppression will hide real naming violations in any code added to the crate in the future.

crates/frost/src/kryptology.rs
```diff
@@ -0,0 +1,615 @@
...
+
+impl BlsPartialSignature {
+    /// Produce a BLS partial signature from a [`KeyPackage`] produced by
+    /// kryptology DKG.
+    ///
+    /// Computes `partial_sig = (key_package.signing_share) * H(msg)` where H
+    /// hashes the message to a G2 point using the Ethereum 2.0 DST.
+    ///
+    /// The `id` must be the original 1-indexed kryptology participant ID.
+    pub fn from_key_package(id: u32, key_package: &KeyPackage, msg: &[u8]) -> BlsPartialSignature {
```
varex83agent: nit: Unnecessary inner scope and reference alias in `from_key_package` — looks like an unfinished zeroization attempt:
```rust
let scalar = key_package.signing_share().to_scalar();
{
    let signing_share: &Scalar = &scalar;  // redundant reborrow
    let h_msg = hash_to_g2(msg);
    BlsPartialSignature { identifier: id, point: p2_mult(&h_msg, signing_share) }
}
```
The inner block drops `signing_share` but `scalar` remains live and unzeroized in the outer scope. Simplify:
```rust
let scalar = key_package.signing_share().to_scalar();
let h_msg = hash_to_g2(msg);
BlsPartialSignature { identifier: id, point: p2_mult(&h_msg, &scalar) }
```
(A real zeroization should use `Zeroizing<Scalar>` once `Scalar: Zeroize` is implemented.)

crates/frost/src/lib.rs
```diff
@@ -0,0 +1,18 @@
...
+#![doc = include_str!("../dkg.md")]
+
+pub mod curve;
+pub mod frost_core;
+pub mod kryptology;
+
+pub use curve::*;
+pub use frost_core::*;
+pub use rand_core;
```
varex83agent: nit: `pub use rand_core` re-exports the entire `rand_core` crate as stable public API, creating hard version coupling — any semver bump to `rand_core` in the workspace becomes a breaking change for downstream users of this crate. If the intent is to let callers use `RngCore + CryptoRng` without adding `rand_core` as a direct dep, re-export only the specific traits: `pub use rand_core::{CryptoRng, RngCore};`. Otherwise remove the `pub`.

crates/frost/src/kryptology.rs
```diff
@@ -0,0 +1,615 @@
...
+    }
+
+    /// Verify a BLS signature against a public key.
+    ///
+    /// Uses the Ethereum 2.0 BLS verification (pairing check) with the
+    /// standard DST.
+    pub fn verify(&self, verifying_key: &VerifyingKey, msg: &[u8]) -> bool {
+        let pk_affine = G1Affine::from(verifying_key.to_element());
+        let pk = blst::min_pk::PublicKey::from(pk_affine.0);
+
```
varex83agent, line 629: nit: `blst_p1_mult` and `blst_p2_mult` are called with `nbits=255` — this is correct (the BLS12-381 scalar field order `r` has exactly 255 significant bits), but future readers will wonder why 255 and not 256. Add a brief comment:
```rust
// BLS12-381 scalar field order has 255 significant bits; blst skips the MSB.
blst_p1_mult(&mut out, &self.0, scalar.b.as_ptr(), 255);
```

varex83agent (dismissed):
This PR adds the `pluto-frost` crate implementing a FROST-compatible DKG and BLS threshold signing over BLS12-381, targeting interoperability with Go's Coinbase Kryptology FROST DKG. The cryptographic math has been verified correct: the Horner polynomial evaluation, challenge preimage format, `expand_msg_xmd` construction (matching RFC 9380 and confirmed against test vectors), `blst_p1_mult` bit-width (255 is correct for the BLS12-381 scalar field), and the `Scalar::ONE` Montgomery constant all check out.

**Major issues requiring changes before merge:**
1. Secret key types (`SigningShare`, `KeyPackage`, `Round1Secret`) have no `Zeroize`/`Drop` implementation — signing key material lingers in process memory.
2. `round2` enforces exactly `max_signers-1` received packages; Go's implementation accepts any count in `[threshold-1, max_signers]` for fault-tolerance, meaning a single offline peer aborts every honest participant in the Rust version.
3. `expand_msg_xmd` is `pub` with three `assert!()` calls that will panic the process if a downstream caller passes e.g. a DST > 255 bytes.
4. `Round1Secret::from_raw` performs no input validation (`id`, `threshold`, `max_signers`), enabling underflow in `round2` and bypassed peer verification when `max_signers=1`.
5. `FrostCoreError` and `DkgError` do not implement `std::error::Error`, breaking usage with `?` in downstream crates.

Minor/nit findings cover test coverage gaps (no ctx≠0 test, no tampered-share test), API surface issues (glob re-exports, public fields on wire types), and a non-constant-time scalar equality used in proof verification.

crates/frost/src/kryptology.rs
varex83: I think this comment is legit

crates/frost/src/kryptology.rs
emlautarom1: Intended to be used for tests only, probably better to gate it behind a `cfg`.

crates/frost/src/kryptology.rs
emlautarom1: No external usage so `pub(crate)` sounds more reasonable.

crates/frost/src/curve.rs
```diff
@@ -0,0 +1,279 @@
...
+pub struct Scalar(pub(crate) blst_fr);
```
iamquang95, line 20: Should `impl Zeroize for Scalar`

crates/frost/src/kryptology.rs
```diff
@@ -0,0 +1,615 @@
...
+    /// The share value as big-endian scalar bytes.
+    pub value: [u8; 32],
+}
+
+/// Secret state held by a participant between round 1 and round 2.
+///
+/// # Security
+///
+/// This MUST NOT be sent to other participants.
+pub struct Round1Secret {
```
iamquang95, line 109: should derive `ZeroizeOnDrop`

crates/frost/src/frost_core.rs
```diff
@@ -0,0 +1,463 @@
...
+
+        Some(VerifiableSecretSharingCommitment::new(cc))
+    }
+}
+
+/// A secret scalar value representing a signer's share of the group secret.
+///
+/// See: https://github.com/ZcashFoundation/frost/blob/3ffc19d8f473d5bc4e07ed41bc884bdb42d6c29f/frost-core/src/keys.rs#L82-L87
+#[derive(Copy, Clone, Debug)]
+pub struct SigningShare(Scalar);
```
iamquang95, line 142: should derive `ZeroizeOnDrop`

crates/frost/src/frost_core.rs
```diff
@@ -0,0 +1,463 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+/// A key package containing all key material for a participant.
+///
+/// See: https://github.com/ZcashFoundation/frost/blob/3ffc19d8f473d5bc4e07ed41bc884bdb42d6c29f/frost-core/src/keys.rs#L617-L643
+#[derive(Debug)]
```
iamquang95: Should derive `ZeroizeOnDrop`

crates/frost/src/kryptology.rs
```diff
@@ -0,0 +1,615 @@
...
+    /// The group verification key (compressed G1, 48 bytes).
+    pub verification_key: [u8; 48],
+    /// This participant's verification share (compressed G1, 48 bytes).
+    pub vk_share: [u8; 48],
+}
+
+/// A Shamir secret share matching Go's `sharing.ShamirShare`.
+///
+/// The `value` field is in **big-endian** byte order.
+#[derive(Clone, Debug, PartialEq, Eq)]
```
iamquang95: Should derive `ZeroizeOnDrop`

crates/frost/src/kryptology.rs
```diff
@@ -0,0 +1,615 @@
...
+///   they sent us.
+pub fn round2(
+    secret: Round1Secret,
+    received_bcasts: &BTreeMap<u32, Round1Bcast>,
+    received_shares: &BTreeMap<u32, ShamirShare>,
+) -> Result<(Round2Bcast, KeyPackage, PublicKeyPackage), DkgError> {
+    let expected = (secret.max_signers - 1) as usize;
+    if received_bcasts.len() != expected || received_shares.len() != expected {
+        return Err(DkgError::IncorrectPackageCount);
+    }
```
iamquang95: This seem stricter than golang implementation of round2 check, they are checking with threshold

https://github.com/coinbase/kryptology/blob/v1.8.0/pkg/dkg/frost/dkg_round2.go#L39-L49

crates/frost/src/kryptology.rs
iamquang95: agree on this

crates/frost/src/kryptology.rs
iamquang95: It's better to have the `#[cfg(test)]`

iamquang95: Also add comment on the behavior

varex83 (approved):
Since it already works, let's merge this and make all fixes in #364

iamquang95:
@varex83 The #364: I addressed the comments, I think it can be merged into this PR. So we have a "full" pr related to FROST implementation, also check the addressed comments which related to this PR.

iamquang95:
@claude

iamquang95:
Seem good, let's merge

iamquang95:
Currently ignore RUSTSEC-2026-0118 and RUSTSEC-2026-0119 since there are some libp2p crates are depedning on that. Not a blocker. Issue is tracked here: https://github.com/NethermindEth/pluto/issues/365

cc @varex83 good to merge

PR #300, opened 2026-03-28
feat: add mev cli tests

PoulavBhowmick03:
Closes #235

PR #329, opened 2026-04-15
test(cli/mev): add integration tests for MEV test command

varex83agent:
## Summary

Follow-up to #300. Adds the integration tests missing from the initial MEV test implementation, as noted in the review thread.

- **default scenario**: healthy mocked MEV node → `Ping OK`, `PingMeasure Good`, `CreateBlock Skip`
- **connection refused**: two unreachable endpoints → `Ping Fail`, `PingMeasure Fail`, `CreateBlock Skip` with error messages
- **timeout**: 100ns timeout → all network tests fail with timeout/interrupted verdict
- **quiet mode**: `--quiet` produces no stdout output
- **unsupported test case**: unknown test name returns `"test case not supported"` error
- **custom test cases**: `--test-cases=Ping` runs only `Ping` per endpoint
- **write to file**: `--output-json` produces valid JSON with `mev` key

Tests ported from [`testmev_internal_test.go`](https://github.com/ObolNetwork/charon/blob/v1.7.1/cmd/testmev_internal_test.go). The `default load scenario` and flag-parsing tests are deferred to a further follow-up since they require a live beacon node or deeper CLI wiring.

## Test plan

- [ ] `cargo test --package pluto-cli --bin pluto -- commands::test::mev::tests` passes (9/9)
- [ ] `cargo clippy` clean
- [ ] `cargo +nightly fmt` clean
- [ ] `cargo deny check` clean


crates/cli/src/commands/test/mev.rs
mskrzypkows: This can silently fail, would be better to return an error.

mskrzypkows: ditto

mskrzypkows: those two as well, why not return an error?

mskrzypkows: ditto

mskrzypkows: ditto

mskrzypkows: ditto

mskrzypkows (commented):
LGTM, I would just change some unwraps to returning errors in case this code would be used in different context.

mskrzypkows (approved):

varex83:
@claude merge main and fix comments

varex83:
@claude review

varex83:
@claude re-review please, check if all issues were addressed

PR #330, opened 2026-04-15
fix: upgrade dependencies and fix RUSTSEC-2026-0098

iamquang95:
The `core2` is yanked: https://crates.io/crates/core2 
https://github.com/bbqsrc/core2 `This repository was archived by the owner on Apr 15, 2026. It is now read-only.`

We have to wait libp2p to update their dependencies


emlautarom1:
Could you check if running `cargo update` alone (without the updates to `deny.toml`) is enough to sort out the issue? It seems like the situation was sorted out already https://github.com/libp2p/rust-libp2p/issues/6383.

PR #331, opened 2026-04-15
ci: use local /review-pr skill for PR reviews

varex83agent:
## Summary

- Replaces the external `code-review` plugin (`plugin_marketplaces` + `plugins`) with the repo's own `/review-pr` skill
- Upgrades `pull-requests` permission from `read` to `write` so the skill can post inline review comments via the GitHub API


varex83 (approved):
Self-approving it, since before we use `claude.yml` to review, and not this one

PR #336, opened 2026-04-16
fix: ignore stale OutboundFailure events in bcast behaviour

pinebit:
## Summary
- Fix a bug where a stale `OutboundFailure` from a previously failed broadcast could kill a subsequent active broadcast
- When a broadcast fails (e.g. one peer times out), in-flight requests to other peers are still running. Their eventual failures would unconditionally call `complete_active_broadcast`, taking down whatever broadcast was active at that point
- The fix checks whether the `op_id` belongs to the current broadcast before acting. If it doesn't match, the event is logged and ignored — consistent with how `SigResponse` and `MessageSent` already handle stale events

## Test plan
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings` passes
- [x] `cargo +nightly fmt --all --check` passes
- [x] `cargo test --workspace --all-features` passes

### DIFF
--- a/crates/dkg/src/bcast/behaviour.rs
+++ b/crates/dkg/src/bcast/behaviour.rs
@@ -346,13 +346,15 @@ impl Behaviour {
                 }
             }
             OutEvent::OutboundFailure { op_id, failure } => {
-                if let Some(state) = self.active_broadcast.as_mut() {
-                    let failed_peer = state
+                let failed_peer = self.active_broadcast.as_mut().and_then(|state| {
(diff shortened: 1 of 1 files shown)


iamquang95 (approved):

emlautarom1:
@iamquang95 Any chance we could add a test for this edge case, or is it very complicated to reproduce?

iamquang95:
@emlautarom1 let me try to add an unit test for this scenario

PR #346, opened 2026-04-21
feat:  test infra

### LINKED ISSUE #234: Implement cli test infra
### Description
Port the module `cmd/testinfra` from the reference Go implementation.

### Acceptance criteria

mskrzypkows:
fixes #234


mskrzypkows:
@claude review

mskrzypkows:
@claude review

mskrzypkows:
Regarding issue 4, I think that it's better to check directly if `fio` is available, without using `which` as a intermediary

crates/cli/src/commands/test/infra.rs
```diff
@@ -1,9 +1,624 @@
...
+        .stdout(std::process::Stdio::piped())
+        .stderr(std::process::Stdio::piped())
+        .kill_on_drop(true)
+        .spawn()
+        .map_err(|e| CliError::Other(format!("exec fio command: {e}")))?
+        .wait_with_output()
+        .await
+        .map_err(|e| CliError::Other(format!("exec fio command: {e}")))?;
+
+    let _ = tokio::fs::remove_file(&filename).await;
```
iamquang95: If the process is canceled before this line, the file will stay in the system.

iamquang95: Maybe `tempfile` can help in this case.

iamquang95 (commented):
One of my concerns about this command is that rust and charon use different speedtest checks, which will yield different results when run on the same machine.

crates/cli/src/commands/test/infra.rs
mskrzypkows: done

mskrzypkows:
@iamquang95 I improved speedtest moudle, it better reflects now the GO version

crates/cli/src/commands/test/speedtest.rs
```diff
@@ -0,0 +1,334 @@
...
+            .expect("deadline does not overflow");
+
+        let mut set = tokio::task::JoinSet::new();
+        for _ in 0..speed_test_concurrency() {
+            let client = client.clone();
+            let url = upload_url.clone();
+            set.spawn(async move {
+                let mut bytes: usize = 0;
+                while Instant::now() < deadline {
+                    let chunk = vec![0u8; UPLOAD_CHUNK_BYTES];
```
iamquang95: nit: this is quite expensive since it's repeated, wrap it in `bytes::Bytes` outside the loop so we can do a cheap clone

Cargo.toml
```diff
@@ -97,6 +97,8 @@ tree_hash_derive = "0.12"
 tar = "0.4"
 flate2 = "1.1"
 wiremock = "0.6"
+sysinfo = "0.33"
+quick-xml = { version = "0.37", features = ["serialize"] }
```
iamquang95: use 0.39, the quick-xml is existed in cargo.lock with 0.39 version

iamquang95 (approved):
other part is lg-tm

PR #347, opened 2026-04-21
fix: Incorrect P2PContext in `bcast` example

emlautarom1:
The `bcast` example (https://github.com/NethermindEth/pluto/blob/c4c922a19c0efc29baf882d3ed438e4ccdc73f82/crates/dkg/examples/bcast.rs) is currently broken due to the usage of an incorrect `P2PContext` which results in active connections to known peers not getting updated, thus failing to broadcast despite the connections actually being active.

This PR implements the minimal fix required to sort out the issue, but some comments are included to avoid similar situations in the future.

### DIFF
--- a/crates/dkg/examples/bcast.rs
+++ b/crates/dkg/examples/bcast.rs
@@ -82,7 +82,6 @@ use pluto_p2p::{
     config::P2PConfig,
     gater, k1,
     p2p::{Node, NodeType},
-    p2p_context::P2PContext,
     relay::{MutableRelayReservation, RelayRouter},
 };
 use pluto_tracing::TracingConfig;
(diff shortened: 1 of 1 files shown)


varex83 (approved):

crates/dkg/examples/bcast.rs
```diff
@@ -578,23 +577,21 @@ async fn main() -> Result<()> {
...
     };
 
-    let p2p_context = P2PContext::new(known_peers.clone());
-    let (bcast_behaviour, component) =
-        bcast::Behaviour::new(cluster_peers.clone(), p2p_context.clone(), key.clone());
-    register_message(&component, local_node_number)
-        .await
-        .expect("Failed to register demo bcast message");
-
+    let mut component = None;
```
emlautarom1, line 580: This `handle` approach is similar to what we have in `parsigex`: https://github.com/NethermindEth/pluto/pull/291#discussion_r3096062351.

The design is very cumbersome to use so it requires some further work.

crates/dkg/examples/bcast.rs
```diff
@@ -578,23 +577,21 @@ async fn main() -> Result<()> {
...
     let mut node: Node<ExampleBehaviour> = Node::new(
         p2p_config,
-        key,
+        key.clone(),
         NodeType::QUIC,
         args.filter_private_addrs,
         known_peers,
         |builder, keypair, relay_client| {
             let p2p_context = builder.p2p_context();
```
iamquang95, line 588: can you try to remove this and use the outer `p2p_context`, that will be cleaner without having the mut `component`

crates/dkg/examples/bcast.rs
emlautarom1, line 621: This set duplicates the information that `P2PContext` and its `PeerStore` already manage. `ClusterInfo` is in the same situation, where pretty much all information is duplicated in some way.

crates/dkg/examples/bcast.rs
emlautarom1: We can, I was just following the patterns that we already have in `bootnode` and `parsigex`:

https://github.com/NethermindEth/pluto/blob/0de6cdc1c4352c32ba43e6167301deedb7d1626e/crates/p2p/examples/bootnode.rs#L141-L163

https://github.com/NethermindEth/pluto/blob/0de6cdc1c4352c32ba43e6167301deedb7d1626e/crates/parsigex/examples/parsigex.rs#L264-L296

iamquang95: Maybe we will revisit these examples to see if we can have a cleaner approach. LGTM now

iamquang95 (approved):

PR #348, opened 2026-04-22
refactor(p2p): not allow builder to set p2p_context

iamquang95:
This fixes a `P2PContext` ownership misuseAPI in the P2P builder flow.

Previously, `Node::new` created a `P2PContext` internally while callers could also create one externally and `PlutoBehaviourBuilder` could overwrite it again. That made it possible for `Node`, `PlutoBehaviour`, and inner behaviours to accidentally use different contexts.

This change makes `P2PContext` an explicit constructor input to `Node::new` / `Node::new_server`, removes `with_p2p_context` from the builder, and moves local peer binding into `Node`.

Benefit: one node now has one authoritative shared `P2PContext`, making the API safer and preventing split state between node-level and behaviour-level components.

### DIFF
--- a/crates/relay-server/src/p2p.rs
+++ b/crates/relay-server/src/p2p.rs
@@ -14,7 +14,10 @@ use crate::{
     config::{Config, create_relay_config},
     web::enr_server,
 };
-use pluto_p2p::p2p::{Node, NodeType};
+use pluto_p2p::{
+    p2p::{Node, NodeType},
+    p2p_context::P2PContext,
+};
 
 /// Runs a relay P2P node.
 #[instrument(skip(config, key, ct))]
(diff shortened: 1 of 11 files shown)


emlautarom1:
Looks like a reasonable solution, :+1:

varex83 (approved):
LGTM!

crates/relay-server/src/p2p.rs
```diff
@@ -25,13 +28,13 @@ pub async fn run_relay_p2p_node(
 ) -> Result<Node<relay::Behaviour>> {
     let relay_config = create_relay_config(config);
     // Relay servers don't track cluster peers - they serve all connections
-    let known_peers: Vec<libp2p::PeerId> = vec![];
+    let p2p_context = P2PContext::default();
```
varex83: I think you can inline this into `new_server` function

PR #349, opened 2026-04-23
fix: clippy on qbft and pin rust-toolchain

iamquang95:
Local and CI lint results could diverge because rust-toolchain.toml was not actually pinning a Rust version. It used channel = "stable" plus an unsupported version field, so different machines could run different rustc/clippy versions and surface different warnings.

This PR pins the workspace toolchain to 1.95.0 in rust-toolchain.toml and remove the invalid version key.

### DIFF
--- a/crates/core/src/qbft/mod.rs
+++ b/crates/core/src/qbft/mod.rs
@@ -1060,7 +1060,7 @@ where
         }
 
         if (qrc.len() as i64) >= d.quorum() && has_highest_prepared {
-            qrc.extend(prepares.into_iter());
+            qrc.extend(prepares);
             return Some(qrc);
         }
     }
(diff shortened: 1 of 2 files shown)


mskrzypkows (approved):

emlautarom1:
We'll need an extra PR to update the Dockerfile to use the appropriate toolchain version:

https://github.com/NethermindEth/pluto/blob/27f2b33a00c9d9e5a9ccc500bfa81ba8a7757468/Dockerfile#L17

emlautarom1:
Also in docs:

https://github.com/NethermindEth/pluto/blob/27f2b33a00c9d9e5a9ccc500bfa81ba8a7757468/README.md#L8

PR #350, opened 2026-04-23
feat: add dkg/exchanger 

### LINKED ISSUE #253: Implement dkg/exchanger
### Description
Implement dkg/exchanger

### Acceptance criteria

varex83:


crates/dkg/src/exchanger.rs
```diff
@@ -0,0 +1,568 @@
...
+        peers: Vec<PeerId>,
+        sig_types: Vec<SigType>,
+    ) -> Self {
+        // Partial signature roots not known yet, so skip verification in parsigex,
+        // rather verify before we aggregate.
+        let st: HashSet<SigType> = sig_types.iter().copied().collect();
+
+        let duty_gater_fn: DutyGaterFn = {
+            let sig_types = sig_types.clone();
+            let st = st.clone();
```
emlautarom1: nit: This matches Charon but I don't understand why we need the same data as `Vec` and `HashSet` (we do a `contains` in both cases).

crates/dkg/src/exchanger.rs
```diff
@@ -0,0 +1,568 @@
...
+                    }
+                    Ok(())
+                }
+            });
+            sigdb
+                .lock()
+                .await
+                .subscribe_internal(sub)
+                .await
+                .expect("subscribe_internal is infallible");
```
emlautarom1: Why did we make `subscribe_internal` return a Result if it's infallible? Can we update its definition?

crates/dkg/src/exchanger.rs
```diff
@@ -0,0 +1,568 @@
...
+        }
+        drop(result_tx);
+
+        // Collect results into actual: one entry per sig_type (last writer wins,
+        // all nodes return equivalent data for each sig_type)
+        let mut actual: SigTypeStore = HashMap::new();
+        while let Some((sig_type, data)) = result_rx.recv().await {
+            actual.insert(sig_type, data);
+        }
+        while join_set.join_next().await.is_some() {}
```
emlautarom1: This can be simplified by using the `JoinSet` values instead of using extra channels


```suggestion
        // Run concurrent exchanges: for each (node, sig_type) pair, spawn a task
        let mut join_set = tokio::task::JoinSet::new();
        for (node_idx, ex) in exchangers.iter().enumerate() {
            for &sig_type in &sig_types {
                let ex = Arc::clone(ex);
                let set = data_to_be_sent[node_idx].clone();
                join_set.spawn(async move {
                    let data = ex.exchange(sig_type, set).await.expect("exchange failed");
                    (sig_type, data)
                });
            }
        }

        // Collect results into actual: one entry per sig_type (last writer wins,
        // all nodes return equivalent data for each sig_type)
        let actual: SigTypeStore = join_set.join_all().await.into_iter().collect();
```

crates/dkg/src/exchanger.rs
```diff
@@ -0,0 +1,568 @@
...
+                    got, want,
+                    "sig_type {sig_type} pk {pk}: wrong share indices"
+                );
+            }
+        }
+        assert_eq!(actual.len(), sig_types.len());
+
+        for stop_tx in stop_txs {
+            let _ = stop_tx.send(());
+        }
```
emlautarom1: Instead of having multiple `stop_txs` we could have a single `CancellationToken` shared across all swarms.

We could also ignore cancellation entirely since the spawned tasks will get dropped when the function completes. The current design is not waiting for them to finish (no `join` on the handles), so it would not make a difference.

emlautarom1 (approved):
LGTM, some small comments on some improvements to the tests.

crates/dkg/src/exchanger.rs
```diff
@@ -0,0 +1,537 @@
...
+    /// in `set`. Returns a map of public key → all partial signatures (one
+    /// per participant).
+    pub async fn exchange(
+        &self,
+        sig_type: SigType,
+        set: ParSignedDataSet,
+    ) -> Result<HashMap<PubKey, Vec<ParSignedData>>, ExchangerError> {
+        if !self.sig_types.contains(&sig_type) {
+            return Err(ExchangerError::UnrecognisedSigType(sig_type));
+        }
```
iamquang95: This check seems not correct. 
As `dkg.go` they will call `SIG_DEPOSIT_DATA + i` (https://github.com/ObolNetwork/charon/blob/v1.7.1/dkg/dkg.go#L701)

```
peerSigs, err := ex.exchange(ctx, sigType(int(sigDepositData)+i), parSig)
```

But the `newExchanger`, they only register the `sigDepositData`
https://github.com/ObolNetwork/charon/blob/v1.7.1/dkg/dkg.go#L242-L246

```
ex := newExchanger(p2pNode, nodeIdx.PeerIdx, peerIDs, []sigType{
		sigLock,
		sigDepositData,
		sigValidatorRegistration,
	}, conf.Timeout)
```

iamquang95: Also golang doesn't have that check

crates/dkg/src/exchanger.rs
```diff
@@ -0,0 +1,537 @@
...
+
+        // Wiring core workflow components
+
+        {
+            let handle_clone = handle.clone();
+            let sub = internal_subscriber(move |duty, set| {
+                let handle = handle_clone.clone();
+                async move {
+                    if let Err(e) = handle.broadcast(duty, set).await {
+                        warn!(error = %e, "Failed to broadcast partial signatures during DKG");
```
iamquang95, line 201: One thing I worry about with the rust libp2p design is how we can manage to handle "enough" fail events at the swarm level.

iamquang95 (changes requested):

iamquang95 (approved):
LGTM

PR #352, opened 2026-04-23
chore: Update Rust toolchain

emlautarom1:
Update the Rust toolchain version listed in other sources after #349

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ RUN apt-get update && \
   protobuf-compiler=3.21.12*
 
 # Install Rust using rustup, the official installer
-RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain 1.89.0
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -14,7 +14,7 @@ RUN apt-get update && \
   protobuf-compiler=3.21.12*
 
 # Install Rust using rustup, the official installer
-RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain 1.89.0
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain 1.95.0
```
iamquang95: Can we copy the rust-toolchain here and let the rustup install the version on that file?

emlautarom1: ~I don't know how to make the rustup installation do that. I could do it by hacking some bash incantation (sed/grep) though.~

Found a way, updated.

varex83 (approved):

iamquang95 (approved):

Issue #129, opened 2026-01-21
Finish relay server


varex83:
Current relay server implementation is incomplete and missing metrics and other validation steps that exist in Go implementation

PR #354, opened 2026-04-24
feat: relay server missing features

mskrzypkows:
#129


mskrzypkows:
@claude review

mskrzypkows:
@claude review again

crates/p2p/src/bandwidth.rs
```diff
@@ -0,0 +1,398 @@
...
+use vise::{Counter, EncodeLabelSet, Family, Global, Metrics};
+
+#[derive(EncodeLabelSet, Hash, Clone, Eq, PartialEq, Debug)]
+struct PeerLabel {
+    peer: String,
+    peer_cluster: String,
+}
+
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "relay_p2p")]
```
varex83: should it have "relay_p2p" metrics prefix? I think we can generalize it

varex83: Also, I would prefer to keep all metrics in a separate file

varex83 (approved):

crates/p2p/src/bandwidth.rs
mskrzypkows: Yes, I'm changing that

PR #359, opened 2026-04-29
fix: yaml format rust skill rust-p2p

iamquang95:
Fix yaml format for rust-p2p skill

### DIFF
--- a/.claude/skills/rust-p2p/SKILL.md
+++ b/.claude/skills/rust-p2p/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: rust-p2p
-description: Use when implementing, reviewing, debugging, or explaining Pluto Rust libp2p code: Node/P2PContext ownership, PlutoBehaviour composition, NetworkBehaviour and ConnectionHandler protocols, relay/force-direct/quic-upgrade behaviours, peerinfo/parsigex protocol handlers, DKG protocol handlers, and P2P tests.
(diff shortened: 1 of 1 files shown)


.claude/skills/rust-p2p/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: rust-p2p
-description: Use when implementing, reviewing, debugging, or explaining Pluto Rust libp2p code: Node/P2PContext ownership, PlutoBehaviour composition, NetworkBehaviour and ConnectionHandler protocols, relay/force-direct/quic-upgrade behaviours, peerinfo/parsigex protocol handlers, DKG protocol handlers, and P2P tests.
+description: "Use when implementing, reviewing, debugging, or explaining Pluto Rust libp2p code: Node/P2PContext ownership, PlutoBehaviour composition, NetworkBehaviour and ConnectionHandler protocols, relay/force-direct/quic-upgrade behaviours, peerinfo/parsigex protocol handlers, DKG protocol handlers, and P2P tests."
```
varex83, line 3: I would keep it as it is, I think it's more conventional using without ""

iamquang95: the `:` inside description break the yaml format, that's why I need to wrap it in `""`

varex83 (approved):

PR #360, opened 2026-04-29
fix: dkg proto files

varex83:
This PR makes proto files structure compatible with what we have in charon


crates/dkg/src/dkgpb/v1/sync.proto
```diff
@@ -1,6 +1,6 @@
 syntax = "proto3";
 
-package sync;
+package dkg.dkgpb.v1;
```
iamquang95, line 4: Prefer to keep the `option go_package = "github.com/obolnetwork/charon/dkg/dkgpb/v1";`  (copied the whole file from charon). Which is easier to keep the proto synced with charon.

iamquang95 (approved):

PR #361, opened 2026-04-29
fix: apply fixes from bohdan/finish-dkg

varex83:
## Summary

This PR fixes several bugs discovered during DKG integration testing, primarily around relay connection lifecycle management in the P2P layer.

### Relay reconnection on disconnect / dial failure (`relay.rs`)

Previously, if a relay connection dropped or a dial attempt failed, the node would not attempt to re-establish the connection, leaving it permanently unable to route traffic through that relay.

- Added `relay_peers: HashMap<PeerId, Peer>` to `MutableRelayReservation` to remember all known relay peers so they can be re-dialed when a connection is lost.
- Added `connected_relays: HashSet<PeerId>` to skip redundant dials when a connection is already established or in-flight.
- `on_swarm_event` now handles `ConnectionClosed` (last connection dropped) and `DialFailure` to trigger a re-dial, clearing stale pending/connected state first.
- Switched relay dials from `DialOpts::unknown_peer_id()` to `DialOpts::peer_id(...).condition(DisconnectedAndNotDialing)` so libp2p can deduplicate concurrent dial attempts.

### Relay-ready settling delay (`relay.rs`)

`RelayRouter` was immediately attempting to route peers through a relay the moment the connection was established, before the relay reservation handshake could complete. This caused circuit-dial attempts to fail silently.

- Added `connected_relays: HashMap<PeerId, Instant>` to `RelayRouter` to record when each relay became connected.
- Added `relay_ready()` which gates routing behind a 2-second `RELAY_READY_DELAY` so the reservation handshake has time to finish.
- `RelayRouter::run_relay_router` now iterates over `connected_relays` (only relays with an active connection) instead of all configured relays, and skips any that haven't settled yet.
- Peer dials via relay also use `DisconnectedAndNotDialing` to avoid redundant dial attempts.

### Fix malformed circuit multiaddresses (`relay.rs`, `utils.rs`)

Relay peer addresses stored in the `Peer` struct sometimes already included a trailing `/p2p/<peer-id>` component. Appending another `/p2p/<relay-id>/p2p-circuit` (or `/p2p-circuit/p2p/<target-id>`) on top produced invalid multiaddresses that libp2p silently rejected.

- Both `queue_relay_dial` and `multi_addrs_via_relay` now strip any existing `/p2p/...` protocol components from the base address before constructing circuit or direct-dial addresses.

### Fix `encode_0x_hex` for empty input (`serde_utils.rs`)

`encode_0x_hex(&[])` previously returned `"0x"` instead of `""`, which caused downstream deserialization failures for optional byte fields encoded as empty strings.


crates/p2p/src/relay.rs
```diff
@@ -206,6 +244,40 @@ impl NetworkBehaviour for MutableRelayReservation {
...
+                        &peer,
+                    );
+                }
+            }
+            FromSwarm::DialFailure(ev) => {
+                if let Some(peer_id) = ev.peer_id
+                    && let Some(peer) = self.relay_peers.get(&peer_id).cloned()
+                {
+                    self.pending_relays.remove(&peer_id);
+                    Self::queue_relay_dial(
```
iamquang95: This should have a back-off delay: https://github.com/ObolNetwork/charon/blob/v1.7.1/p2p/relay.go#L47

crates/ssz/src/serde_utils.rs
```diff
@@ -21,7 +21,11 @@ pub fn trim_0x_prefix(value: &str) -> &str {
 }
 
 /// Encodes bytes as lowercase `0x`-prefixed hex.
+/// In case of empty bytes, returns an empty string.
```
iamquang95: This change will affect the `eth2util` serialization.
From here: https://github.com/NethermindEth/pluto/blob/main/crates/eth2api/src/spec/bellatrix.rs#L121

The bellatrix `extra_data` use the `Hex0x` which uses `encode_0x_hex`. 

In `go-eth2-client`, this field expect "0x" when the `extra_data` is empty: https://github.com/attestantio/go-eth2-client/blob/master/spec/bellatrix/executionpayload.go#L91

varex83:
@claude review it

varex83:
> crates/p2p/src/relay.rs
Bug: RelayRouter::poll() strands events for up to 60 seconds
After run_relay_router() fills self.events, the function falls through to Poll::Pending without returning any of those events. Since no immediate waker is armed after the interval consumes its tick, the swarm won't re-poll RelayRouter until the next interval tick 60 seconds later — meaning every batch of relay-circuit dials is delayed by a full interval.

I think that's how it's done in ping implementation in libp2p, so we should be okay with that, wdyt @iamquang95 ?

iamquang95:
@varex83 I would implememnt by this
```
if let Some(event) = self.events.pop_front() {
    return Poll::Ready(event);
}

if self.interval.poll_tick(cx).is_ready() {
    self.run_relay_router();
}

if let Some(event) = self.events.pop_front() {
    return Poll::Ready(event);
}

Poll::Pending
```

the rust-libp2p push to event on `on_connection_handler_event` then `poll()` and drain the event https://github.com/libp2p/rust-libp2p/blob/22fb4c784fc55ad8b15d05fdc9f98d663107d4cb/protocols/ping/src/lib.rs#L143

But the relay-router add `Dial` events inside `poll()` after i`nterval.poll_tick()` returns ready, then returns `Pending` without draining them.

iamquang95 (approved):

varex83:
@claude please review

iamquang95 (approved):

PR #366, opened 2026-05-03
feat: implement dkg/frost

### LINKED ISSUE #254: Implement dkg/frost

iamquang95:
Fix: https://github.com/NethermindEth/pluto/issues/254

### DIFF
--- a/crates/dkg/src/frost.rs
+++ b/crates/dkg/src/frost.rs
@@ -5,7 +5,7 @@ use std::collections::{BTreeMap, HashMap};
 use async_trait::async_trait;
 use pluto_crypto::{
     tblsconv::{privkey_from_bytes, pubkey_from_bytes},
-    types::{PrivateKey, PublicKey},
+    types::PublicKey,
 };
 use pluto_frost::{
     G1Affine, G1Projective, KeyPackage,
@@ -15,6 +15,8 @@ use pluto_frost::{
 use tokio_util::sync::CancellationToken;
 
 use crate::share::Share;
+
+type Round1Output = (BTreeMap<MsgKey, Round1Bcast>, BTreeMap<MsgKey, ShamirShare>);
 
 /// Identifies the source and target nodes and validator index the message
 /// belongs to.
@@ -34,7 +36,7 @@ pub(crate) struct MsgKey {
 
 /// Abstracts the transport of frost DKG messages.
 #[async_trait]
-pub(crate) trait FTransport: Sync {
+pub(crate) trait FTransport: Send + Sync {
     /// Returns results of all round 1 communication: the received round 1
     /// broadcasts from all other nodes and the round 1 P2P sends to this
     /// node.
@@ -43,7 +45,7 @@ pub(crate) trait FTransport: Sync {
         cancellation: CancellationToken,
         bcast: BTreeMap<MsgKey, Round1Bcast>,
         shares: BTreeMap<MsgKey, ShamirShare>,
-    ) -> Result<(BTreeMap<MsgKey, Round1Bcast>, BTreeMap<MsgKey, ShamirShare>), FrostError>;
+    ) -> Result<Round1Output, FrostError>;
 
     /// Returns results of all round 2 communication: the received round 2
     /// broadcasts from all other nodes.
@@ -66,12 +68,6 @@ pub(crate) enum FrostError {
     /// Failed during local round 2 execution.
     #[error("exec round 2: {0}")]
     ExecRound2(#[source] pluto_frost::kryptology::KryptologyError),
-    /// Transport failed during round 1.
-    #[error("transport round 1: {0}")]
-    TransportRound1(#[source] Box<FrostError>),
-    /// Transport failed during round 2.
-    #[error("transport round 2: {0}")]
-    TransportRound2(#[source] Box<FrostError>),
     /// Failed to convert public key bytes.
     #[error("public key conversion: {0}")]
     PublicKey(#[from] pluto_crypto::tblsconv::ConvError),
@@ -165,6 +161,8 @@ impl DkgParticipant {
             .round1_secret
             .take()
             .ok_or(FrostError::MissingRoundState)?;
+        // get_round2_inputs keeps this node's broadcast. Strip it here to
+        // match Charon's participant behavior; kryptology::round2 rejects self IDs.
         let bcasts = bcasts
             .iter()
             .filter(|(id, _)| **id != self.id)
@@ -198,16 +196,10 @@ pub(crate) async fn run_frost_parallel<T: FTransport>(
         new_frost_participants(num_validators, num_nodes, threshold, share_idx, dkg_ctx)?;
 
     let (cast_r1, p2p_r1) = round1(&mut validators)?;
-    let (cast_r1_result, p2p_r1_result) = tp
-        .round1(cancellation.clone(), cast_r1, p2p_r1)
-        .await
-        .map_err(|e| FrostError::TransportRound1(Box::new(e)))?;
+    let (cast_r1_result, p2p_r1_result) = tp.round1(cancellation.clone(), cast_r1, p2p_r1).await?;
 
     let cast_r2 = round2(&mut validators, &cast_r1_result, &p2p_r1_result)?;
-    let cast_r2_result = tp
-        .round2(cancellation, cast_r2)
-        .await
-        .map_err(|e| FrostError::TransportRound2(Box::new(e)))?;
+    let cast_r2_result = tp.round2(cancellation, cast_r2).await?;
 
     make_shares(&validators, &cast_r2_result)
 }
@@ -239,10 +231,7 @@ fn other_ids(num_nodes: u32, share_idx: u32) -> Vec<u32> {
 
 /// Executes round 1 for each validator and returns all round 1
 /// broadcast and p2p messages for all validators.
-#[allow(clippy::type_complexity)]
-fn round1(
-    validators: &mut BTreeMap<u32, DkgParticipant>,
-) -> Result<(BTreeMap<MsgKey, Round1Bcast>, BTreeMap<MsgKey, ShamirShare>), FrostError> {
+fn round1(validators: &mut BTreeMap<u32, DkgParticipant>) -> Result<Round1Output, FrostError> {
     let mut cast_results = BTreeMap::new();
     let mut p2p_results = BTreeMap::new();
 
@@ -348,7 +337,7 @@ fn make_shares(
 
         shares.push(Share {
             pub_key: point_to_pubkey(G1Affine::from(pub_key).to_compressed())?,
-            secret_share: scalar_to_secret_share(&kryptology::scalar_to_be(&secret_share))?,
+            secret_share: privkey_from_bytes(&kryptology::scalar_to_be(&secret_share))?,
             public_shares: pub_shares.get(&v_idx).cloned().unwrap_or_default(),
         });
     }
@@ -357,13 +346,10 @@ fn make_shares(
 }
 
 fn point_to_pubkey(point: [u8; 48]) -> Result<PublicKey, FrostError> {
-    // `pubkey_from_bytes` only checks length; transport bytes still need G1 validation.
+    // `pubkey_from_bytes` only checks length; transport bytes still need G1
+    // validation.
     G1Projective::from_compressed(&point).ok_or(FrostError::InvalidPublicKeyPoint)?;
     Ok(pubkey_from_bytes(&point)?)
-}
-
-fn scalar_to_secret_share(scalar: &[u8; 32]) -> Result<PrivateKey, FrostError> {
-    Ok(privkey_from_bytes(scalar)?)
 }
 
 fn validate_participant_inputs(
@@ -383,6 +369,8 @@ fn validate_participant_inputs(
 }
 
 fn dkg_context_byte(dkg_ctx: &str) -> u8 {
+    // Match Charon's strconv.Atoi(ctx) with ignored errors. Production passes a
+    // hex definition hash ("0x..."), so this intentionally becomes 0.
     dkg_ctx
         .parse::<isize>()
         .map(|value| value.to_le_bytes()[0])
@@ -398,7 +386,6 @@ mod tests {
 
     use super::*;
 
-    #[derive(Default)]
     struct FrostMemTransport {
         nodes: usize,
         inner: Mutex<FrostMemTransportInner>,
@@ -431,8 +418,7 @@ mod tests {
             cancellation: CancellationToken,
             bcast: BTreeMap<MsgKey, Round1Bcast>,
             shares: BTreeMap<MsgKey, ShamirShare>,
-        ) -> Result<(BTreeMap<MsgKey, Round1Bcast>, BTreeMap<MsgKey, ShamirShare>), FrostError>
-        {
+        ) -> Result<Round1Output, FrostError> {
             let source_id = bcast
                 .keys()
                 .next()
@@ -442,6 +428,11 @@ mod tests {
 
             {
                 let mut inner = self.inner.lock().await;
+                if inner.round1 == self.nodes {
+                    inner.round1 = 0;
+                    inner.round1_bcast.clear();
+                    inner.round1_shares.clear();
+                }
                 for (key, round1_bcast) in bcast {
                     inner.round1_bcast.insert(
                         MsgKey {
@@ -496,6 +487,10 @@ mod tests {
         ) -> Result<BTreeMap<MsgKey, Round2Bcast>, FrostError> {
             {
                 let mut inner = self.inner.lock().await;
+                if inner.round2 == self.nodes {
+                    inner.round2 = 0;
+                    inner.round2_bcast.clear();
+                }
                 for (key, round2_bcast) in bcast {
                     inner.round2_bcast.insert(
                         MsgKey {
@@ -532,26 +527,47 @@ mod tests {
 
     #[tokio::test]
     async fn frost_dkg() {
-        let cancellation = CancellationToken::new();
         const NODES: u32 = 3;
         const THRESHOLD: u32 = 3;
         const VALS: u32 = 2;
 
+        let node_shares = run_mem_dkg(NODES, THRESHOLD, VALS).await;
+        verify_returned_shares(
+            &node_shares,
+            usize::try_from(THRESHOLD).expect("threshold should fit"),
+        );
+    }
+
+    #[tokio::test]
+    async fn frost_dkg_partial_quorum() {
+        const NODES: u32 = 3;
+        const THRESHOLD: u32 = 2;
+        const VALS: u32 = 2;
+
+        let node_shares = run_mem_dkg(NODES, THRESHOLD, VALS).await;
+        verify_returned_shares(
+            &node_shares,
+            usize::try_from(THRESHOLD).expect("threshold should fit"),
+        );
+    }
+
+    async fn run_mem_dkg(nodes: u32, threshold: u32, vals: u32) -> Vec<Vec<Share>> {
+        let cancellation = CancellationToken::new();
         let tp = Arc::new(FrostMemTransport::new(
-            usize::try_from(NODES).expect("nodes should fit"),
+            usize::try_from(nodes).expect("nodes should fit"),
         ));
 
         let mut tasks = Vec::new();
-        for i in 0..NODES {
+        for i in 0..nodes {
             let tp = Arc::clone(&tp);
             let cancellation = cancellation.clone();
             tasks.push(tokio::spawn(async move {
                 run_frost_parallel(
                     cancellation,
                     tp.as_ref(),
-                    VALS,
-                    NODES,
-                    THRESHOLD,
+                    vals,
+                    nodes,
+                    threshold,
                     i.checked_add(1).expect("share index should not overflow"),
                     "0",
                 )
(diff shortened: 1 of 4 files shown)


iamquang95:
@claude

crates/dkg/src/frost.rs
```diff
@@ -0,0 +1,691 @@
...
+
+/// Abstracts the transport of frost DKG messages.
+#[async_trait]
+pub(crate) trait FTransport: Sync {
+    /// Returns results of all round 1 communication: the received round 1
+    /// broadcasts from all other nodes and the round 1 P2P sends to this
+    /// node.
+    async fn round1(
+        &self,
+        cancellation: CancellationToken,
```
varex83: Can we pass it by reference?
```suggestion
        cancellation: &CancellationToken,
```

iamquang95:
`dkg_context_byte vs Go's actual context encoding (frost.rs:378–383)` Seems not a correct finding.
Go flow:
`defHash := fmt.Sprintf("%#x", def.DefinitionHash)` (charon/dkg/dkg.go:205)
`runFrostParallel(........, ctx = defHash)` (charon/dkg/dkg.go:286-287)
And in https://github.com/coinbase/kryptology/blob/master/pkg/dkg/frost/participant.go#L55
`strconv.Atoi(ctx)` and ignores the error

There is no `[]byte(ctx)[0]`

crates/dkg/src/frost.rs
```diff
@@ -0,0 +1,691 @@
+#![allow(dead_code)]
```
varex83agent, line 1: **Integration gap**: `run_frost_parallel` is not yet called from `dkg.rs`. The Go equivalent is dispatched at `dkg/dkg.go:286` inside `case "default", "frost":`. The module is declared `mod frost` (private) in `lib.rs` and `#![allow(dead_code)]` suppresses the compiler warning, masking this gap entirely. Until `dkg.rs` dispatches to `frost::run_frost_parallel`, the entire module is dead in production.

This attribute should be removed once the wiring is in place (or this PR should include the wiring). Landing `#![allow(dead_code)]` on hand-written code as a permanent state is against Pluto conventions.

crates/dkg/src/frost.rs
```diff
@@ -0,0 +1,691 @@
...
+    ) -> Result<BTreeMap<MsgKey, Round2Bcast>, FrostError>;
+}
+
+/// FROST DKG orchestration errors.
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum FrostError {
+    /// Failed to construct a participant.
+    #[error("new participant: {0}")]
+    NewParticipant(#[source] pluto_frost::kryptology::KryptologyError),
+    /// Failed during local round 1 execution.
```
varex83agent, line 66: `FrostError::TransportRound1` and `TransportRound2` both wrap `Box<FrostError>` — a recursive self-referential error. Since `FTransport::round1/2` already return `FrostError`, these variants add a layer of boxing without adding any new type information. Callers matching on transport errors must unwrap the box first, and the `#[source]` chain becomes a self-referential loop in tooling.

Consider two alternatives:
- Add plain context marker variants without a payload, propagating the inner error directly via `?`
- Or keep the wrapping but use `Box<dyn std::error::Error + Send + Sync>` so the source type is explicit

crates/dkg/src/frost.rs
```diff
@@ -0,0 +1,691 @@
...
+    /// Identifies the distributed validator (Ith parallel participant) the
+    /// message belongs to. It is 0-indexed.
+    pub(crate) val_idx: u32,
+    /// Identifies the source node/participant ID of the message.
+    /// It is 1-indexed and equivalent to `cluster.NodeIdx.ShareIdx`.
+    pub(crate) source_id: u32,
+    /// Identifies the target node/participant ID of the message.
+    /// It is 1-indexed and equivalent to `cluster.NodeIdx.ShareIdx`.
+    /// The zero value indicates outgoing broadcast messages.
+    pub(crate) target_id: u32,
```
varex83agent, line 35: `FTransport` requires `Sync` but not `Send`. For trait object usage in multi-threaded Tokio (`Arc<dyn FTransport>`), the underlying type must be both `Send + Sync`. Without a `Send` bound, `Arc<dyn FTransport>` is not `Send` (since `Arc<T>: Send` requires `T: Send + Sync`), preventing it from being sent across await points in `tokio::spawn` tasks.

The tests compile because `FrostMemTransport` happens to be `Send + Sync`, but a `dyn FTransport` object in production would not be usable with `tokio::spawn` without this bound.

Fix: `pub(crate) trait FTransport: Send + Sync`

crates/dkg/src/frost.rs
```diff
@@ -0,0 +1,691 @@
...
+            })
+            .collect::<Result<_, _>>()?;
+
+        Ok((cast, shares))
+    }
+
+    fn round2(
+        &mut self,
+        bcasts: &BTreeMap<u32, Round1Bcast>,
+        shares: &BTreeMap<u32, ShamirShare>,
```
varex83agent, line 174: `DkgParticipant::round2` removes `self.id` from `bcasts` and `shares` before calling `kryptology::round2`. This is necessary because the Rust `kryptology::round2` returns `Err(InvalidParticipantId)` if own ID is present in the received bcasts — unlike Go's `DkgParticipant.Round2` which silently skips `id == dp.Id`.

The correctness of this removal depends on an implicit contract with `get_round2_inputs`: that function does NOT pre-filter own ID (it passes all bcasts for the `v_idx`), so the removal here is the only place own ID is stripped. If `get_round2_inputs` is later changed to pre-filter, or if `kryptology::round2` is changed to accept own ID, one side breaks silently.

Please add a comment documenting this invariant.

crates/dkg/src/frost.rs
```diff
@@ -0,0 +1,691 @@
...
+
+    Ok(())
+}
+
+fn dkg_context_byte(dkg_ctx: &str) -> u8 {
+    dkg_ctx
+        .parse::<isize>()
+        .map(|value| value.to_le_bytes()[0])
+        .unwrap_or_default()
+}
```
varex83agent, line 426: `dkg_context_byte` silently returns `0` for any string that is not a valid decimal integer. In production, Charon passes the cluster definition hash as `dgkCtx` — a hex string like `"0x3f4a..."` which always fails `parse::<isize>()` and yields `ctx = 0`.

This means:
1. All production DKG sessions unconditionally use context byte `0`, defeating the domain-separation purpose of the context entirely.
2. The failure is silent — no log, no error.

This matches the Go source behavior (`strconv.Atoi` also silently returns `0` on failure), so it's not a regression. However, it's a latent design flaw worth tracking. At minimum, add a comment explaining the production context is always `0` due to hex input. Longer term, consider passing context as `u8` directly from the caller.

crates/dkg/src/frost.rs
```diff
@@ -0,0 +1,691 @@
...
+    let other_ids = other_ids(num_nodes, share_idx);
+    let mut participants = BTreeMap::new();
+
+    for v_idx in 0..num_validators {
+        participants.insert(
+            v_idx,
+            DkgParticipant::new(share_idx, threshold, dkg_ctx, other_ids.clone())?,
+        );
+    }
+
```
varex83agent, line 272: `#[allow(clippy::type_complexity)]` should be replaced with a type alias. The return type `(BTreeMap<MsgKey, Round1Bcast>, BTreeMap<MsgKey, ShamirShare>)` is used both here and in the `FTransport::round1` signature. Extracting it as:
```rust
type Round1Output = (BTreeMap<MsgKey, Round1Bcast>, BTreeMap<MsgKey, ShamirShare>);
```
removes the allow attribute and makes `FTransport::round1` and this function share a named type, improving readability.

crates/dkg/src/frost.rs
```diff
@@ -0,0 +1,691 @@
...
+    Ok(shares)
+}
+
+fn point_to_pubkey(point: [u8; 48]) -> Result<PublicKey, FrostError> {
+    G1Projective::from_compressed(&point).ok_or(FrostError::InvalidPublicKeyPoint)?;
+    Ok(pubkey_from_bytes(&point)?)
+}
+
```
varex83agent, line 402: nit: `point_to_pubkey` decompresses the G1 point via `G1Projective::from_compressed` purely for validation, then immediately discards the result and calls `pubkey_from_bytes` with the original bytes. This performs the decompression (including subgroup check) twice.

If `pubkey_from_bytes` already validates the compressed encoding (which it should for correctness), the `G1Projective::from_compressed` call is redundant. If it does not validate, add a comment explaining why the manual validation step is needed before the `pubkey_from_bytes` call.

crates/dkg/src/frost.rs
```diff
@@ -0,0 +1,691 @@
...
+fn scalar_to_secret_share(scalar: &[u8; 32]) -> Result<PrivateKey, FrostError> {
+    Ok(privkey_from_bytes(scalar)?)
+}
+
+fn validate_participant_inputs(
```
varex83agent, line 403: nit: `scalar_to_secret_share` is a trivial wrapper around `privkey_from_bytes`. The `?` operator already handles the error conversion via the `#[from] ConvError` impl on `FrostError`. The function can be inlined at its single call site in `make_shares`:
```rust
secret_share: privkey_from_bytes(&kryptology::scalar_to_be(&secret_share))?,
```

crates/dkg/src/frost.rs
```diff
@@ -0,0 +1,691 @@
...
+mod tests {
+    use std::sync::Arc;
+
+    use pluto_crypto::{blst_impl::BlstImpl, tbls::Tbls, types::Index};
+    use tokio::sync::{Mutex, Notify};
+
+    use super::*;
+
+    #[derive(Default)]
+    struct FrostMemTransport {
```
varex83agent, line 437: `FrostMemTransport` derives `Default`, giving it `nodes: 0`. Because the condition `inner.round1 == self.nodes` is checked on entry to the wait loop, a default-constructed transport (0 == 0) immediately returns empty maps to every caller — no error, no panic, silently broken DKG output.

`FrostMemTransportInner` legitimately needs `Default` (used in `Mutex::new(…::default())`). The outer `FrostMemTransport` does not — every test uses `FrostMemTransport::new(n)`. Remove `#[derive(Default)]` from `FrostMemTransport` or replace it with a `fn default() -> Self { panic!("use FrostMemTransport::new(nodes)") }` to guard against accidental use.

crates/dkg/src/frost.rs
```diff
@@ -0,0 +1,691 @@
...
+            &self,
+            cancellation: CancellationToken,
+            bcast: BTreeMap<MsgKey, Round1Bcast>,
+            shares: BTreeMap<MsgKey, ShamirShare>,
+        ) -> Result<(BTreeMap<MsgKey, Round1Bcast>, BTreeMap<MsgKey, ShamirShare>), FrostError>
+        {
+            let source_id = bcast
+                .keys()
+                .next()
+                .map(|key| key.source_id)
```
varex83agent, line 473: `FrostMemTransport` round counters (`inner.round1`, `inner.round2`) are never reset. The Go reference transport (`frost_internal_test.go` lines 64–68) resets its maps when `round1 == 0`, making it reusable across multiple DKG runs.

If `run_frost_parallel` were called a second time on the same transport instance, `inner.round1` is already equal to `self.nodes`, so every caller immediately returns stale data from the previous run. The current tests each create fresh transports so this is not currently a problem, but it diverges from the Go reference and is a footgun. Add a comment documenting the single-use constraint, or reset counters at the start of each round call (matching the Go pattern).

crates/dkg/src/frost.rs
```diff
@@ -0,0 +1,691 @@
...
+
+        let mut tasks = Vec::new();
+        for i in 0..NODES {
+            let tp = Arc::clone(&tp);
+            let cancellation = cancellation.clone();
+            tasks.push(tokio::spawn(async move {
+                run_frost_parallel(
+                    cancellation,
+                    tp.as_ref(),
+                    VALS,
```
varex83agent: Both `frost_dkg` and `make_shares_sorts_by_validator_index_and_maps_public_shares` use `threshold = NODES = 3` (3-of-3). This is the degenerate case requiring all participants. The most common production configuration is a strict-majority threshold (e.g., 2-of-3 or 3-of-5).

No test covers a threshold strictly less than the node count, leaving the partial-quorum path completely untested. Add a 2-of-3 variant of `frost_dkg`.

crates/dkg/src/frost.rs
```diff
@@ -0,0 +1,691 @@
...
+        assert_eq!(dkg_context_byte("1"), 1);
+        assert_eq!(dkg_context_byte("257"), 1);
+        assert_eq!(dkg_context_byte("-1"), 255);
+    }
+
+    #[test]
+    fn point_to_pubkey_rejects_invalid_compressed_point() {
+        let invalid_but_correct_length = [42u8; 48];
+
+        assert!(matches!(
```
varex83agent, line 808: `get_round2_inputs_filters_by_validator_index` creates participants for a single node (`share_idx=1`). All round-1 share keys therefore have `source_id=1`. When `get_round2_inputs` maps them to `source_id → share` for validator 0, both entries (one for target 2 and one for target 3) collide on key `1`, producing `v0_shares.len() == 1`.

The test asserts `len == 1`, which passes — but only because of the collision from single-node input, not because the validator-index filter is doing correct work. In production, `p2p_r1_result` contains shares from N-1 distinct source nodes, so `v0_shares.len()` would be N-1 = 2.

The test should be rewritten with a manually constructed multi-source `BTreeMap` that directly exercises the filter (multiple source IDs for each validator index).

crates/dkg/src/frost.rs
```diff
@@ -0,0 +1,691 @@
...
+
+        for share_idx in 1..=3 {
+            let tp = Arc::clone(&tp);
+            let cancellation = cancellation.clone();
+            tasks.push(tokio::spawn(async move {
+                run_frost_parallel(cancellation, tp.as_ref(), 3, 3, 3, share_idx, "0").await
+            }));
+        }
+
+        let shares = tasks
```
varex83agent: `tasks.remove(0)` extracts one node's shares for assertion; the other two tasks are awaited (to catch panics) but their share vectors are dropped and never inspected. If `make_shares` had a per-node bug in `public_shares` indexing that only manifests on nodes 2 or 3, this test would miss it.

Collect and assert all three nodes' share vectors.

crates/dkg/src/frost.rs
```diff
@@ -0,0 +1,691 @@
...
+            self.notify.notify_waiters();
+
+            loop {
+                let notified = self.notify.notified();
+                {
+                    let inner = self.inner.lock().await;
+                    if inner.round1 == self.nodes {
+                        return Ok((
+                            inner.round1_bcast.clone(),
+                            inner
```
varex83agent, line 515: nit: No test covers the cancellation path. The `select! { _ = cancellation.cancelled() => return Err(FrostError::Cancelled) }` branch has no corresponding test.

A test should start a multi-node DKG with fewer callers than `self.nodes` (so the barrier never completes), cancel the token, and assert that `run_frost_parallel` returns `Err(FrostError::TransportRound1(…))` wrapping `FrostError::Cancelled`.

varex83agent (dismissed):
This PR implements the FROST DKG orchestration layer (`crates/dkg/src/frost.rs`) that parallels Go's `dkg/frost.go`. The implementation is functionally correct and the end-to-end test validates cryptographic correctness via threshold signature aggregation. Two major concerns block merge: the module is not yet wired into `dkg.rs` and the file-level `#![allow(dead_code)]` masks this integration gap; and the test transport's `Default` derive creates a silently broken zero-node state. Several minor findings cover test coverage gaps (no partial-quorum test, degenerate `get_round2_inputs` test, no cancellation test) and design concerns (`FrostError` recursive boxing, undocumented invariant in `round2`). No bugs found in the crypto path; Go parity is confirmed.

crates/dkg/src/frost.rs
iamquang95: This is not included in this PR. Should be on other PR

crates/dkg/src/frost.rs
iamquang95: Added comment

crates/dkg/src/frost.rs
```diff
@@ -0,0 +1,751 @@
+#![allow(dead_code)]
```
varex83agent, line 1: **Still open from prior review.** `#![allow(dead_code)]` is still present and `run_frost_parallel` is still not called from `dkg.rs`. The module is private (`mod frost` in `lib.rs`) and fully unreachable in production.

Options:
1. Wire `frost::run_frost_parallel` into `dkg.rs` in this PR and remove the attribute.
2. If the integration is intentionally deferred, replace the file-level allow with a targeted `#[allow(dead_code)]` on the specific items that are not yet used, or open a tracking issue and note it in a `// TODO(#NNN):` comment — landing a blanket file-level lint suppression on hand-written code is against Pluto conventions.

varex83agent (dismissed):
The new commit addresses 13 of 14 original findings thoroughly. `FTransport: Send + Sync` fixed; `TransportRound1/2` recursive boxing removed (direct `?` propagation); `round2` implicit contract documented with an explanatory comment and rewritten as a filter; `dkg_context_byte` annotated; `FrostMemTransport::default()` footgun removed; transport counters reset on reuse; partial-quorum test (2-of-3) added; `get_round2_inputs` test rewritten with realistic multi-source input; all nodes' outputs verified in `make_shares` test; cancellation test added; `scalar_to_secret_share` inlined; `Round1Output` type alias introduced.

One finding remains open: `#![allow(dead_code)]` is still at line 1 and `run_frost_parallel` is still not wired into `dkg.rs`. The module is dead in production. This must be removed before merge — either by wiring the call site or by tracking the gap with a `// TODO:` and a follow-up issue rather than a file-level lint suppression.

crates/dkg/src/frost.rs
```diff
@@ -0,0 +1,751 @@
...
+    MissingRoundState,
+    /// Failed to convert a numeric value to the target representation.
+    #[error(transparent)]
+    IntConversion(#[from] std::num::TryFromIntError),
+    /// Cancellation was requested while waiting for transport data.
+    #[error("frost dkg cancelled")]
+    Cancelled,
+}
+
+struct DkgParticipant {
```
varex83, line 92: I think this struct misses "round" field, which is used in some additional checks in the `.round` functions

crates/dkg/src/frost.rs
```diff
@@ -0,0 +1,751 @@
...
+    let mut validators =
+        new_frost_participants(num_validators, num_nodes, threshold, share_idx, dkg_ctx)?;
+
+    let (cast_r1, p2p_r1) = round1(&mut validators)?;
+    let (cast_r1_result, p2p_r1_result) = tp.round1(cancellation.clone(), cast_r1, p2p_r1).await?;
+
+    let cast_r2 = round2(&mut validators, &cast_r1_result, &p2p_r1_result)?;
+    let cast_r2_result = tp.round2(cancellation, cast_r2).await?;
+
+    make_shares(&validators, &cast_r2_result)
```
varex83: No debug logs there, would nice to have them

crates/dkg/src/frost.rs
```diff
@@ -0,0 +1,751 @@
...
+        bcast: BTreeMap<MsgKey, Round1Bcast>,
+        shares: BTreeMap<MsgKey, ShamirShare>,
+    ) -> Result<Round1Output, FrostError>;
+
+    /// Returns results of all round 2 communication: the received round 2
+    /// broadcasts from all other nodes.
+    async fn round2(
+        &self,
+        cancellation: CancellationToken,
+        bcast: BTreeMap<MsgKey, Round2Bcast>,
```
varex83: Is there and specific reason we want to use `BTreeMap` here? 

I think in charon it's done with standard `map` which is `HashMap`

varex83 (approved):
LGTM overall, few nits and comments

crates/dkg/src/frost.rs
iamquang95: Good catch :+1:

crates/dkg/src/frost.rs
iamquang95: Only `validators` need order, so I will keep it as `BTreeMap`, others will be `HashMap`

crates/dkg/src/frost.rs
```diff
@@ -0,0 +1,855 @@
...
+    #[error(transparent)]
+    IntConversion(#[from] std::num::TryFromIntError),
+    /// Cancellation was requested while waiting for transport data.
+    #[error("frost dkg cancelled")]
+    Cancelled,
+}
+
+struct DkgParticipant {
+    id: u32,
+    round: u8,
```
mskrzypkows, line 94: I think that keeping the round number is redundant. The information about the stage of the DKG is kept by the ` Option<Round1Secret>`. If it's `None` we know that it should be round 1, if it's `Some` we know that it should be round 2.

iamquang95: I put it explicitly here for easier reasoning, also matched with kryptology implementation.

mskrzypkows: In this case I think that we need to verify both variables. In round1 there is no check for `round1_secret.is_none()`.

iamquang95: We will eventually override the `round1_secret` in round1 and `key_package` in round2. But I think it's fine to have a check on these field on round1 and round2. 
I will add that.

mskrzypkows (approved):

PR #367, opened 2026-05-04
feat: implement dkg/frostp2p

### LINKED ISSUE #255: Implement dkg/frostp2p
### Description
Implement dkg/frostp2p

### Acceptance criteria

iamquang95:
Fix: https://github.com/NethermindEth/pluto/issues/255

### DIFF
--- a/crates/dkg/src/frost.rs
+++ b/crates/dkg/src/frost.rs
@@ -78,6 +78,15 @@ pub(crate) enum FrostError {
     /// A FROST protobuf message failed validation.
     #[error("{0}")]
     InvalidMessage(&'static str),
+    /// FROST transport configuration is invalid.
+    #[error("frost config: {0}")]
+    ConfigError(&'static str),
+    /// FROST transport internal state is invalid.
+    #[error("frost internal state: {0}")]
+    InternalState(&'static str),
+    /// FROST transport channel closed unexpectedly.
+    #[error("frost channel closed: {0}")]
+    ChannelClosed(&'static str),
     /// Failed to convert public key bytes.
     #[error("public key conversion: {0}")]
     PublicKey(#[from] pluto_crypto::tblsconv::ConvError),
--- a/crates/dkg/src/frost.rs
+++ b/crates/dkg/src/frost.rs
@@ -84,9 +84,6 @@ pub(crate) enum FrostError {
     /// Too many round-2 broadcast messages were collected.
     #[error("too many round 2 casts messages")]
     TooManyRound2CastsMessages,
-    /// The number of validators cannot be represented locally.
-    #[error("invalid number of validators")]
-    InvalidNumberOfValidators,
     /// A FROST message key was missing.
     #[error("frost msg key cannot be nil")]
     MissingMsgKey,
--- a/crates/dkg/src/frost.rs
+++ b/crates/dkg/src/frost.rs
@@ -42,7 +42,7 @@ pub(crate) trait FTransport: Send + Sync {
     /// broadcasts from all other nodes and the round 1 P2P sends to this
     /// node.
     async fn round1(
-        &self,
+        &mut self,
         cancellation: &CancellationToken,
         bcast: HashMap<MsgKey, Round1Bcast>,
         shares: HashMap<MsgKey, ShamirShare>,
@@ -51,7 +51,7 @@ pub(crate) trait FTransport: Send + Sync {
     /// Returns results of all round 2 communication: the received round 2
     /// broadcasts from all other nodes.
     async fn round2(
-        &self,
+        &mut self,
         cancellation: &CancellationToken,
         bcast: HashMap<MsgKey, Round2Bcast>,
     ) -> Result<HashMap<MsgKey, Round2Bcast>, FrostError>;
@@ -69,6 +69,15 @@ pub(crate) enum FrostError {
     /// Failed during local round 2 execution.
     #[error("exec round 2: {0}")]
     ExecRound2(#[source] pluto_frost::kryptology::KryptologyError),
+    /// Reliable broadcast failed.
+    #[error("bcast: {0}")]
+    Bcast(#[from] crate::bcast::Error),
+    /// Direct FROST P2P send failed.
+    #[error("frost p2p send: {0}")]
+    FrostP2P(#[from] crate::frostp2p::FrostP2PError),
+    /// A FROST protobuf message failed validation.
+    #[error("{0}")]
+    InvalidMessage(&'static str),
     /// Failed to convert public key bytes.
     #[error("public key conversion: {0}")]
     PublicKey(#[from] pluto_crypto::tblsconv::ConvError),
@@ -211,7 +220,7 @@ impl DkgParticipant {
 /// rounds) and returns a list of shares (one for each distributed validator).
 pub(crate) async fn run_frost_parallel<T: FTransport>(
     cancellation: CancellationToken,
-    tp: &T,
+    tp: &mut T,
     num_validators: u32,
     num_nodes: u32,
     threshold: u32,
@@ -460,9 +469,9 @@ mod tests {
     }
 
     #[async_trait]
-    impl FTransport for FrostMemTransport {
+    impl FTransport for Arc<FrostMemTransport> {
         async fn round1(
-            &self,
+            &mut self,
             cancellation: &CancellationToken,
             bcast: HashMap<MsgKey, Round1Bcast>,
             shares: HashMap<MsgKey, ShamirShare>,
@@ -529,7 +538,7 @@ mod tests {
         }
 
         async fn round2(
-            &self,
+            &mut self,
             cancellation: &CancellationToken,
             bcast: HashMap<MsgKey, Round2Bcast>,
         ) -> Result<HashMap<MsgKey, Round2Bcast>, FrostError> {
@@ -607,12 +616,12 @@ mod tests {
 
         let mut tasks = Vec::new();
         for i in 0..nodes {
-            let tp = Arc::clone(&tp);
+            let mut tp = Arc::clone(&tp);
             let cancellation = cancellation.clone();
             tasks.push(tokio::spawn(async move {
                 run_frost_parallel(
                     cancellation,
-                    tp.as_ref(),
+                    &mut tp,
                     vals,
                     nodes,
                     threshold,
@@ -645,11 +654,11 @@ mod tests {
         let tp = Arc::new(FrostMemTransport::new(2));
 
         let task = {
-            let tp = Arc::clone(&tp);
+            let mut tp = Arc::clone(&tp);
             let cancellation = cancellation.clone();
-            tokio::spawn(async move {
-                run_frost_parallel(cancellation, tp.as_ref(), 1, 2, 2, 1, "0").await
-            })
+            tokio::spawn(
+                async move { run_frost_parallel(cancellation, &mut tp, 1, 2, 2, 1, "0").await },
+            )
         };
 
         tokio::task::yield_now().await;
@@ -845,10 +854,10 @@ mod tests {
         let mut tasks = Vec::new();
 
         for share_idx in 1..=3 {
-            let tp = Arc::clone(&tp);
+            let mut tp = Arc::clone(&tp);
             let cancellation = cancellation.clone();
             tasks.push(tokio::spawn(async move {
-                run_frost_parallel(cancellation, tp.as_ref(), 3, 3, 3, share_idx, "0").await
+                run_frost_parallel(cancellation, &mut tp, 3, 3, 3, share_idx, "0").await
             }));
         }
 
--- a/crates/dkg/src/frost.rs
+++ b/crates/dkg/src/frost.rs
@@ -75,15 +75,72 @@ pub(crate) enum FrostError {
     /// Direct FROST P2P send failed.
     #[error("frost p2p send: {0}")]
     FrostP2P(#[from] crate::frostp2p::FrostP2PError),
-    /// A FROST protobuf message failed validation.
-    #[error("{0}")]
-    InvalidMessage(&'static str),
+    /// Too many round-1 broadcast messages were collected.
+    #[error("too many round 1 casts messages")]
+    TooManyRound1CastsMessages,
+    /// Too many round-1 direct P2P messages were collected.
+    #[error("too many round 1 p2p messages")]
+    TooManyRound1P2PMessages,
+    /// Too many round-2 broadcast messages were collected.
+    #[error("too many round 2 casts messages")]
+    TooManyRound2CastsMessages,
+    /// The number of validators cannot be represented locally.
+    #[error("invalid number of validators")]
+    InvalidNumberOfValidators,
+    /// A FROST message key was missing.
+    #[error("frost msg key cannot be nil")]
+    MissingMsgKey,
+    /// A round-1 P2P message source ID was invalid.
+    #[error("invalid round 1 p2p source ID")]
+    InvalidRound1P2PSourceId,
+    /// A round-1 P2P message target ID was invalid.
+    #[error("invalid round 1 p2p target ID")]
+    InvalidRound1P2PTargetId,
+    /// A round-1 P2P message validator index was invalid.
+    #[error("invalid round 1 p2p validator index")]
+    InvalidRound1P2PValidatorIndex,
+    /// A round-1 P2P message did not contain exactly one share per validator.
+    #[error("invalid round 1 p2p shares count")]
+    InvalidRound1P2PSharesCount,
+    /// A round-1 P2P message repeated a validator index.
+    #[error("duplicate round 1 p2p validator index")]
+    DuplicateRound1P2PValidatorIndex,
+    /// Failed to decode the round-1 Wi scalar.
+    #[error("decode wi scalar")]
+    DecodeWiScalar,
+    /// Failed to decode the round-1 C_i scalar.
+    #[error("decode c1 scalar")]
+    DecodeC1Scalar,
+    /// Failed to decode a round-1 commitment point.
+    #[error("decode commitment")]
+    DecodeCommitment,
+    /// Failed to decode a round-1 Shamir share scalar.
+    #[error("decode shamir scalar")]
+    DecodeShamirScalar,
+    /// Failed to decode a round-2 verification key point.
+    #[error("decode verification key scalar")]
+    DecodeVerificationKeyScalar,
+    /// Failed to decode a round-2 verification key share point.
+    #[error("decode vk share")]
+    DecodeVkShare,
     /// FROST transport configuration is invalid.
     #[error("frost config: {0}")]
     ConfigError(&'static str),
-    /// FROST transport internal state is invalid.
-    #[error("frost internal state: {0}")]
-    InternalState(&'static str),
+    /// The FROST P2P inbound receiver was already taken.
+    #[error("frost p2p inbound receiver already taken")]
+    P2PInboundReceiverAlreadyTaken,
+    /// The FROST broadcast event receiver was already taken.
+    #[error("frost bcast event receiver already taken")]
+    BcastEventReceiverAlreadyTaken,
+    /// The round-1 casts receiver was dropped before local self-delivery.
+    #[error("frost round 1 casts receiver dropped before self-delivery")]
+    Round1CastsReceiverDropped,
+    /// The round-2 casts receiver was dropped before local self-delivery.
+    #[error("frost round 2 casts receiver dropped before self-delivery")]
+    Round2CastsReceiverDropped,
+    /// Round-1 P2P construction attempted to send a private share to self.
+    #[error("unexpected p2p message to self")]
+    UnexpectedP2PMessageToSelf,
     /// FROST transport channel closed unexpectedly.
     #[error("frost channel closed: {0}")]
     ChannelClosed(&'static str),
--- a/crates/dkg/src/frostp2p.rs
+++ b/crates/dkg/src/frostp2p.rs
@@ -7,9 +7,9 @@
 //!   reliable broadcast through [`bcast::Component`].
 //!
 //! The outer DKG network behaviour must install both `bcast::Behaviour` and
-//! [`FrostP2PBehaviour`]. It must also forward every [`bcast::Event`] emitted by
-//! `bcast::Behaviour` into [`FrostP2PHandle::handle_bcast_event`]. Without that
-//! event bridge, [`FrostP2P`] cannot observe broadcast completion and
+//! [`FrostP2PBehaviour`]. It must also forward every [`bcast::Event`] emitted
+//! by `bcast::Behaviour` into [`FrostP2PHandle::handle_bcast_event`]. Without
+//! that event bridge, [`FrostP2P`] cannot observe broadcast completion and
 //! `round1`/`round2` will wait until cancellation.
 
 #![allow(dead_code)]
@@ -352,8 +352,8 @@ impl FrostP2PHandle {
     /// Forwards bcast completion events into the FROST round state machine.
     ///
     /// The outer DKG network behaviour must call this for every event emitted
-    /// by `bcast::Behaviour`; `FrostP2P::round1` and `FrostP2P::round2` block on
-    /// these forwarded events after starting their broadcasts.
+    /// by `bcast::Behaviour`; `FrostP2P::round1` and `FrostP2P::round2` block
+    /// on these forwarded events after starting their broadcasts.
     pub(crate) fn handle_bcast_event(&self, event: bcast::Event) -> Result<(), FrostError> {
         self.bcast_event_tx
             .send(event)
--- a/crates/dkg/src/frostp2p.rs
+++ b/crates/dkg/src/frostp2p.rs
@@ -1,6 +1,88 @@
 //! FROST DKG P2P transport.
 //!
-//! This module is split across two integration surfaces:
+//! This module provides the network transport used by `frost.rs`. The local
+//! FROST code creates cryptographic round messages; this module moves those
+//! messages between cluster nodes over libp2p.
+//!
+//! Round 1 has a public broadcast path and a private direct-P2P path:
+//!
+//! ```text
+//! ROUND 1
+//! =======
+//!
+//! Public broadcast, same data to everyone:
+//!
+//!   node1 === Round1Bcast(commitments/proof) ===> node1,node2,node3,node4
+//!   node2 === Round1Bcast(commitments/proof) ===> node1,node2,node3,node4
+//!   node3 === Round1Bcast(commitments/proof) ===> node1,node2,node3,node4
+//!   node4 === Round1Bcast(commitments/proof) ===> node1,node2,node3,node4
+//!
+//! Private direct P2P, different data per target:
+//!
+//!              +-- ShamirShare(for share_idx 2) --> node2
+//!   node1 -----+-- ShamirShare(for share_idx 3) --> node3
+//!              +-- ShamirShare(for share_idx 4) --> node4
+//!
+//!              +-- ShamirShare(for share_idx 1) --> node1
+//!   node2 -----+-- ShamirShare(for share_idx 3) --> node3
+//!              +-- ShamirShare(for share_idx 4) --> node4
+//!
+//!   ... same pattern for node3 and node4.
+//!
+//! Each direct message contains the private shares for that target node across
+//! all validators in the DKG run. The shares cannot be broadcast because they
+//! are secret, and node X does not send the same share to every peer.
+//! ```
+//!
+//! Round 2 is broadcast-only. After round 1, each node has public commitments
+//! from all nodes and private shares sent specifically to itself. It verifies
+//! those private shares and broadcasts public verification material:
+//!
+//! ```text
+//! ROUND 2
+//! =======
+//!
+//!   node1 === Round2Bcast(public verification shares) ===> node1,node2,node3,node4
+//!   node2 === Round2Bcast(public verification shares) ===> node1,node2,node3,node4
+//!   node3 === Round2Bcast(public verification shares) ===> node1,node2,node3,node4
+//!   node4 === Round2Bcast(public verification shares) ===> node1,node2,node3,node4
+//!
+//! No direct P2P is needed in round 2 because there is no new per-target secret.
+//! ```
+//!
+//! End-to-end this module bridges the async FROST transport API to libp2p's
+//! event-driven swarm:
+//!
+//! ```text
+//! run_frost_parallel
+//!        |
+//!        v
+//!   FrostP2P::round1
+//!        |-----------------------> bcast::Component
+//!        |                              |
+//!        |                              v
+//!        |                       bcast::Behaviour
+//!        |                              |
+//!        |                              v
+//!        |                    FrostP2PHandle::handle_bcast_event
+//!        |
+//!        +-----------------------> FrostP2PSender
+//!                                       |
+//!                                       v
+//!                                FrostP2PBehaviour
+//!                                       |
+//!                                       v
+//!                                FrostP2PHandler
+//!                                       |
+//!                                       v
+//!                              direct libp2p streams
+//!
+//!   FrostP2P::round2
+//!        |
+//!        +-----------------------> bcast::Component
+//! ```
+//!
+//! The module is split across two integration surfaces:
 //!
 //! - [`FrostP2PBehaviour`] owns the direct round-1 P2P libp2p protocol.
 //! - [`FrostP2P`] implements the FROST transport by combining direct P2P with
--- a/crates/dkg/src/frostp2p.rs
+++ b/crates/dkg/src/frostp2p.rs
@@ -84,6 +84,7 @@ pub(crate) enum FrostP2PError {
 #[derive(Debug)]
 pub(crate) enum InEvent {
     Send { op_id: u64, msg: FrostRound1P2p },
+    CancelPending,
 }
 
 #[derive(Debug)]
@@ -161,8 +162,10 @@ impl ConnectionHandler for FrostP2PHandler {
     }
 
     fn on_behaviour_event(&mut self, event: Self::FromBehaviour) {
-        let InEvent::Send { op_id, msg } = event;
-        self.pending_open.push_back((op_id, msg));
+        match event {
+            InEvent::Send { op_id, msg } => self.pending_open.push_back((op_id, msg)),
+            InEvent::CancelPending => self.pending_open.clear(),
+        }
     }
 
     fn poll(
@@ -264,11 +267,16 @@ async fn write_outbound_message(
     })
 }
 
+type SendResultTx = oneshot::Sender<Result<(), FrostP2PError>>;
+
 #[derive(Debug)]
-struct SendCommand {
-    peer_id: PeerId,
-    msg: FrostRound1P2p,
-    result_tx: oneshot::Sender<Result<(), FrostP2PError>>,
+enum SendCommand {
+    Send {
+        peer_id: PeerId,
+        msg: FrostRound1P2p,
+        result_tx: SendResultTx,
+    },
+    CancelAll,
 }
 
 /// User-facing FROST direct-P2P sender.
@@ -278,17 +286,37 @@ pub(crate) struct FrostP2PSender {
 }
 
 impl FrostP2PSender {
+    fn new(cmd_tx: mpsc::UnboundedSender<SendCommand>) -> Self {
+        Self { cmd_tx }
+    }
+
     /// Sends a round-1 P2P message to `peer_id` and waits for stream delivery.
-    pub async fn send(&self, peer_id: PeerId, msg: &FrostRound1P2p) -> Result<(), FrostP2PError> {
+    pub async fn send(
+        &self,
+        peer_id: PeerId,
+        msg: &FrostRound1P2p,
+        cancellation: &CancellationToken,
+    ) -> Result<(), FrostError> {
+        if cancellation.is_cancelled() {
+            return Err(FrostError::Cancelled);
+        }
+
         let (result_tx, result_rx) = oneshot::channel();
         self.cmd_tx
-            .send(SendCommand {
+            .send(SendCommand::Send {
                 peer_id,
                 msg: msg.clone(),
                 result_tx,
             })
             .map_err(|_| FrostP2PError::BehaviourClosed)?;
-        result_rx.await.map_err(|_| FrostP2PError::ResultClosed)?
+
+        tokio::select! {
+            _ = cancellation.cancelled() => {
+                let _ = self.cmd_tx.send(SendCommand::CancelAll);
+                Err(FrostError::Cancelled)
+            }
+            result = result_rx => Ok(result.map_err(|_| FrostP2PError::ResultClosed)??),
+        }
     }
 }
 
@@ -331,7 +359,7 @@ pub(crate) struct FrostP2PBehaviour {
     cmd_rx: mpsc::UnboundedReceiver<SendCommand>,
     pending_events: VecDeque<ToSwarm<(), InEvent>>,
     pending_by_peer: HashMap<PeerId, VecDeque<(u64, FrostRound1P2p)>>,
-    result_by_op: HashMap<u64, (PeerId, oneshot::Sender<Result<(), FrostP2PError>>)>,
+    result_by_op: HashMap<u64, (PeerId, SendResultTx)>,
     next_op_id: u64,
 }
 
@@ -341,7 +369,7 @@ impl FrostP2PBehaviour {
         let (inbound_tx, inbound_rx) = mpsc::unbounded_channel();
         let (cmd_tx, cmd_rx) = mpsc::unbounded_channel();
         let (bcast_event_tx, bcast_event_rx) = mpsc::unbounded_channel();
-        let sender = FrostP2PSender { cmd_tx };
+        let sender = FrostP2PSender::new(cmd_tx);
         (
             Self {
                 p2p_context,
@@ -377,10 +405,18 @@ impl FrostP2PBehaviour {
 
     fn drain_commands(&mut self, cx: &mut Context<'_>) {
         while let Poll::Ready(Some(command)) = self.cmd_rx.poll_recv(cx) {
-            let op_id = self.next_op_id();
-            self.result_by_op
-                .insert(op_id, (command.peer_id, command.result_tx));
-            self.enqueue_send(command.peer_id, op_id, command.msg);
+            match command {
+                SendCommand::Send {
+                    peer_id,
+                    msg,
+                    result_tx,
+                } => {
+                    let op_id = self.next_op_id();
+                    self.result_by_op.insert(op_id, (peer_id, result_tx));
+                    self.enqueue_send(peer_id, op_id, msg);
+                }
+                SendCommand::CancelAll => self.cancel_all_sends(),
+            }
         }
     }
 
@@ -422,6 +458,36 @@ impl FrostP2PBehaviour {
     fn complete_send(&mut self, op_id: u64, result: Result<(), FrostP2PError>) {
         if let Some((_peer_id, result_tx)) = self.result_by_op.remove(&op_id) {
             let _ = result_tx.send(result);
+        }
+    }
+
+    fn cancel_all_sends(&mut self) {
+        let peers = self
+            .result_by_op
+            .values()
+            .map(|(peer_id, _)| *peer_id)
+            .collect::<HashSet<_>>();
+        self.result_by_op.clear();
+        self.pending_by_peer.clear();
+        self.pending_events.retain(|event| {
+            !matches!(
+                event,
+                ToSwarm::NotifyHandler {
+                    event: InEvent::Send { .. },
+                    ..
+                }
+            )
+        });
+
+        for peer_id in peers {
+            // This can only cancel handler-local queued opens. If the handler has
+            // already started writing, stale completion is ignored because all
+            // result waiters were removed above.
+            self.pending_events.push_back(ToSwarm::NotifyHandler {
+                peer_id,
+                handler: NotifyHandler::Any,
+                event: InEvent::CancelPending,
+            });
         }
     }
 
@@ -735,10 +801,7 @@ impl FTransport for FrostP2P {
 
         let p2p_msgs = self.build_round1_p2p_by_peer(&shares)?;
         for (peer_id, msg) in p2p_msgs {
-            tokio::select! {
-                _ = cancellation.cancelled() => return Err(FrostError::Cancelled),
-                result = self.frost_sender.send(peer_id, &msg) => result?,
-            }
+            self.frost_sender.send(peer_id, &msg, cancellation).await?;
         }
 
         let mut cast_msgs = Vec::with_capacity(self.num_peers);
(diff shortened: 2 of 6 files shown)


iamquang95:
@claude

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,1458 @@
...
+                }
+                msg = self.round1_p2p_rx.recv() => {
+                    let (peer_id, msg) = msg.ok_or(FrostError::InvalidMessage("round 1 p2p channel closed"))?;
+                    validate_round1_p2p(
+                        peer_id,
+                        &self.share_idx_by_peer,
+                        self.local_share_idx,
+                        &msg,
+                        self.num_validators,
+                    )?;
```
varex83agent: **bug:** A single malformed inbound P2P message aborts the entire DKG. `validate_round1_p2p(...)?` propagates `FrostError::InvalidMessage` out of the `loop` and out of `round1`, ending the DKG for every cluster member.

In Charon (frostp2p.go:209-238), the equivalent validation runs inside `newP2PCallback` at the per-stream handler level. On failure the handler returns an error which closes that stream; the round-1 channel write `round1P2PRecv <- msg` never happens, so the DKG round 1 simply waits for the missing message (and only times out via ctx). One malicious or buggy peer cannot abort the round.

Fix: move per-message validation into `FrostP2PBehaviour::on_connection_handler_event` (or into `read_inbound_message`), discard invalid messages with a `warn!` log, and forward only validated messages over `inbound_tx`. Then this `select!` arm needs no validation at all (or only as defense-in-depth that does not abort).

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,1458 @@
...
+impl FrostP2PHandler {
+    fn new() -> Self {
+        Self {
+            pending_open: VecDeque::new(),
+            active_futures: FuturesUnordered::new(),
+        }
+    }
+
+    fn handle_fully_negotiated_inbound(&mut self, mut stream: Stream) {
+        self.active_futures.push(
```
varex83agent, line 280: **major:** Unbounded inbound substream futures permit per-connection memory exhaustion.

`handle_fully_negotiated_inbound` unconditionally pushes a future into `active_futures` for every fully-negotiated `/charon/dkg/frost/2.0.0/round1/p2p` substream. Each future calls `read_protobuf_with_max_size` with `MAX_MESSAGE_SIZE = 128 MiB`, and `pluto_p2p::proto::read_length_delimited` allocates `vec![0_u8; msg_len]` upfront based solely on the peer-supplied varint length (proto.rs:44). Until the peer finishes streaming or the 5s `RECEIVE_TIMEOUT` fires, that allocation persists. A single authenticated cluster peer can open many concurrent substreams declaring 128 MiB lengths and slow-stream them, multiplying memory by N substreams.

Recommendations: cap concurrent inbound futures per-connection (or per-peer), and tighten `MAX_MESSAGE_SIZE` for this protocol — the round-1 P2P payload is bounded by `num_validators * ~50 bytes`, so a few KiB to a hundred KiB cap is realistic.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,1458 @@
...
+            "frost bcast event receiver already taken",
+        ))
+    }
+}
+
+/// libp2p behaviour for FROST round-1 direct P2P.
+pub(crate) struct FrostP2PBehaviour {
+    peers: HashSet<PeerId>,
+    p2p_context: P2PContext,
+    inbound_tx: mpsc::UnboundedSender<(PeerId, FrostRound1P2p)>,
```
varex83agent: **major:** `inbound_tx` is an unbounded mpsc that begins queueing decoded `FrostRound1P2p` messages the moment the behaviour is added to the swarm. The receiver is only taken inside `new_frost_p2p`, and only drained inside `round1`. There is no cap on the number of queued messages or their cumulative decoded size.

Combined with the `validate-then-?` ordering in `round1` (separate finding) and the 128 MiB inbound size cap (separate finding), a single peer can pre-load many large inbound messages between behaviour construction and the round-1 `recv` loop.

Fix: bound the channel (e.g. one slot per known peer) and/or reject inbound messages outside the round-1 acceptance window.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,1458 @@
...
+    /// Receives `(sender_peer_id, message)` for inbound round-1 P2P messages.
+    inbound_rx: Option<mpsc::UnboundedReceiver<(PeerId, FrostRound1P2p)>>,
+    sender: FrostP2PSender,
+    bcast_event_tx: mpsc::UnboundedSender<bcast::Event>,
+    bcast_event_rx: Option<mpsc::UnboundedReceiver<bcast::Event>>,
+}
+
+impl FrostP2PHandle {
+    /// Forwards bcast completion events into the FROST round state machine.
+    pub(crate) fn handle_bcast_event(&self, event: bcast::Event) -> Result<(), FrostError> {
```
varex83agent, line 511: **major:** `FrostP2PHandle::handle_bcast_event` is the only path that feeds events into `bcast_event_rx`, which `wait_for_bcast_completion` (line 938) blocks on. Nothing in this PR (or in `PlutoBehaviour` integration) calls `handle_bcast_event`, so `round1` and `round2` will hang forever after `broadcast()`.

The file-level `#![allow(dead_code)]` (line 3) silently hides that the FROST transport is currently unreachable end-to-end. Either land the orchestration wiring in this PR (preferred) or document the integration contract on `handle_bcast_event` and reference the follow-up issue. At minimum, please call out in the PR description that this is staged.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,1458 @@
...
+    context: &'static str,
+    bytes: &Bytes,
+) -> Result<[u8; N], FrostError> {
+    bytes
+        .as_ref()
+        .try_into()
+        .map_err(|_| FrostError::InvalidMessage(context))
+}
+
+#[cfg(test)]
```
varex83agent, line 1543: **major:** Missing parity unit tests for the validation callbacks. Charon's `frostp2p_internal_test.go` covers every branch: `TestBcastCallback` exercises invalid round1/round2 source ID, target ID, validator index, commitments count, and nil casts message; `TestP2PCallback` exercises invalid P2P source ID, target ID, validator index, and nil P2P message. None of these are reproduced for `register_round1_bcast` (lines 752-777), `register_round2_bcast` (lines 813-832), or `validate_round1_p2p` (lines 990-1013).

Additionally, no end-to-end test wires two `FrostP2PBehaviour` instances into swarms and exercises the full Send → Dial → ConnectionEstablished → flush_pending_for_peer → OutboundSubstreamRequest → write/read → OutEvent::Sent/Received cycle. Given the amount of state-machine and waker logic in `FrostP2PBehaviour::poll`, `enqueue_send`, `flush_pending_for_peer`, `fail_peer_sends`, and `cancel_all_sends`, a swarm-level test (similar to those in `bcast/behaviour.rs`) is needed to catch regressions.

Note: Charon's 'invalid cast round'/'invalid round 1 casts message' cases do not directly port because Pluto uses typed `register_message::<M>()`; call that out as a deliberate API divergence rather than skipped coverage.

crates/dkg/src/frost.rs
```diff
@@ -69,6 +69,15 @@ pub(crate) enum FrostError {
...
     ExecRound2(#[source] pluto_frost::kryptology::KryptologyError),
+    /// Reliable broadcast failed.
+    #[error("bcast: {0}")]
+    Bcast(#[from] crate::bcast::Error),
+    /// Direct FROST P2P send failed.
+    #[error("frost p2p send: {0}")]
+    FrostP2P(#[from] crate::frostp2p::FrostP2PError),
+    /// A FROST protobuf message failed validation.
+    #[error("{0}")]
+    InvalidMessage(&'static str),
```
varex83agent: `FrostError::InvalidMessage(&'static str)` is used both for adversarial-input cases (`"invalid round 1 p2p source ID"`, `"frost msg key cannot be nil"`) and for purely internal-bug paths in `frostp2p.rs` such as `"frost p2p inbound receiver already taken"` (line 351), `"bug: unexpected p2p message to self"` (line 966), `"frost bcast event channel closed"` (lines 344, 947), and `"round 1 casts channel closed"` (line 878).

Logging or alerting that filters on `InvalidMessage` to attribute remote misbehaviour will see false positives from these internal paths and vice versa. Consider splitting into `InternalState` / `ConfigError` / `ChannelClosed` / `InvalidMessage`, or at minimum a structured discriminator.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,1458 @@
...
+            Box::new(|_, _| Ok(())),
+            Box::new(move |peer_id, _, msg| {
+                let tx = tx.clone();
+                let dedup = dedup.clone();
+                let share_idx_by_peer = share_idx_by_peer.clone();
+                Box::pin(async move {
+                    {
+                        let mut dedup = dedup
+                            .lock()
+                            .map_err(|_| bcast::Error::InvalidMessage("dedup mutex poisoned"))?;
```
varex83agent: Bcast dedup uses `std::sync::Mutex<HashSet<PeerId>>`. If a panic ever poisons the mutex, the closure maps the `PoisonError` to `bcast::Error::InvalidMessage("dedup mutex poisoned")` — a local invariant violation is then attributed to the remote peer and surfaced as a validation error.

Use `parking_lot::Mutex` (no poisoning) or recover via `PoisonError::into_inner` so dedup keeps working after non-fatal panics, and avoid attributing local errors to peers. Same applies to `register_round2_bcast` at line 806.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,1458 @@
...
+
+    fn cancel_all_sends(&mut self) {
+        let peers = self
+            .result_by_op
+            .values()
+            .map(|(peer_id, _)| *peer_id)
+            .collect::<HashSet<_>>();
+        self.result_by_op.clear();
+        self.pending_by_peer.clear();
+        self.pending_events.retain(|event| {
```
varex83agent: `cancel_all_sends` filters out `ToSwarm::NotifyHandler { event: InEvent::Send { .. }, .. }` events but leaves `ToSwarm::Dial` events queued. Residual dials still fire after cancellation, briefly opening connections that have no pending data to send (`flush_pending_for_peer` finds `pending_by_peer` empty because it was just cleared). Net effect is a few wasted dials.

Consider also retaining-out matching `ToSwarm::Dial` entries (or filtering only those whose peer was in the cancelled set), or document at the call site that the wasted dial is intentional.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,1458 @@
...
+                    let op_id = self.next_op_id();
+                    self.result_by_op.insert(op_id, (peer_id, result_tx));
+                    self.enqueue_send(peer_id, op_id, msg);
+                }
+                SendCommand::CancelAll => self.cancel_all_sends(),
+            }
+        }
+    }
+
+    fn enqueue_send(&mut self, peer_id: PeerId, op_id: u64, msg: FrostRound1P2p) {
```
varex83agent, line 639: `enqueue_send` does not validate that `peer_id` is in `self.peers`. If a Send command arrives for an unknown-but-connected peer, `connection_handler_for_peer` returned `dummy::ConnectionHandler`, which silently discards the routed `Either::Left(InEvent::Send { .. })`. The op_id stays in `result_by_op` until either the connection closes (which fires `fail_peer_sends`) or the behaviour itself is dropped — meanwhile the awaiting `FrostP2PSender::send` hangs indefinitely (modulo cancellation).

Production callers only send to peers from `peers_by_share_idx`, but a defensive check (`return PeerNotConnected` or a new `UnknownPeer` error) would prevent silent hangs after future refactors.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,1458 @@
...
+                msg = self.round1_p2p_rx.recv() => {
+                    let (peer_id, msg) = msg.ok_or(FrostError::InvalidMessage("round 1 p2p channel closed"))?;
+                    validate_round1_p2p(
+                        peer_id,
+                        &self.share_idx_by_peer,
+                        self.local_share_idx,
+                        &msg,
+                        self.num_validators,
+                    )?;
+                    if !p2p_seen.insert(peer_id) {
```
varex83agent: After fixing the validate-then-? ordering (separate finding), `p2p_seen` dedup is keyed only on `PeerId`, which is first-write-wins per peer. A peer that races a stale-but-valid message (e.g. recycled `val_idx < num_validators`) before the legitimate one would pin the stale message into the round.

Charon dedups on `PeerId` too (frostp2p.go:228) so this matches Charon, but the round-1 P2P payload contains shares for multiple validators identified by `val_idx`. Consider tighter keying (e.g. `(peer_id, val_idx)` per-share, or asserting that the first message from each peer carries shares for all validators) for stronger guarantees.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,1458 @@
...
+                            .map_err(|_| bcast::Error::InvalidMessage("dedup mutex poisoned"))?;
+                        if !dedup.insert(peer_id) {
+                            debug!(%peer_id, "ignoring duplicate round 1 message");
+                            return Ok(());
+                        }
+                    }
+
+                    let source_id = *share_idx_by_peer
+                        .get(&peer_id)
+                        .ok_or(bcast::Error::InvalidPeerIndex(peer_id))?;
```
varex83agent: When `register_round1_bcast` (and `register_round2_bcast` at line 815) receives a message from a peer not in `share_idx_by_peer`, Pluto returns `bcast::Error::InvalidPeerIndex(peer_id)`. Charon (frostp2p.go:150) instead does `peers[pID].ShareIdx`, which Go zero-initializes to 0 for unknown peers; the next check `0 != source_id` then triggers `"invalid round 1 cast source ID"`.

The bcast layer itself filters non-cluster peers, so this branch is largely defensive — but if it fires, the operator-visible error will not match Charon's. Either return `InvalidMessage("invalid round 1 cast source ID")` for parity, or document the deliberate variant divergence in a code comment.

crates/dkg/src/lib.rs
```diff
@@ -23,6 +23,9 @@ pub mod dkg;
 /// Kryptology-compatible FROST DKG orchestration.
 mod frost;
 
+/// FROST DKG P2P transport.
+pub mod frostp2p;
```
varex83agent: `pub mod frostp2p` declares the path public, but every item in `frostp2p.rs` (constants, types, free functions) is `pub(crate)`. The module currently exports no public items.

Either downgrade to `pub(crate) mod frostp2p` to match the contents and the sibling `mod frost;` (line 24, private), or, if downstream users will eventually consume `FrostP2P` / `FrostP2PHandle` / `new_frost_p2p`, mark those items `pub` (or `pub(crate)`-with-public-re-export) at the same time.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,1458 @@
...
+        let (result_tx, result_rx) = oneshot::channel();
+        self.cmd_tx
+            .send(SendCommand::Send {
+                peer_id,
+                msg: msg.clone(),
+                result_tx,
+            })
+            .map_err(|_| FrostP2PError::BehaviourClosed)?;
+
+        tokio::select! {
```
varex83agent, line 485: `tokio::select!` here (and in `round1` / `round2` loops at lines 875 / 924, and in `wait_for_bcast_completion` at line 944) is not biased. If both arms become ready in the same poll, a successful send can still be reported as `Cancelled` (or a cancel can lose to a result). Add `biased;` if cancellation should win deterministically when both fire, or document that pseudo-random selection is acceptable here.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,1458 @@
...
+    FrostRound2Cast {
+        key: Some(key_to_proto(key)),
+        verification_key: Bytes::copy_from_slice(&cast.verification_key),
+        vk_share: Bytes::copy_from_slice(&cast.vk_share),
+    }
+}
+
+fn round2_cast_from_proto(cast: &FrostRound2Cast) -> Result<(MsgKey, Round2Bcast), FrostError> {
+    let verification_key = bytes_to_g1("decode verification key scalar", &cast.verification_key)?;
+    let vk_share = bytes_to_g1("decode c1 scalar", &cast.vk_share)?;
```
varex83agent: nit: this preserves a Charon typo — `bytes_to_g1("decode c1 scalar", &cast.vk_share)` mirrors `errors.Wrap(err, "decode c1 scalar")` at frostp2p.go:506, but `vk_share` is a G1 point, not a c1 scalar. Reproducing it is correct under the strict-equivalence policy, but please add a `// matches charon typo at frostp2p.go:506` comment so future readers don't 'fix' it and silently regress parity.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,1458 @@
...
+        }
+        share_idx_by_peer.insert(peer_id, share_idx);
+    }
+
+    if !peers_by_share_idx.contains_key(&local_share_idx) {
+        return Err(FrostError::InvalidMessage(
+            "local frost share index missing from peer map",
+        ));
+    }
+    if peers.len() != peers_by_share_idx.len() {
```
varex83agent: nit: `peers.len() != peers_by_share_idx.len()` is unreachable. The duplicate-share-index check at line 700 already guarantees equality (a `HashMap<PeerId, u32>` cannot contain duplicate keys, and any duplicate `share_idx` would have already returned). Remove the branch or convert to `debug_assert_eq!` for documentation.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,1458 @@
...
+        bcast: HashMap<MsgKey, Round1Bcast>,
+        shares: HashMap<MsgKey, ShamirShare>,
+    ) -> Result<(HashMap<MsgKey, Round1Bcast>, HashMap<MsgKey, ShamirShare>), FrostError> {
+        let casts_msg = build_round1_casts(&bcast);
+        self.bcast_comp
+            .broadcast(ROUND1_CAST_ID, &casts_msg)
+            .await?;
+        self.wait_for_bcast_completion(ROUND1_CAST_ID, cancellation)
+            .await?;
+        let _ = self.round1_casts_tx.send(casts_msg);
```
varex83agent: nit: `let _ = self.round1_casts_tx.send(casts_msg);` (and line 919 for round 2) discards the send result. The receiver is owned by the same `FrostP2P` struct and is never closed before this send, so the only way `send` can fail is a logic bug (e.g., receiver detached by a refactor). Prefer `.expect("frost round1 casts receiver dropped before self-delivery")` so an inadvertent refactor fails loudly instead of corrupting the round count.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,1458 @@
...
+        cancellation: &CancellationToken,
+    ) -> Result<(), FrostError> {
+        loop {
+            tokio::select! {
+                _ = cancellation.cancelled() => return Err(FrostError::Cancelled),
+                event = self.bcast_event_rx.recv() => {
+                    match event.ok_or(FrostError::InvalidMessage("frost bcast event channel closed"))? {
+                        bcast::Event::BroadcastCompleted { msg_id } if msg_id == expected_msg_id => return Ok(()),
+                        bcast::Event::BroadcastFailed { msg_id, error } if msg_id == expected_msg_id => return Err(error.into()),
+                        event => debug!(?event, expected_msg_id, "ignoring unrelated bcast event"),
```
varex83agent, line 1289: nit: `wait_for_bcast_completion` silently demotes any `BroadcastFailed` event for a different `msg_id` to `debug!`. In the current sequential round1 → round2 flow this cannot happen, but if `FrostP2P` ever shares its bcast event channel with another broadcaster, a failure for the other broadcast would be lost. Consider raising the log level to `warn!` for unexpected `BroadcastFailed`, or `debug_assert!` that no other broadcasts target this event channel.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,1458 @@
+//! FROST DKG P2P transport.
+
+#![allow(dead_code)]
```
varex83agent, line 106: nit: `#![allow(dead_code)]` is acceptable for a staged port, but it should reference an issue / follow-up PR (`// TODO(#NNN): remove once dkg orchestration wires FrostP2P`) so it doesn't outlive the integration. As-is, it masks the missing `handle_bcast_event` wiring (separate finding) and any new dead code introduced later.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,1458 @@
...
+    PeerNotConnected(PeerId),
+    /// The send result channel closed.
+    #[error("send result channel closed")]
+    ResultClosed,
+}
+
+#[derive(Debug)]
+pub(crate) enum InEvent {
+    Send { op_id: u64, msg: FrostRound1P2p },
+    CancelPending,
```
varex83agent: nit: `InEvent::CancelPending` reads as 'cancel a single pending operation' but actually clears the entire `pending_open` queue (line 174). Rename to `InEvent::CancelAllPending` (or `ClearPending`) for consistency with the user-facing `SendCommand::CancelAll` (line 286).

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,1458 @@
+//! FROST DKG P2P transport.
```
varex83agent, line 1: nit: the module-level docstring (`//! FROST DKG P2P transport.`) does not describe what the module exposes or how the layers fit together. A short paragraph explaining `FrostP2PBehaviour` (libp2p NetworkBehaviour) → `FrostP2PHandler` (per-connection handler) → `FrostP2PSender` (clone-able mpsc handle) → `FrostP2P` (transport that combines bcast + direct P2P) would orient new readers, especially given the file's length.

varex83agent (dismissed):
Solid first cut of the FROST P2P transport — protocol IDs, MsgKey conversions, scalar/G1 validators, and the bcast/p2p validation rules all line up with Charon v1.7.1. Polling, cancellation, and the swarm/handler split are well structured. Below is the punch list before merge.

**Bug (must fix):**
1. `FrostP2P::round1` validates inbound P2P messages inline in the `select!` arm and propagates failures with `?` (frostp2p.rs:892). A single malformed P2P message from one peer aborts the entire DKG, where Charon's stream-handler-level validation (frostp2p.go:209-238) only drops the offending stream. Move validation into `FrostP2PBehaviour::on_connection_handler_event` (or `read_inbound_message`) and forward only validated messages.

**Major issues:**
- Inbound substream futures are unbounded, with each future allocating up to `MAX_MESSAGE_SIZE = 128 MiB` upfront (frostp2p.rs:122 + proto.rs:44). A single authenticated peer can multiply memory by N concurrent substreams.
- `inbound_tx` is unbounded and starts queueing decoded `FrostRound1P2p` messages the moment the behaviour is added to the swarm; `take_inbound_rx` and the round-1 consume loop happen later (frostp2p.rs:366).
- `FrostP2PHandle::handle_bcast_event` has no caller wired into `PlutoBehaviour`; without it `wait_for_bcast_completion` will deadlock at runtime. The `#![allow(dead_code)]` masks that this PR's transport is currently unreachable end-to-end.
- No parity tests for the bcast/p2p validation callbacks (Charon's `TestBcastCallback` / `TestP2PCallback`) and no swarm-level test that exercises send through dial through flush_pending through write/read through completion.

**Minor / nit:** Several smaller items inline — error-variant overloading, std::sync::Mutex poisoning attributed to peers, dial events not retracted on cancel, `enqueue_send` silently hanging on unknown peers, plus a few cosmetic items.

Requesting changes for the bug and the four major items.

crates/dkg/src/frostp2p.rs
iamquang95: The charon use the `128MiB`, let's keep the same limit as it. This limit is also used on other p2p crates. Need a repo-wise strict than this.

crates/dkg/src/frostp2p.rs
iamquang95: This should call from outer Behaviour (DkgBehaviour)

crates/dkg/src/frostp2p.rs
iamquang95: This should be repo-wise changes type of Mutex

varex83:
Can you merge main please?

iamquang95:
@claude

iamquang95:
@claude review the PR

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,2141 @@
...
+        E: std::error::Error + Send + Sync + 'static,
+    {
+        let message = match error {
+            StreamUpgradeError::NegotiationFailed => "protocol negotiation failed".to_string(),
+            StreamUpgradeError::Timeout => "operation timed out".to_string(),
+            StreamUpgradeError::Io(error) => error.to_string(),
+            StreamUpgradeError::Apply(error) => error.to_string(),
+        };
+        self.active_futures
+            .push(async move { Some(OutEvent::Failed { op_id, message }) }.boxed());
```
varex83, line 313: This approach is interesting, and I like it. But I see one potential problem - we can lose the order of events, will `op_id` be sufficient for ordering of events?

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,2141 @@
...
+        self.emit_event(FrostP2PEvent::BroadcastStarted { round: 1 });
+        self.bcast_comp
+            .broadcast(ROUND1_CAST_ID, &casts_msg)
+            .await?;
+        self.wait_for_bcast_completion(ROUND1_CAST_ID, cancellation)
+            .await?;
+        if let Err(error) = self.round1_casts_tx.send(casts_msg) {
+            error!(%error, "frost round 1 casts receiver dropped before self-delivery");
+            return Err(FrostError::InternalState(
+                "frost round 1 casts receiver dropped before self-delivery",
```
varex83: Is it possible to make those errors enum-only?

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,2141 @@
...
+pub(crate) const ROUND1_CAST_ID: &str = "/charon/dkg/frost/2.0.0/round1/cast";
+/// bcast message ID for FROST round-2 broadcasts.
+pub(crate) const ROUND2_CAST_ID: &str = "/charon/dkg/frost/2.0.0/round2/cast";
+/// Direct P2P protocol for FROST round-1 Shamir share delivery.
+pub(crate) const ROUND1_P2P_PROTOCOL: StreamProtocol =
+    StreamProtocol::new("/charon/dkg/frost/2.0.0/round1/p2p");
+
+/// Maximum FROST P2P protobuf message size. Charon's default libp2p
+/// delimited-reader limit is 128 MiB.
+pub(crate) const MAX_MESSAGE_SIZE: usize = pluto_p2p::proto::MAX_MESSAGE_SIZE;
```
varex83: I think we can use `pluto_p2p::proto::MAX_MESSAGE_SIZE`

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,2141 @@
...
+/// Charon's default direct-P2P inbound read timeout.
+pub(crate) const RECEIVE_TIMEOUT: Duration = Duration::from_secs(5);
+/// Charon's default direct-P2P send timeout.
+pub(crate) const SEND_TIMEOUT: Duration = Duration::from_secs(7);
+
+const SCALAR_LEN: usize = 32;
+const G1_COMPRESSED_LEN: usize = 48;
```
varex83: This should be part of `pluto-crypto`

varex83 (approved):
LGTM!

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,2141 @@
...
+impl FrostP2PHandler {
+    fn new() -> Self {
+        Self {
+            pending_open: VecDeque::new(),
+            active_futures: FuturesUnordered::new(),
+        }
+    }
+
+    fn handle_fully_negotiated_inbound(&mut self, mut stream: Stream) {
+        self.active_futures.push(
```
varex83agent, line 280: **Unbounded inbound substreams enable memory DoS.** `handle_fully_negotiated_inbound` pushes every incoming substream's reader into the unbounded `FuturesUnordered active_futures` (line 268), and `read_inbound_message` reads up to `MAX_MESSAGE_SIZE = 128 MiB` per stream over a 5-second window. There is no per-connection or per-peer cap on concurrent inbound substreams, and the behaviour-level dedup at line 823 only fires AFTER a stream has been read.

A single authenticated FROST peer can open many round-1 P2P substreams in parallel; each reader will allocate a buffer up to 128 MiB once the prost varint length prefix is read. Worth either rejecting additional inbound substreams while one is in flight (drop streams whose reader is already running for that peer), or capping `active_futures` length and dropping new inbound streams once full.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,2141 @@
...
+            return Err(bcast::Error::InvalidMessage(
+                "invalid round 1 cast target ID",
+            ));
+        }
+        if key.val_idx >= num_validators {
+            return Err(bcast::Error::InvalidMessage(
+                "invalid round 1 cast validator index",
+            ));
+        }
+        if cast.commitments.len() != threshold {
```
varex83agent, line 1101: **`validate_round1_casts` / `validate_round2_casts` should cap `msg.casts.len()` and require unique `val_idx`.** Both validators iterate `msg.casts` and check each entry's fields, but never bound the list length or enforce uniqueness on `val_idx`. The contract is that each peer sends exactly one cast per validator (so `casts.len() == num_validators` with distinct `val_idx`), yet a malicious authenticated peer can send a `FrostRound1Casts` containing tens of thousands of casts (all with `val_idx = 0`, each carrying a `threshold`-element commitment list and 32-byte scalars) and validation will pass. Each cast then triggers `bytes_to_g1` (a non-trivial subgroup check via `G1Projective::from_compressed`) in `make_round1_response`; all but the last are silently overwritten in the result map.

This is asymmetric with `validate_round1_p2p` (line 1317-1346), which already enforces both `shares.len() == num_validators` and val_idx uniqueness. Mirror those checks here:
```rust
if msg.casts.len() != usize::try_from(num_validators)? { ... }
if !seen_val_idx.insert(key.val_idx) { ... }
```
Applies equally to `validate_round2_casts` at line 1090.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,2141 @@
...
+    tx: mpsc::UnboundedSender<FrostRound1Casts>,
+    threshold: usize,
+    num_validators: u32,
+) -> Result<(), FrostError> {
+    let dedup = Arc::new(Mutex::new(HashSet::<PeerId>::new()));
+    let share_idx_by_peer = Arc::new(share_idx_by_peer);
+    bcast_comp
+        .register_message::<FrostRound1Casts>(
+            ROUND1_CAST_ID,
+            Box::new(|_, _| Ok(())),
```
varex83agent: **`check` callback is a no-op so peers sign unvalidated FROST broadcasts.** `register_round1_bcast` (and `register_round2_bcast` at line 1025) registers `Box::new(|_, _| Ok(()))` as the `CheckFn`. That means when this node is asked to *sign* a peer's broadcast (`bcast/handler.rs:376`), it signs anything with the right msg_id without running `validate_round1_casts` / `validate_round2_casts`. Validation only runs in the post-broadcast `callback`.

Net effect: a malformed FROST broadcast can collect full signatures from every peer (wasted bandwidth and CPU) before being rejected at delivery, aborting the whole DKG. Move the cheap structural checks (source_id, target_id, val_idx range, `commitments.len() == threshold`, casts-length cap) into the `check` callback so honest peers refuse to sign in the first place — matching the spirit of Charon's `newCheckMsg`.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,2141 @@
...
+                    debug!(%peer_id, "ignoring duplicate round 1 p2p message");
+                    return;
+                }
+                match self.inbound_tx.try_send((peer_id, msg)) {
+                    Ok(()) => {
+                        self.pending_events.push_back(ToSwarm::GenerateEvent(
+                            FrostP2PEvent::DirectReceived { peer_id },
+                        ));
+                    }
+                    Err(mpsc::error::TrySendError::Full((peer_id, _))) => {
```
varex83agent: **`TrySendError::Full` rolls back dedup but the message is already lost.** When `inbound_tx.try_send` returns Full, the code removes the peer from `accepted_round1_p2p` and warns. But the libp2p stream has already delivered the message — the sender will not retry. So if the channel is ever full at delivery time, the round-1 P2P message is permanently lost and `FrostP2P::round1` deadlocks waiting for it.

Today this is unreachable because the channel cap is `peers.len().max(1)` and each peer sends only one (due to dedup), but the rollback makes the code look like a recoverable retry path that does not exist. Two options: (a) replace the rollback with a hard `FrostP2PEvent::DirectSendFailed` (or similar transport error) so the round fails fast instead of silently deadlocking; (b) raise the channel to unbounded so the failure mode cannot occur. Same applies to the `Closed` arm at line 837.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,2141 @@
...
+                            return Ok(());
+                        }
+                        bcast::Event::BroadcastFailed { msg_id, error } if msg_id == expected_msg_id => {
+                            self.emit_event(FrostP2PEvent::BroadcastFailed {
+                                round: round_for_msg_id(expected_msg_id),
+                                error: error.to_string(),
+                            });
+                            return Err(error.into());
+                        }
+                        bcast::Event::BroadcastFailed { msg_id, error } => {
```
varex83agent, line 1286: **`wait_for_bcast_completion` silently drops unrelated bcast events.** The arms at 1261-1264 swallow any `BroadcastFailed`/`BroadcastCompleted` whose `msg_id` differs from the expected one (warn/debug + continue). Today this is safe because `bcast::Behaviour` only emits completion events for the LOCAL active broadcast and `FrostP2P` issues round 1 and round 2 strictly sequentially.

Forward-looking fragility: the docstring at the top of the file already invites the outer DKG behaviour to bridge ALL `bcast::Event`s into this handle. As soon as another caller registers a different msg_id on the same `bcast::Component`, a stuck `BroadcastFailed` for that id would be discarded here and never re-delivered. Either route only FROST-relevant events into `bcast_event_tx`, buffer unrelated events, or document the single-caller assumption explicitly.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,2141 @@
...
+    share_idx_by_peer: &HashMap<PeerId, u32>,
+    local_share_idx: u32,
+    msg: &FrostRound1P2p,
+    num_validators: u32,
+) -> Result<(), FrostError> {
+    let source_id = *share_idx_by_peer
+        .get(&peer_id)
+        .ok_or(FrostError::InvalidMessage("invalid round 1 p2p source ID"))?;
+    // A direct round-1 P2P message contains this peer's share for every validator.
+    if msg.shares.len()
```
varex83agent: **Stricter than Charon — document the divergence.** Pluto requires `msg.shares.len() == num_validators` (line 1317) and rejects duplicate `val_idx` (line 1342). Charon's `newP2PCallback` (frostp2p.go:218-225) only validates `source_id`, `target_id`, and `val_idx` range — it accepts duplicates and any share count.

A correctly-running Charon sender always emits exactly num_validators distinct shares per direct message, so this stricter check does not reject any currently-valid Charon-emitted message in practice. But the project's golden rule is functional equivalence with Charon. Either relax to Charon parity, or add a doc comment on `validate_round1_p2p` explaining the intentional defense-in-depth so future cross-implementation audits can find it.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,2141 @@
...
+            Box::new(move |peer_id, _, msg| {
+                let tx = tx.clone();
+                let dedup = dedup.clone();
+                let share_idx_by_peer = share_idx_by_peer.clone();
+                Box::pin(async move {
+                    {
+                        let mut dedup = dedup
+                            .lock()
+                            .map_err(|_| bcast::Error::InvalidMessage("dedup mutex poisoned"))?;
+                        if !dedup.insert(peer_id) {
```
varex83agent: **Dedup-then-validate poisons the dedup slot on a bad first message.** In `register_round1_bcast` the dedup HashSet is updated BEFORE `validate_round1_casts` runs — same in `register_round2_bcast` at line 1035. If a peer's first delivery fails validation, the entry stays in `dedup`, and any subsequent (valid) retransmit from the same peer is silently treated as a duplicate.

This is a 1:1 port of Charon's behaviour (`dkg/frostp2p.go:142,158`), but Pluto's other dedup site for direct P2P at line 813-826 correctly does validate-then-dedup. Reordering for consistency would prevent a single bad message from permanently blocking that peer's broadcast contribution.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,2141 @@
...
+            Box::new(|_, _| Ok(())),
+            Box::new(move |peer_id, _, msg| {
+                let tx = tx.clone();
+                let dedup = dedup.clone();
+                let share_idx_by_peer = share_idx_by_peer.clone();
+                Box::pin(async move {
+                    {
+                        let mut dedup = dedup
+                            .lock()
+                            .map_err(|_| bcast::Error::InvalidMessage("dedup mutex poisoned"))?;
```
varex83agent: **`bcast::Error::InvalidMessage("dedup mutex poisoned")` mis-categorizes an internal error as peer misbehavior.** The poisoned-mutex branch at line 991 (and line 1034 for round 2) routes a host-process panic recovery through the user-facing `InvalidMessage` variant, conflating it with peer-induced protocol violations and obscuring debugging.

Add a dedicated variant (e.g. `bcast::Error::Internal(...)`), use `parking_lot::Mutex` (no poisoning), or simply panic with `expect("dedup mutex poisoned")` since the poisoning is unreachable in current code.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,2141 @@
...
+            error!(%error, "frost round 1 casts receiver dropped before self-delivery");
+            return Err(FrostError::InternalState(
+                "frost round 1 casts receiver dropped before self-delivery",
+            ));
+        }
+
+        let p2p_msgs = self.build_round1_p2p_by_peer(&shares)?;
+        self.emit_event(FrostP2PEvent::DirectSendStarted {
+            peer_count: p2p_msgs.len(),
+        });
```
varex83agent, line 1176: **Direct-P2P sends are awaited sequentially.** `for (peer_id, msg) in p2p_msgs { self.frost_sender.send(peer_id, &msg, cancellation).await?; }` blocks on each peer in turn, even though the sends are independent and use a unique-stream-per-send protocol.

This matches Charon's serialized loop (frostp2p.go:289-298), so it preserves parity, but a slow peer will serialize the whole round and `?` aborts on the first failed send (so subsequent peers may never receive their share). Drive the sends with `FuturesUnordered` / `try_join_all` so they overlap; if you want to keep parity with Charon, leave a comment justifying it.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,2141 @@
...
+
+/// libp2p behaviour for FROST round-1 direct P2P.
+pub(crate) struct FrostP2PBehaviour {
+    peers: HashSet<PeerId>,
+    p2p_context: P2PContext,
+    share_idx_by_peer: HashMap<PeerId, u32>,
+    local_share_idx: u32,
+    num_validators: u32,
+    inbound_tx: mpsc::Sender<(PeerId, FrostRound1P2p)>,
+    accepted_round1_p2p: HashSet<PeerId>,
```
varex83agent, line 550: **`accepted_round1_p2p` and bcast dedup HashSets are never reset; transport is implicitly single-use.** `FrostP2PBehaviour::accepted_round1_p2p` (initialized once at line 569) and the dedup HashSets in `register_round1_bcast` / `register_round2_bcast` are only ever inserted into. A second DKG run on the same `FrostP2PBehaviour` / `FrostP2P` instance would reject every peer's round-1 P2P and round-1/2 bcast as duplicates.

This matches Charon's single-use `frostP2P` struct (constructed per DKG run), but Pluto's behaviour is structurally a long-lived libp2p component installed in the swarm. Either document explicitly that these structs are single-use and must be rebuilt for a subsequent DKG run, or add a reset hook to clear dedup state between runs.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,2141 @@
...
+        for cast in &msg.casts {
+            let (key, cast) = round2_cast_from_proto(cast)?;
+            cast_map.insert(key, cast);
+        }
+    }
+
+    Ok(cast_map)
+}
+
+fn bytes_to_scalar(context: &'static str, bytes: &Bytes) -> Result<[u8; 32], FrostError> {
```
varex83agent: Use the named constants: `bytes_to_scalar` returns `[u8; 32]` and `bytes_to_g1` (line 1505) returns `[u8; 48]` while the file already defines `SCALAR_LEN: usize = 32` and `G1_COMPRESSED_LEN: usize = 48` (lines 165-166) and uses them in the bodies. Returning `[u8; SCALAR_LEN]` / `[u8; G1_COMPRESSED_LEN]` keeps the abstraction consistent and prevents drift if either constant changes.

crates/dkg/src/frostp2p_integ_test.rs
```diff
@@ -0,0 +1,548 @@
...
+            | FrostP2PEvent::RoundCompleted { round } => {
+                anyhow::bail!("unexpected FROST round event from node {node_index}: {round}");
+            }
+        }
+
+        Ok(())
+    }
+}
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
```
varex83agent, line 196: **Integ test only covers the happy path.** `frost_p2p_full_dkg_round_trip` exercises the 4-node / 2-validator success path. There is no integration coverage for: (a) cancellation mid-round (verifying `FrostError::Cancelled` propagation and `CancelAllPending` cleanup), (b) a peer disconnect mid-round (verifying `FrostP2PEvent::DirectSendFailed`), or (c) a malformed bcast/P2P message. Charon's `frostp2p_internal_test.go` covers validation paths at unit level (which Pluto matches), but the libp2p surface added here (`FrostP2PBehaviour` / `FrostP2PHandler`) has only unit tests for individual paths.

Additionally, several core helpers have no targeted unit coverage and are worth locking in:
- `wait_for_bcast_completion` SUCCESS path (only failure tested at line 2071)
- `make_round1_response` / `make_round2_response` decode failure modes
- `build_round1_p2p_by_peer` (`unknown target` and `unexpected p2p message to self` branches)
- `flush_pending_for_peer` after `ConnectionEstablished`
- `fail_peer_sends` for an active op
- `enqueue_send` for a connected peer (`NotifyHandler` path) and for a disconnected peer (queue + Dial)

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,2141 @@
...
+//! [`FrostP2PBehaviour`]. It must also forward every [`bcast::Event`] emitted
+//! by `bcast::Behaviour` into [`FrostP2PHandle::handle_bcast_event`]. Without
+//! that event bridge, [`FrostP2P`] cannot observe broadcast completion and
+//! `round1`/`round2` will wait until cancellation.
+//!
+//! FROST observation events are emitted through [`FrostP2PBehaviour`] as swarm
+//! events. Transport-level code forwards round and broadcast milestones back to
+//! the behaviour through the same command channel used for direct sends.
+
+#![allow(dead_code)]
```
varex83agent, line 106: nit: `#![allow(dead_code)]` masks unused public surface. Several items (`FrostP2PEvent::ProtocolCompleted`, several `FrostP2PError` variants, `new_frost_p2p`, the whole `FrostP2P*` API) are still only exercised from the integ test in this PR — `dkg.rs` does not yet wire this transport. Once the follow-up PR integrates it, drop the crate-wide allow so unused symbols surface naturally.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,2141 @@
...
+                .or_default()
+                .shares
+                .push(shamir_share_to_proto(*key, share));
+        }
+
+        Ok(p2p_msgs)
+    }
+}
+
+fn round_for_msg_id(msg_id: &'static str) -> u8 {
```
varex83agent, line 1322: nit: `round_for_msg_id` silently falls back to round 0 for unknown msg_ids. It is only called with the static `ROUND1_CAST_ID` / `ROUND2_CAST_ID` constants, but if a future change adds a new round or typos a constant the function silently emits `FrostP2PEvent::BroadcastCompleted { round: 0 }`. Prefer `Option<u8>` with an explicit `unreachable!()` / `debug_assert!` at the call site.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,2141 @@
...
+            })
+            .map_err(|_| FrostP2PError::BehaviourClosed)?;
+
+        tokio::select! {
+            biased;
+            _ = cancellation.cancelled() => {
+                let _ = self.cmd_tx.send(SendCommand::CancelAll);
+                Err(FrostError::Cancelled)
+            }
+            result = result_rx => Ok(result.map_err(|_| FrostP2PError::ResultClosed)??),
```
varex83agent, line 491: nit: `Ok(result.map_err(|_| FrostP2PError::ResultClosed)??)` is awkward — `Ok(...)` wrapping with `??`. Cleaner:
```rust
result = result_rx => result
    .map_err(|_| FrostP2PError::ResultClosed)?
    .map_err(FrostError::from),
```

varex83agent (dismissed):
This PR ports Charon's `dkg/frostp2p.go` to Rust as a new libp2p `NetworkBehaviour` + FROST transport (~2700 lines). Architecture, naming, and protocol IDs match the Go reference closely; the bcast/direct-P2P split, dedup ordering, and round-flow ordering are faithful ports. Tests are thorough at the unit level and a 4-node end-to-end integ test verifies a full DKG round-trip.

Main concerns (must address before merge):
1. **Unbounded inbound substreams** (frostp2p.rs:280) — `FrostP2PHandler` pushes every inbound substream's reader into an unbounded `FuturesUnordered`, each able to allocate up to `MAX_MESSAGE_SIZE = 128 MiB`. An authenticated peer can open many substreams in parallel and force the receiver to allocate enormous buffers before the read times out.
2. **Unbounded `casts` lists in bcast validators** (frostp2p.rs:1080) — `validate_round1_casts` / `validate_round2_casts` neither cap `msg.casts.len()` nor enforce val_idx uniqueness. A peer can send hundreds of thousands of valid-looking casts within `MAX_MESSAGE_SIZE`, each forcing a G1 decompression in `make_round1_response`. Mirror the strict checks already in `validate_round1_p2p`.

Minor findings cover: bcast `check` callback being a no-op (so peers sign before validation), dedup-then-validate ordering poisoning dedup slots on bad first messages, sequential direct-P2P sends, parity divergence in `validate_round1_p2p` (stricter than Charon), and small style/consistency items. See inline comments.

**Verdict:** REQUEST_CHANGES on the two `major` items. The minor/nit items are non-blocking but worth addressing where cheap.

crates/dkg/src/frostp2p.rs
```diff
@@ -0,0 +1,2141 @@
...
+    }
+
+    Ok(())
+}
+
+#[async_trait]
+impl FTransport for FrostP2P {
+    async fn round1(
+        &mut self,
+        cancellation: &CancellationToken,
```
mskrzypkows, line 1156: Can cancellation token be set a member of FrostP2P so it doesn't need to be passed here? In few other structs where it's propagated it potentially can be a member too.

crates/dkg/src/frostp2p.rs
iamquang95: This logic doesn't need order, the `op_id` is enough for correlate the result.

crates/dkg/src/frostp2p.rs
iamquang95: Expected

crates/dkg/src/frostp2p.rs
iamquang95: keep it

crates/dkg/src/frostp2p.rs
iamquang95: comment added

crates/dkg/src/frostp2p.rs
iamquang95: I feel passing the cancellation via argument makes it explicit to the caller that this function will cancel via that's token. Storing this in frostp2p makes it implicit. In the trade-off, it's a bit of boilerplate.

PR #368, opened 2026-05-04
feat: peers test

mskrzypkows:


mskrzypkows:
@claude review

crates/cli/src/commands/test/peers.rs
```diff
@@ -14,17 +90,820 @@ pub struct TestPeersArgs {
...
+            }
+            "DirectConn" => {
+                let r = TestResult::new("DirectConn");
+                if state.direct_connected {
+                    r.ok()
+                } else if !state.connected {
+                    r.fail(CliError::Other(
+                        "no relay connection established".to_string(),
+                    ))
+                } else if has_load_test && !state.direct_dial_attempted {
```
varex83agent: **Bug**: this branch is gated on `has_load_test`, but `has_load_test` has nothing to do with whether identify produced direct addresses for the peer. Running `--test-cases DirectConn` (without `PingLoad`) makes this branch unreachable, so a peer that never delivered a non-relay address via Identify falls through to the generic `direct connection not established within timeout` error rather than the more accurate `no direct addresses available from identify`. The condition should simply be `else if !state.direct_dial_attempted`.

crates/cli/src/commands/test/peers.rs
```diff
@@ -14,17 +90,820 @@ pub struct TestPeersArgs {
...
+            }
+            "PingMeasure" => {
+                let r = TestResult::new("PingMeasure");
+                if let Some(&(_, rtt)) = state.ping_rtts.first() {
+                    evaluate_rtt(rtt, r, THRESHOLD_MEASURE_AVG, THRESHOLD_MEASURE_POOR)
+                } else {
+                    r.fail(CliError::Other("no ping result received".to_string()))
+                }
+            }
+            "PingLoad" => {
```
varex83agent, line 929: **Bug / parity gap**: this is not a load test. Charon's `peerPingLoadTest` (testpeers.go:434-471) spawns a new `pingPeerContinuously` goroutine every second over `LoadTestDuration`, generating dozens of overlapping back-to-back ping streams. Pluto only filters previously observed `state.ping_rtts` (which come from libp2p's idle ping-keepalive at its default ~15s interval) by `time_since_connect < load_test_duration`. Over a 30s window this yields ~2 RTT samples per peer with zero applied load, so the verdict is effectively a duplicate of `PingMeasure` rather than reflecting under-load behaviour. Needs an active pinger driven over the load duration (e.g. issue ping requests on a 1s ticker, or run a dedicated request/response loop on the existing connection).

crates/cli/src/commands/test/peers.rs
```diff
@@ -14,17 +90,820 @@ pub struct TestPeersArgs {
...
+            peer_id,
+            info,
+            ..
+        })) => {
+            if needs_direct
+                && let Some(state) = states.get_mut(&peer_id)
+                && !state.direct_dial_attempted
+                && !state.direct_connected
+            {
+                state.direct_dial_attempted = true;
```
varex83agent: **Major parity gap**: `direct_dial_attempted` is set before `node.dial(...)` and never re-attempts. Charon's `peerDirectConnTest` (testpeers.go:481-488) loops `int(DirectConnectionTimeout.Seconds())` times calling `Connect(WithForceDirectDial(...))` with 1s sleeps — i.e. up to 120 retries by default. If the first dial fails (NAT not yet hole-punched, transient routing) Pluto silently waits the full `direct_connection_timeout` and reports failure where Charon would have succeeded. The 5s retry tick at line 658 doesn't help either because it only re-dials peers without `state.connected`, but here the peer is already relay-connected. Either retry on every Identify event, or schedule periodic direct-dial attempts until `direct_connected` or timeout. Consider also asserting that the relay connection still coexists with the direct one (Charon's `len(ConnsToPeer) >= 2` check).

crates/cli/src/commands/test/peers.rs
```diff
@@ -14,17 +90,820 @@ pub struct TestPeersArgs {
     /// Comma-separated list of each peer ENR address.
     #[arg(long = "enrs", value_delimiter = ',')]
     pub enrs: Option<Vec<String>>,
```
varex83agent, line 94: **Major parity gap**: Charon (testpeers.go:91-115, 120-128) supports three sources for ENRs — `--enrs`, `--lock-file`, **and `--definition-file`** — and enforces in a `PreRunE` wrapper that exactly one is specified. This struct omits `--definition-file` entirely, and `fetch_enrs` (peers.rs:352-364) silently prefers `--enrs` over `--lock-file` when both are given (no mutual-exclusion error), so a misconfigured invocation runs against the wrong peer set. Add `definition_file` to `TestPeersArgs`, mirror Charon's three-way switch in `fetch_enrs`, and reject conflicting combinations up front.

crates/cli/src/commands/test/peers.rs
```diff
@@ -14,17 +90,820 @@ pub struct TestPeersArgs {
...
+
+        results.insert(key, target_results);
+    }
+
+    results
+}
+
+async fn relay_ping_test(url: &str) -> TestResult {
+    let result = TestResult::new("PingRelay");
+    let client = reqwest::Client::new();
```
varex83agent, line 478: **Major**: `relay_ping_test` and `relay_ping_measure_test` build `reqwest::Client` with no `timeout`/`connect_timeout`, and `ct` is only inspected before the call (line 424), not during. A slow / hung relay (the defaults point at the public `*.relay.obol.tech` URLs) will stall `client.get(url).send().await` indefinitely; Ctrl-C / the parent CancellationToken cannot interrupt the in-flight request. Combined with sequential execution (line 419) and the relay group running before peer/self tests start (line 216), one stalled relay pegs the entire suite. Use `Client::builder().timeout(...).build()` (a single shared client) and wrap awaits in `tokio::select!` against `ct.cancelled()`. Charon's `relayPingTest` (testpeers.go:672-695) uses `http.NewRequestWithContext` so cancellation propagates into the HTTP layer.

crates/cli/src/commands/test/peers.rs
```diff
@@ -14,17 +90,820 @@ pub struct TestPeersArgs {
...
+        args.p2p_relays.clone()
+    };
+
+    // Relay HTTP tests run independently of the libp2p node.
+    let relay_results = run_relay_http_tests(&relay_urls, &relay_tests, ct.child_token()).await;
+
+    // Build ENR hash (sorted all-ENRs including self) for relay routing.
+    let self_enr = Record::from_key(&private_key)?;
+    let mut all_enrs = enr_strings.clone();
+    if !all_enrs.contains(&self_enr.to_string()) {
```
varex83agent: **Major**: `if !all_enrs.contains(&self_enr.to_string())` compares the freshly-encoded self ENR (canonical no-padding base64url from `Record::Display`) against ENR strings as they appear in the lock file. ENR strings in older Charon lock files are base64-padded with `=` (the parser explicitly strips this in `eth2util::enr` line ~214). When the lock-file copy of *this peer's own* ENR is in the legacy padded form, `contains` returns false and the self ENR is appended a second time. After sorting and joining, the SHA-256 over `all_enrs.join(",")` differs from what other cluster peers compute, and `enr_hash` is what `new_relays` uses to derive the relay routing room — so peers can silently land in different rooms and never see each other through relay routing. Normalize by round-tripping each entry through `Record::try_from(...)?.to_string()` before sorting, or hash the set of derived peer IDs instead of raw ENR strings.

crates/cli/src/commands/test/peers.rs
```diff
@@ -14,17 +90,820 @@ pub struct TestPeersArgs {
...
+        .map_err(|e| CliError::Other(format!("failed to derive peer ID: {e}")))?;
+
+    let relay_urls: Vec<String> = if args.p2p_relays.is_empty() {
+        DEFAULT_RELAY_URLS.iter().map(|s| s.to_string()).collect()
+    } else {
+        args.p2p_relays.clone()
+    };
+
+    // Relay HTTP tests run independently of the libp2p node.
+    let relay_results = run_relay_http_tests(&relay_urls, &relay_tests, ct.child_token()).await;
```
varex83agent: **Major**: Charon (testpeers.go:189-197) launches `testAllRelays`, `testAllPeers`, and `testSelf` as three concurrent errgroup goroutines, all sharing one timeout context. Pluto awaits `run_relay_http_tests` first, then starts the libp2p node and runs peer + self via `tokio::join!`. Per-relay tests are also serial inside `run_relay_http_tests` (line 419). With a slow relay this serially delays peer test startup and skews `execution_time`. Move the relay group into the same `tokio::join!` (or spawn it concurrently) and parallelise the per-relay loop with `FuturesUnordered` to match Charon's wall-clock semantics.

crates/cli/src/commands/test/peers.rs
```diff
@@ -14,17 +90,820 @@ pub struct TestPeersArgs {
...
+            "no --p2p-tcp-address configured".to_string(),
+        ));
+    }
+
+    // rust-libp2p multistream-select V1: the listener waits for the dialer to send
+    // the header first, then echoes it back. Wire format: varint(len) + message,
+    // so "/multistream/1.0.0\n" (19 bytes) is sent as 0x13 + 19 bytes = 20 bytes.
+    const MULTISTREAM_HEADER: &[u8] = b"\x13/multistream/1.0.0\n";
+
+    for addr in addrs {
```
varex83agent: **Major**: this `for addr in addrs` loop probes each TCP address sequentially, with up to 5 retries x 200ms read-timeout each (~1s/addr worst case). Charon's `libp2pTCPPortOpenTest` (testpeers.go:561-578) probes each address as a parallel errgroup goroutine. Combined with `run_self_tests` itself looping sequentially (line 476), self-tests can be Nx slower than they need to be on multi-address configs. Use `futures::future::join_all` over the addresses, and consider running each self-test case concurrently as well.

crates/cli/src/commands/test/peers.rs
```diff
@@ -1,9 +1,85 @@
...
+}
+
+const THRESHOLD_MEASURE_AVG: Duration = Duration::from_millis(50);
+const THRESHOLD_MEASURE_POOR: Duration = Duration::from_millis(240);
+const THRESHOLD_LOAD_AVG: Duration = Duration::from_millis(50);
+const THRESHOLD_LOAD_POOR: Duration = Duration::from_millis(240);
+const THRESHOLD_RELAY_MEASURE_AVG: Duration = Duration::from_millis(50);
+const THRESHOLD_RELAY_MEASURE_POOR: Duration = Duration::from_millis(240);
+
+const DEFAULT_RELAY_URLS: &[&str] = &[
```
varex83agent: Hard-coded `DEFAULT_RELAY_URLS` baked into `peers.rs` rather than being declared on the clap arg means `pluto alpha test peers --help` won't show what the actual relay set is. Charon binds the default in `bindP2PFlags` so it appears in `--help`. Also worth sanity-checking the canonical Obol relay URLs (`0.relay.obol.tech`, `1.relay.obol.tech`, `2.relay.obol.dev` — the inconsistent `.tech` vs `.dev` matches Charon's list, but verify). If the fallback default is intentional Pluto convention, declare it on the clap arg instead so it's discoverable; otherwise drop it and require the flag like Charon does.

crates/cli/src/commands/test/peers.rs
```diff
@@ -14,17 +90,820 @@ pub struct TestPeersArgs {
...
+        .map_err(|e| CliError::Other(format!("failed to load private key: {e}")))?;
+
+    verify_p2p_key(&cluster_peers, &private_key).map_err(|e| {
+        CliError::Other(format!("private key does not match any cluster peer: {e}"))
+    })?;
+
+    let self_peer_id = peer_id_from_key(private_key.public_key())
+        .map_err(|e| CliError::Other(format!("failed to derive peer ID: {e}")))?;
+
+    let relay_urls: Vec<String> = if args.p2p_relays.is_empty() {
```
varex83agent: When `(peer_tests || self_tests)` is non-empty but `args.p2p_tcp_addrs` is empty, the libp2p node won't bind a listener, the relay reservation address never advertises, `dialed_via_relay` stays false forever, and no peer dials are ever issued. The event loop spins until the test-config timeout (default 1h) and every test fails with a generic timeout. The `Libp2pTCPPortOpen` self-test also fails with the more helpful `no --p2p-tcp-address configured` message. Validate up front that `p2p_tcp_addrs` is non-empty when peer or self tests are queued, or at minimum log a `tracing::warn!`.

crates/cli/src/commands/test/peers.rs
```diff
@@ -14,17 +90,820 @@ pub struct TestPeersArgs {
...
+    node: &mut Node<TestBehaviour>,
+) {
+    for (peer, _) in target_peers {
+        for relay_peer in relay_peers.iter().filter_map(|r| r.peer().ok().flatten()) {
+            for relay_addr in &relay_peer.addresses {
+                let mut circuit_addr = relay_addr.clone();
+                circuit_addr.push(Protocol::P2p(relay_peer.id));
+                circuit_addr.push(Protocol::P2pCircuit);
+                circuit_addr.push(Protocol::P2p(peer.id));
+                let _ = node.dial(circuit_addr);
```
varex83agent: Both this dial (and the direct-dial branch at line 782) discard the immediate result of `node.dial()`. A failure here (malformed multiaddr, transport not supported) gives the user no diagnostic and never reaches `OutgoingConnectionError` because the dial never made it to the swarm. The only feedback then is the generic `TimeoutInterrupted` test result. Log at debug or warn level on dial-attempt failure: `if let Err(e) = node.dial(addr) { tracing::debug!(?e, %addr, "dial attempt failed"); }`.

crates/cli/src/commands/test/peers.rs
```diff
@@ -14,17 +90,820 @@ pub struct TestPeersArgs {
...
+                let r = TestResult::new("Ping");
+                if state.connected {
+                    r.ok()
+                } else if let Some(ref err) = state.connection_error {
+                    r.fail(CliError::Other(err.clone()))
+                } else {
+                    r.fail(CliError::TimeoutInterrupted)
+                }
+            }
+            "PingMeasure" => {
```
varex83agent, line 918: Charon's `peerPingMeasureTest` (testpeers.go:417-432) actively issues exactly one ping via `pingPeerOnce` and measures its RTT. Pluto reads `state.ping_rtts.first()`, which is whatever the libp2p ping-keepalive happened to record first — typically the scheduled keep-alive ~15s after the connection was established. Semantically different (cold vs. warm path) and Pluto can report a stale measurement. Issue a fresh ping on demand instead of relying on background keepalive samples.

crates/cli/src/commands/test/peers.rs
```diff
@@ -14,17 +90,820 @@ pub struct TestPeersArgs {
...
+        publish_result_to_obol_api(
+            all,
+            &args.test_config.publish_addr,
+            &args.test_config.publish_private_key_file,
+        )
+        .await?;
+    }
+
+    tracing::info!("Keeping TCP node alive until keep-alive time is reached...");
+    #[allow(clippy::arithmetic_side_effects)]
```
varex83agent: `#[allow(clippy::arithmetic_side_effects)]` masks a real overflow: `tokio::time::Instant::now() + Duration::MAX` panics, and `humantime::parse_duration` will accept arbitrarily large inputs. The pattern at lines 621-623 already does the right thing: `Instant::now().checked_add(d).unwrap_or_else(Instant::now)`. Use the same here (or extract a helper, since this `checked_add(...).unwrap_or_else(Instant::now)` shape recurs three times in this file) and drop the `#[allow]`.

crates/cli/src/commands/test/peers.rs
```diff
@@ -14,17 +90,820 @@ pub struct TestPeersArgs {
...
+            &cluster_peers,
+            self_peer_id,
+            &peer_tests,
+            &relay_peers,
+            &enr_strings,
+            &args,
+            cancel.clone(),
+        ),
+        async move {
+            let result = self_task.await.unwrap_or_default();
```
varex83agent: If the spawned `self_task` panics, `JoinError` is silently converted to `HashMap::new()` and the user sees no failures, no log line. Either log the join error (`Err(e) => { tracing::error!(?e, "self test task panicked"); HashMap::new() }`) or, better, hoist the future into the `tokio::join!` directly without `tokio::spawn` so panics surface naturally and the `args.p2p_tcp_addrs` / `self_tests_clone` clones at lines 263-264 disappear.

crates/cli/src/commands/test/peers.rs
```diff
@@ -14,17 +90,820 @@ pub struct TestPeersArgs {
     /// Comma-separated list of each peer ENR address.
     #[arg(long = "enrs", value_delimiter = ',')]
     pub enrs: Option<Vec<String>>,
-    // TODO: Add remaining flags from Go implementation
+
+    /// Path to cluster lock file.
+    #[arg(long = "lock-file", default_value = "")]
+    pub lock_file: String,
```
varex83agent: `lock_file: String` with `default_value = ""` and `is_empty()` checks at every use site (lines 165, 358) is the empty-string-as-null antipattern. The neighbouring `private_key_file` already uses `PathBuf`, so the type for a path argument is also inconsistent within the same struct. Prefer `pub lock_file: Option<PathBuf>` with no `default_value`. The validation at line 165 then reads `args.lock_file.is_none()`, and `fetch_enrs_from_lock` can take `impl AsRef<Path>` like every other path-using helper.

crates/cli/src/commands/test/peers.rs
```diff
@@ -14,17 +90,820 @@ pub struct TestPeersArgs {
...
-    // - Relay tests
-    unimplemented!("peers test not yet implemented")
+pub async fn run(
+    args: TestPeersArgs,
+    writer: &mut dyn Write,
+    ct: CancellationToken,
+) -> Result<TestCategoryResult> {
+    let enrs_empty = args.enrs.as_ref().is_none_or(Vec::is_empty);
+    if enrs_empty && args.lock_file.is_empty() {
+        return Err(CliError::Other(
```
varex83agent, line 204: `CliError::Other(format!(...))` recurs ~12 times in this file (lines 166, 199, 202, 207, 230, 261, 268, 274, 297, 326, 449, 480, ...). The crate's `CliError` already has typed variants and `#[from]` impls for `RecordError`, `K1Error`, `std::io::Error`, etc. Two patterns to fix: (1) `.map_err(|e| CliError::Other(format!("failed to ...: {e}")))` should become `?` (typed error propagation); (2) `r.fail(CliError::Other("...".to_string()))` inside `build_peer_results` and the self-test should use `r.fail(TestResultError::from_string("..."))` directly (avoids the `String -> CliError -> Display -> TestResultError` round-trip), or get a typed `PeersTestError` enum.

crates/cli/src/commands/test/peers.rs
```diff
@@ -14,17 +90,820 @@ pub struct TestPeersArgs {
...
+    if !args.lock_file.is_empty() {
+        return fetch_enrs_from_lock(&args.lock_file);
+    }
+    Err(CliError::Other(
+        "--enrs or --lock-file must be specified".to_string(),
+    ))
+}
+
+fn fetch_enrs_from_lock(path: &str) -> Result<Vec<String>> {
+    let content = std::fs::read_to_string(path)?;
```
varex83agent: `std::fs::read_to_string(path)` followed by `serde_json::from_str::<Lock>` has no size cap and no recursion-depth guard; a multi-GB or pathologically nested JSON file passed via `--lock-file` will OOM or stack-overflow before any other validation. Threat model is limited (the user picks the path themselves) but operators routinely receive lock files from a coordinator. Either cap the read at a sane upper bound (real lock files are tens of KB), or use `serde_json::from_reader` over a `take(N)`-bounded reader. While here, consider invoking `Lock::verify_signatures` before extracting ENRs.

crates/cli/src/commands/test/peers.rs
```diff
@@ -14,17 +90,820 @@ pub struct TestPeersArgs {
...
+            _ => TestResult::new(test.name)
+                .fail(CliError::Other("unsupported self test".to_string())),
+        };
+        results.push(result);
+    }
+
+    HashMap::from([("self".to_string(), results)])
+}
+
+async fn libp2p_tcp_port_open_test(addrs: &[String]) -> TestResult {
```
varex83agent, line 522: `libp2p_tcp_port_open_test` is ~85 lines, of which the inner `'retry: { for attempt in 0..5 { ... } }` block is a labelled-block escape used to bind `err: Option<CliError>`. This is the lack-of-`break-with-value` workaround. Extract a helper:

```rust
async fn probe_libp2p_addr(addr: &str) -> Result<(), CliError> { /* the inner loop */ }
```

Then the outer function becomes a flat `for addr in addrs { probe_libp2p_addr(addr).await.map(...)?; }`. Nesting drops from 6 levels to 3 and the helper becomes unit-testable independently with a fake TCP listener.

crates/cli/src/commands/test/peers.rs
```diff
@@ -14,17 +90,820 @@ pub struct TestPeersArgs {
...
+    /// All ping RTTs with the time they were observed.
+    ping_rtts: Vec<(Instant, Duration)>,
+    /// Whether a direct (non-relay) connection is established.
+    direct_connected: bool,
+    /// Whether we have already attempted a direct dial using identify
+    /// addresses.
+    direct_dial_attempted: bool,
+}
+
+#[allow(clippy::too_many_arguments)]
```
varex83agent, line 633: Eight parameters with `#[allow(clippy::too_many_arguments)]` is the canonical signal to introduce a struct. `PeerEventLoop<'a>` carrying `node`, `cluster_peers`, `self_peer_id`, `queued_tests`, `relay_peers`, `enr_strings`, `args`, `ct` lets `dial_peers_via_relay`, `handle_swarm_event`, `all_peers_done`, `peer_is_done`, and `build_peer_results` become `&self` / `&mut self` methods sharing the `states` map without threading it explicitly. The `#[allow]` disappears and reviewer eyes don't have to reconcile parameter order at every call site.

crates/cli/src/commands/test/peers.rs
```diff
@@ -14,17 +90,820 @@ pub struct TestPeersArgs {
...
+        .enumerate()
+        .map(|(i, s)| {
+            let record = Record::try_from(s.as_str())?;
+            Peer::from_enr(&record, i)
+                .map_err(|e| CliError::Other(format!("failed to parse peer from ENR: {e}")))
+        })
+        .collect()
+}
+
+fn format_enr(enr: &str) -> String {
```
varex83agent, line 415: nit: `&enr[..13]` and `&enr[enr.len().saturating_sub(4)..]` would panic on multi-byte UTF-8 boundaries. Today the helper is only reached via `peer_target_name` on strings that already passed `Record::try_from`, which enforces an `enr:` prefix and base64url body — guaranteed ASCII — so the slicing is sound in this PR. Flagging only because the function is `pub(super)`-visible and a future caller logging an unparsed ENR (e.g. inside a parse error path) would re-introduce a panic. Use `enr.char_indices()` to find safe boundaries, or operate on `enr.as_bytes()`.

crates/cli/src/commands/test/peers.rs
```diff
@@ -14,17 +90,820 @@ pub struct TestPeersArgs {
...
+    let p2p_cfg = P2PConfig {
+        relays: vec![],
+        external_ip: None,
+        external_host: None,
+        tcp_addrs: args.p2p_tcp_addrs.clone(),
+        udp_addrs: vec![],
+        disable_reuse_port: false,
+    };
+
+    let relay_peers_for_reservation = relay_peers.clone();
```
varex83agent: nit: `let relay_peers_for_reservation = relay_peers.clone();` exists solely so the closure passed to `Node::new` can move it. The closure is `FnOnce`, so just inline `relay_peers: relay_peers.clone()` inside the closure body — no separate binding and verbose name needed. The current shape suggests a `Fn`/`FnMut` constraint that doesn't exist.

varex83agent (dismissed):
This PR turns the previously stubbed `peers` test category into a real implementation (~900 LOC). The overall shape of the suite (Ping / PingMeasure / PingLoad / DirectConn / Libp2pTCPPortOpen / Relay HTTP tests) maps to Charon's `testpeers.go`, and several non-trivial pieces (the relay reservation behaviour, ENR-hash relay routing, multistream self-probe) are well thought through. Self-tested as the commit messages note.

That said, there are two concrete bugs and a handful of meaningful parity / robustness gaps that should be addressed before merge:

## Bugs
1. `build_peer_results` for `DirectConn` (peers.rs:894) gates the "no direct addresses available from identify" diagnostic on `has_load_test`, which has nothing to do with whether identify produced direct addresses. Running `--test-cases DirectConn` alone makes that branch unreachable and yields a misleading error.
2. `PingLoad` (peers.rs:866) is not actually a load test. Charon spawns a fresh continuous-ping goroutine every second over `LoadTestDuration`; Pluto only filters the libp2p ping-keepalive samples that happened to land in the window. Over the default 30s window that's ~2 RTT samples per peer with no real load applied — verdicts effectively duplicate `PingMeasure`.

## Major parity / robustness issues
- DirectConn attempts a single dial after the first Identify event and never retries; Charon retries up to ~120x over `DirectConnectionTimeout` (peers.rs:777).
- Missing `--definition-file` flag and no mutual-exclusion validation between `--enrs` / `--lock-file` (peers.rs:86, 165).
- Relay HTTP tests run sequentially before peer/self tests, with no per-request timeout and no in-flight cancellation; one stalled relay blocks the whole suite (peers.rs:216, 446).
- Self-ENR dedup uses raw string comparison against the lock file's ENR strings; canonical-vs-padded base64url encodings can produce different `enr_hash` values across cluster peers, putting them in different relay routing rooms (peers.rs:221).
- `Libp2pTCPPortOpen` probes addresses sequentially (and self-tests run sequentially); Charon parallelises both via `errgroup` (peers.rs:478, 504).
- **Zero unit tests** for ~900 lines of new logic; Charon's `testpeers_internal_test.go` is ~576 lines. Several pure functions (`peer_is_done`, `build_peer_results`, `format_enr`, `fetch_enrs`) are easy to cover and would have caught the bug at line 894.

Minor / nit findings cover error typing, `Option<PathBuf>` instead of empty-string sentinel, overflow-prone `Instant + Duration`, swallowed `JoinError`, hard-coded relay defaults, and a few code-organisation suggestions.

Verdict: requesting changes for the two bugs and the missing-flag / no-cancellation / no-tests issues.

mskrzypkows:
Regarding ping load test: the 1s ping interval already generates load; PingLoad's `evaluate_highest_rtt` on 30 samples is semantically different from PingMeasure's `first()`. Fixed the filter to use saturating arithmetic.

crates/cli/src/commands/test/peers.rs
mskrzypkows: There is a global timeout for tests, that's sufficient.

crates/cli/src/commands/test/peers.rs
mskrzypkows: code modified

crates/cli/src/commands/test/peers.rs
mskrzypkows: Pluto sets `DEFAULT_PING_INTERVAL = 1s` so it collects the same number of samples as 
Charon. We cannot force some pinging for Pluto as the libp2p architecture is different and we need to rely on events.

crates/cli/src/commands/test/peers.rs
```diff
@@ -1,30 +1,1375 @@
...
+
+    must_output_to_file_on_quiet(args.test_config.quiet, &args.test_config.output_json)?;
+
+    tracing::info!("Starting pluto peers and relays test");
+
+    let timeout_ct = ct.clone();
+    let timeout = args.test_config.timeout;
+    tokio::spawn(async move {
+        tokio::time::sleep(timeout).await;
+        timeout_ct.cancel();
```
iamquang95: This will cancel the `keep_node_alive(..., ct)` later as well, the keep-alive should stay longer (after test phase)

crates/cli/src/commands/test/peers.rs
```diff
@@ -1,30 +1,1375 @@
...
+                // an on-demand ping (pings are driven by the libp2p keepalive schedule),
+                // so .last() is the closest approximation to a fresh measurement.
+                if let Some(&(_, rtt)) = state.ping_rtts.last() {
+                    evaluate_rtt(rtt, r, THRESHOLD_MEASURE_AVG, THRESHOLD_MEASURE_POOR)
+                } else {
+                    r.fail(TestResultError::from_string("no ping result received"))
+                }
+            }
+            "PingLoad" => {
+                let r = TestResult::new("PingLoad");
```
iamquang95, line 932: That's one is hard with the rust libp2p implementation, but the current is not matched with charon PingLoad. Maybe it's worth to leave a comment here to state the gap between pluto and charon

crates/cli/src/commands/test/peers.rs
```diff
@@ -1,30 +1,1375 @@
...
+    res.execution_time = Some(CliDuration::new(elapsed));
+    res.score = Some(score);
+
+    write_and_publish_results(&res, writer, &args).await?;
+    keep_node_alive(&mut node, args.keep_alive, ct).await;
+
+    Ok(res)
+}
+
+fn run_self_tests_in_new_thread(
```
iamquang95: nit: it's tokio task instead of a thread

crates/cli/src/commands/test/peers.rs
```diff
@@ -1,30 +1,1375 @@
...
+    /// traffic. Empty default doesn't bind to local port therefore only
+    /// supports outgoing connections.
+    #[arg(long = "p2p-tcp-address", value_delimiter = ',')]
+    pub p2p_tcp_addrs: Vec<String>,
+
+    /// Comma-separated list of libp2p relay URLs or multiaddrs.
+    #[arg(
+        long = "p2p-relays",
+        value_delimiter = ',',
+        default_values = &[
```
iamquang95: use `pluto_p2p::config::DEFAULT_RELAYS`

crates/cli/src/commands/test/peers.rs
```diff
@@ -1,30 +1,1375 @@
...
+    let node: Node<TestBehaviour> = Node::new(
+        p2p_cfg,
+        private_key,
+        NodeType::TCP,
+        true,
+        p2p_context,
+        |builder, _keypair, relay_client| {
+            builder.with_gater(gater).with_inner(TestBehaviour {
+                relay: relay_client,
+                reservation: MutableRelayReservation::new(relay_peers_for_reservation),
```
iamquang95: inline `relay_peers.clone()` here

crates/cli/src/commands/test/peers.rs
```diff
@@ -1,30 +1,1375 @@
...
+    args: &TestPeersArgs,
+    ct: CancellationToken,
+) -> (HashMap<String, Vec<TestResult>>, Node<TestBehaviour>) {
+    let target_peers: Vec<(&Peer, &str)> = cluster_peers
+        .iter()
+        .zip(enr_strings.iter().map(String::as_str))
+        .filter(|(p, _)| p.id != self_peer_id)
+        .collect();
+
+    if queued_tests.is_empty() && target_peers.is_empty() {
```
iamquang95: `queued_tests.is_empty()` is enough to return

crates/cli/src/commands/test/peers.rs
```diff
@@ -1,30 +1,1375 @@
...
+
+fn run_self_tests_in_new_thread(
+    cancel: CancellationToken,
+    tcp_addrs: Vec<String>,
+    self_tests: Vec<TestCaseName>,
+    only_self_tests: bool,
+) -> JoinHandle<HashMap<String, Vec<TestResult>>> {
+    // Self tests run concurrently with peer tests; give the node a moment to bind.
+    tokio::spawn(async move {
+        tokio::time::sleep(Duration::from_millis(500)).await;
```
iamquang95: nit: make 500ms here as constant

crates/cli/src/commands/test/peers.rs
```diff
@@ -1,30 +1,1375 @@
...
+            }
+        })
+        .collect()
+}
+
+fn build_enr_hash(private_key: &k256::SecretKey, enr_strings: &[String]) -> Result<String> {
+    let self_enr = Record::from_key(private_key)?;
+    let mut all_enrs = enr_strings.to_vec();
+    if !all_enrs.contains(&self_enr.to_string()) {
+        all_enrs.push(self_enr.to_string());
```
iamquang95: nit: make `self_end.to_string` once

mskrzypkows:
@iamquang95  Resolved comments

crates/cli/src/commands/test/peers.rs
```diff
@@ -1,30 +1,1376 @@
...
+        args.test_config.test_cases.as_deref(),
+    );
+
+    if peer_tests.is_empty() && self_tests.is_empty() && relay_tests.is_empty() {
+        return Err(CliError::TestCaseNotSupported);
+    }
+
+    let enr_strings = fetch_enrs(&args).await?;
+    tracing::debug!("enr_strings: {:?}", enr_strings);
+    let cluster_peers = parse_peers(&enr_strings)?;
```
iamquang95, line 248: https://github.com/ObolNetwork/charon/blob/v1.7.1/cmd/testpeers.go#L820C2-L820C31

Add local peer here

crates/cli/src/commands/test/peers.rs
```diff
@@ -1,30 +1,1376 @@
...
+            return Err(CliError::from(e));
+        }
+    };
+
+    if let Err(e) = stream.write_all(header).await {
+        tracing::debug!(attempt, addr, err = %e, "write error");
+        return Err(CliError::from(e));
+    }
+
+    let mut buf = [0u8; 20];
```
iamquang95: 20 -> `MULTISTREAM_HEADER.len()`

iamquang95 (approved):
LGTM, the only concern is the `PingLoad` is differ from charon, which will provide different result.

crates/cli/src/commands/test/peers.rs
mskrzypkows: The relay doesn't work properly with it, local peer is already in enr_strings

varex83 (approved):

PR #373, opened 2026-05-05
feat: dkg

### LINKED ISSUE #251: Implement dkg/dkg
### Description
Implement dkg/dkg

### Acceptance criteria

varex83:


crates/dkg/src/frostp2p.rs
```diff
@@ -634,7 +634,7 @@ impl NetworkBehaviour for FrostP2PBehaviour {
                     return;
                 }
                 if !self.accepted_round1_p2p.insert(peer_id) {
-                    debug!(%peer_id, "ignoring duplicate round 1 p2p message");
+                    debug!(%peer_id, "Ignoring duplicate round 2 message");
```
varex83agent: **Bug — wrong round in log message.** This branch is the round-1 P2P handler: `validate_round1_p2p` is called above (line 626), the dedup set is `accepted_round1_p2p` (line 636), and adjacent error logs (lines 633, 644, 648) all say "round 1 p2p". The new wording "Ignoring duplicate round 2 message" was likely an editing mistake — should be `"Ignoring duplicate round 1 p2p message"`.

crates/dkg/src/node.rs
```diff
@@ -0,0 +1,210 @@
...
+                return true;
+            }
+
+            sig_types.contains(&duty.slot.inner())
+        }),
+    );
+    let (parsigex_comp, parsigex_handle) = parsigex::Behaviour::new(parsigex_config);
+
+    let (git_hash, _) = version::git_commit();
+    let peerinfo_config = peerinfo::Config::new(LocalPeerInfo::new(
```
varex83agent, line 104: **peerinfo missing peer list.** Charon's `setupP2P` calls `peerinfo.New(p2pNode, peerIDs, version, defHash, ...)` (`charon/dkg/dkg.go:1286`) so per-peer metrics (peerinfo index gauges, etc.) are seeded. Pluto constructs `Config::new(LocalPeerInfo::new(...))` with no `.with_peers(peer_ids.clone())`, so any operator dashboard keyed on peer index is missing data for DKG runs. Add `.with_peers(peer_ids.clone())` to the config builder.

crates/dkg/src/dkg.rs
```diff
@@ -112,6 +121,108 @@ pub enum DkgError {
...
+    UnsupportedDkgAlgorithm {
+        /// Algorithm name from the cluster definition.
+        algorithm: String,
+    },
+}
+
+impl From<frost::FrostError> for DkgError {
+    fn from(error: frost::FrostError) -> Self {
+        Self::Frost(error.to_string())
+    }
```
varex83agent: **Source chain lost.** `FrostError` already derives `thiserror::Error` with proper `#[source]` wiring, but the manual `From` impl stringifies it via `error.to_string()` so the underlying error (e.g. the `KryptologyError` behind `ExecRound1`) is dropped. Replace with `Frost(#[from] FrostError)` on the variant. The likely reason for the workaround is that `FrostError` is `pub(crate)` while `DkgError` is `pub` — fix is to expose `FrostError` (or wrap behind a public adapter) rather than flatten.

crates/dkg/src/dkg.rs
```diff
@@ -259,25 +374,620 @@ fn default_tracing_config() -> TracingConfig {
...
+        num_validators,
+        ct.child_token(),
+    )
+    .await?;
+
+    let node_idx = def.node_idx(node.local_peer_id()).map_err(|source| {
+        DkgError::LocalPeerNotInDefinition {
+            peer_id: match source {
+                DefinitionError::PeerNotFound { peer_id } => peer_id,
+                _ => *node.local_peer_id(),
```
varex83agent: **Discards non-`PeerNotFound` definition errors.** The `_ => *node.local_peer_id()` arm catches every other `DefinitionError` variant (version mismatch, malformed peer, etc.) and reports it as `LocalPeerNotInDefinition` keyed on the *local* peer id — misleading and loses the underlying cause. `DkgError::Definition` already exists with `#[from] DefinitionError`; map only `PeerNotFound` explicitly and propagate the rest:

```rust
let node_idx = def.node_idx(node.local_peer_id()).map_err(|source| match source {
    DefinitionError::PeerNotFound { peer_id } => DkgError::LocalPeerNotInDefinition { peer_id },
    other => DkgError::Definition(other),
})?;
```

crates/dkg/src/dkg.rs
```diff
@@ -259,25 +374,620 @@ fn default_tracing_config() -> TracingConfig {
...
+pub async fn run(conf: Config, ct: CancellationToken) -> Result<(), DkgError> {
+    if ct.is_cancelled() {
         return Err(DkgError::ShutdownRequestedBeforeStartup);
     }
 
+    let (lock_ct, lock_task) = start_private_key_lock(&conf).await?;
+    let result = run_inner(conf, ct).await;
+
+    lock_ct.cancel();
+    let _ = lock_task.await;
```
varex83agent: **JoinError swallowed.** If the private-key lock task (or its inner `run_task`) panics, `let _ = lock_task.await` discards it — DKG appears to complete cleanly while the mutex on the secp256k1 key file is gone. Same pattern at `dkg.rs:619` (`network_task`) and `dkg.rs:904` (sync runtime task drain). Either log on `Err`, or model these as `JoinHandle<Result<(), _>>` and propagate. `log_private_key_lock_result` (line 416) is a good pattern — extend it to the network/sync tasks.

crates/dkg/src/dkg.rs
```diff
@@ -259,25 +374,620 @@ fn default_tracing_config() -> TracingConfig {
...
+
+async fn drive_dkg_network(
+    mut node: pluto_p2p::p2p::Node<crate::node::DkgBehaviour>,
+    frost_handle: frostp2p::FrostP2PHandle,
+    cancellation: CancellationToken,
+) {
+    loop {
+        tokio::select! {
+            _ = cancellation.cancelled() => break,
+            event = node.select_next_some() => {
```
varex83agent, line 969: **Only `Bcast` events are forwarded.** All other `PlutoBehaviourEvent::Inner(...)` variants — parsigex, sync, peerinfo, frost-p2p, relay-router — fall through and are dropped without even a debug log. Each component does its own internal state work via its `poll`, so this is correctness-safe today, but `parsigex::Event::BroadcastError`/`Failed` and `sync::Event` carry diagnostic info Charon surfaces via its connection logger / parsigex internals. Add at least a `debug!` arm logging unmatched variants so operators can triage broadcast failures.

crates/dkg/src/dkg.rs
```diff
@@ -259,25 +374,620 @@ fn default_tracing_config() -> TracingConfig {
...
+    }
+
+    for deposit_data in &deposit_datas {
+        pluto_eth2util::deposit::write_deposit_data_file(deposit_data, &network, &conf.data_dir)
+            .await?;
+        debug!("Saved deposit data file(s) to disk");
+    }
+
+    // Signature verification and disk key write was step 6, advance to step 7.
+    sync_runtime.next_step().await?;
```
varex83agent, line 814: **Possible redundant peer round-trip at end of ceremony.** Step 7 is set + awaited (line 818) then `sync_runtime.shutdown().await` is called (line 820), which itself does `client.shutdown()` for each client and then `server.await_all_shutdown(...)`. In Charon's flow the step-7 `AwaitAllAtStep` is not paired with a separate `Shutdown`-wait — please cross-check with `charon/dkg/dkg.go` to confirm we're not blocking on slow peers twice at the very last step (an avoidable hang).

crates/dkg/src/dkg.rs
```diff
@@ -259,25 +374,620 @@ fn default_tracing_config() -> TracingConfig {
...
+            let _ = task.await;
+        }
+
+        Ok(())
+    }
+}
+
+impl Drop for SyncRuntime {
+    fn drop(&mut self) {
+        self.cancellation.cancel();
```
varex83agent, line 895: **`SyncRuntime::Drop` leaves spawned tasks behind without awaiting.** On any `?` between `start_sync_protocol` (line 647) and `shutdown` (line 820), `Drop` only calls `cancellation.cancel()` — the per-client `tokio::spawn`'d tasks (line 922) are detached. They exit eventually via cancellation, but anything they emit before noticing the token (errors logged from `drive_dkg_network` after `bcast::Component` channels close, etc.) is racy noise. Consider running the ceremony body inside a helper that always reaches `runtime.shutdown().await` (or wrap the JoinHandles in `JoinSet` / `AbortOnDrop`).

crates/dkg/src/dkg.rs
```diff
@@ -16,26 +20,31 @@ pub use crate::{
...
 use pluto_tracing::TracingConfig;
-use std::collections::HashMap;
 use url::Url;
 
 const DEFAULT_DATA_DIR: &str = ".charon";
 const DEFAULT_DEFINITION_FILE: &str = ".charon/cluster-definition.json";
 const DEFAULT_PUBLISH_ADDRESS: &str = "https://api.obol.tech/v1";
 const DEFAULT_PUBLISH_TIMEOUT: Duration = Duration::from_secs(30);
-const DEFAULT_SHUTDOWN_DELAY: Duration = Duration::from_secs(1);
+const DEFAULT_SHUTDOWN_DELAY: Duration = Duration::from_secs(3);
```
varex83agent: **`DEFAULT_SHUTDOWN_DELAY` diverges from Charon.** Charon's `bindShutdownDelayFlag` defaults to `time.Second` (`charon/cmd/dkg.go:77`); the previous Pluto value (1s) matched. Bumping to 3s without comment risks silent drift. Either revert to 1s, or add a comment explaining why Pluto needs longer (e.g. async drop ordering taking longer to flush libp2p).

crates/dkg/src/dkg.rs
```diff
@@ -259,25 +374,620 @@ fn default_tracing_config() -> TracingConfig {
...
+    !matches!(
+        version,
+        V1_0 | V1_1 | V1_2 | V1_3 | V1_4 | V1_5 | V1_6 | V1_7
+    )
+}
+
+struct SyncRuntime {
+    server: crate::sync::Server,
+    clients: Vec<crate::sync::Client>,
+    step: i64,
```
varex83agent, line 855: nit: `step: i64` permits negative values that have no semantic meaning here — `next_step` only ever increments from 0. Use `u64` (or `u32`) so the type expresses the invariant. `Client::set_step` (`sync/client.rs:103`) is also `i64` but internal — both can change together.

crates/dkg/src/dkg.rs
```diff
@@ -259,25 +374,620 @@ fn default_tracing_config() -> TracingConfig {
...
+            pluto_app::utils::bundle_output(data_dir, "dkg.tar.gz")
+        })
+        .await??;
+    }
+
+    debug!(
+        seconds = conf.shutdown_delay.as_secs(),
+        "Graceful shutdown delay"
+    );
+    tokio::time::sleep(conf.shutdown_delay).await;
```
varex83agent, line 830: nit: `tokio::time::sleep(conf.shutdown_delay).await` ignores the cancellation token. If the user hits Ctrl-C during the graceful-delay window the process keeps sleeping. Wrap in a `tokio::select!` against `ct.cancelled()` so shutdown is responsive.

crates/dkg/src/dkg.rs
```diff
@@ -259,25 +374,620 @@ fn default_tracing_config() -> TracingConfig {
...
+            pluto_eth2util::deposit::default_deposit_amounts(def.compounding)
+        } else {
+            vec![pluto_eth2util::deposit::DEFAULT_DEPOSIT_AMOUNT]
+        }
+    } else {
+        pluto_eth2util::deposit::dedup_amounts(&def.deposit_amounts)
+    }
+}
+
+fn support_partial_deposits(version: &str) -> bool {
```
varex83agent: Duplicates `pluto_cluster::definition::Definition::support_partial_deposits` (the existing helper exposes the same v1.0–v1.7 exclusion list). Two copies risk silent drift if Charon adds v1.11 with partial-deposit support. Either make the cluster-side helper public and reuse, or move it next to the other `pluto_cluster::version::support_*` helpers.

crates/dkg/src/dkg.rs
```diff
@@ -259,25 +374,620 @@ fn default_tracing_config() -> TracingConfig {
...
+    })?;
+
+    let peer_ids = def.peer_ids()?;
+    let exchanger = Exchanger::new(
+        ct.child_token(),
+        handlers.parsigex.clone(),
+        peer_ids,
+        sig_types,
+    )
+    .await;
```
varex83agent, line 561: `Exchanger::new` is constructed without a timeout. Charon's `newExchanger` takes `conf.Timeout` (`charon/dkg/dkg.go:246`) and plumbs it through to parsigex via `WithSendTimeout`/`WithReceiveTimeout`; `conf.timeout` is parsed in Pluto's `Config` but never consumed here. Either wire it through to `Exchanger`, or drop the `Config::timeout` field so users don't think it does something.

crates/cli/src/main.rs
```diff
@@ -32,12 +32,31 @@ async fn run() -> std::result::Result<(), CliError> {
...
-    // Top level cancellation token for graceful shutdown on Ctrl+C
+    // Top level cancellation token for graceful shutdown on Ctrl+C / SIGTERM.
     let ct = CancellationToken::new();
     tokio::spawn({
         let ct = ct.clone();
         async move {
-            let _ = tokio::signal::ctrl_c().await;
+            #[cfg(unix)]
+            {
+                use tokio::signal::unix::{SignalKind, signal};
```
varex83agent, line 42: If `signal(SignalKind::terminate())` returns `Err`, the fallback only awaits `ctrl_c`. systemd-managed deployments deliver SIGTERM by default — under that fallback path Pluto would die from the default SIGTERM action without releasing the privkey lock or finishing key writes. The failure is silent today; at minimum log a warning so operators can detect the degraded mode.

crates/dkg/src/node.rs
```diff
@@ -0,0 +1,210 @@
...
+        p2p_context.clone(),
+        &key,
+        def_hash.clone(),
+        version::VERSION.to_minor(),
+    )?;
+
+    let parsigex_config = parsigex::Config::new(
+        local_peer_id,
+        p2p_context.clone(),
+        Arc::new(|_duty, _pk, _sig| Box::pin(async { Ok(()) })),
```
varex83agent, line 88: nit: this no-op verifier mirrors Charon's `noopVerifier` (which is correct — partial-sig verification is deferred to aggregation in `aggregate.rs` since the roots aren't known to parsigex at receive time). Please add the rationale as an inline comment so future readers don't mistake it for an auth bypass; reference the `BlstImpl::verify` / `verify_threshold_partials` calls.

crates/dkg/src/dkg.rs
```diff
@@ -259,25 +374,620 @@ fn default_tracing_config() -> TracingConfig {
...
+    if !pluto_cluster::version::support_node_signatures(&lock.version) {
+        lock.node_signatures.clear();
+    }
+
+    // Node signatures was step 5, advance to step 6.
+    sync_runtime.next_step().await?;
+
+    if !conf.no_verify && append_config.is_none_or(|append| !append.unverified) {
+        lock.verify_signatures(eth1)
+            .await
```
varex83agent, line 756: `AppendConfig::unverified` is a public bool that disables both lock signing (existing-shares load is skipped at line 680, lock signing in `signing.rs`) and lock verification here at line 760. Mirrors Charon's `--unverified`, fine. Defense-in-depth: log an `info!`/`warn!` at the top of `run_inner` whenever it's set so the audit trail makes the unverified-mode entry obvious — `signing.rs` only logs partway through the ceremony.

crates/dkg/src/node.rs
```diff
@@ -0,0 +1,210 @@
...
+    pub(crate) parsigex: parsigex::Behaviour,
+    pub(crate) peerinfo: peerinfo::Behaviour,
+    pub(crate) frost_p2p: frostp2p::FrostP2PBehaviour,
+}
+
+type Result<T> = std::result::Result<T, DkgError>;
+
+pub(crate) struct Handlers {
+    pub(crate) bcast: bcast::Component,
+    pub(crate) sync: Vec<sync::Client>,
```
varex83agent, line 44: nit: `sync: Vec<sync::Client>` reads as a single handle. The caller renames it on assignment (`let sync_clients = handlers.sync.clone();` at `dkg.rs:590`), which is a strong signal — rename to `sync_clients` for symmetry with the adjacent `sync_server`.

crates/dkg/src/dkg.rs
```diff
@@ -259,25 +374,620 @@ fn default_tracing_config() -> TracingConfig {
...
+        handlers.parsigex.clone(),
+        peer_ids,
+        sig_types,
+    )
+    .await;
+
+    let peer_share_indices = peers
+        .iter()
+        .map(|peer| Ok((peer.id, u32::try_from(peer.share_idx())?)))
+        .collect::<Result<HashMap<_, _>, DkgError>>()?;
```
varex83agent, line 566: nit: `peer_share_indices` is built here, and the same `HashMap<PeerId, u32>` is built inside `setup_p2p` (`node.rs:113-121`). Two passes over `peers` for the same data. Have `setup_p2p` return the map (or expose it on `Handlers`) and reuse.

crates/dkg/src/dkg.rs
```diff
@@ -541,36 +1253,32 @@ mod tests {
...
         .await
-        .expect_err("backend handoff should panic until implemented");
+        .expect_err("mismatched P2P key should fail before networking");
 
-        assert!(join_err.is_panic());
+        assert!(matches!(
+            err,
+            DkgError::PeerError(pluto_p2p::peer::PeerError::UnknownPublicKey)
+        ));
     }
```
varex83agent, line 1271: **Test coverage gap.** This PR adds ~700 lines of orchestration (`run_inner`, `run_ceremony`, `start_sync_protocol`, `SyncRuntime`, `drive_dkg_network`, `deposit_amounts_for_definition`, `support_partial_deposits`, `private_key_lock_path`, `relay_addr_for_resolution`); the only new test is `run_reaches_p2p_key_verification_after_preflight` covering the mismatched P2P key path. Most of the new helpers are pure functions or small state machines — easy to unit-test. At minimum please add: `support_partial_deposits` table test (mirroring Charon's `TestSupportPartialDeposits`), `deposit_amounts_for_definition` table test, `relay_addr_for_resolution` table test, and a mainnet+TestConfig rejection test.

crates/p2p/src/proto.rs
```diff
@@ -57,8 +57,14 @@ pub async fn write_fixed_size_delimited<S: AsyncWrite + Unpin>(
     let len = i64::try_from(payload.len())
         .map_err(|_| io::Error::new(io::ErrorKind::InvalidData, "payload length overflow"))?;
 
-    stream.write_all(&len.to_le_bytes()).await?;
-    stream.write_all(payload).await?;
+    // Charon's `readSizedProto` uses a single `reader.Read(buf)` for the
+    // payload (not `io.ReadFull`), so it requires the length prefix and payload
+    // to arrive in one libp2p chunk. Coalesce them into one `write_all` to
```
varex83agent, line 62: nit: `8usize.saturating_add(payload.len())` is overcautious — `payload.len()` is already a valid `usize`, so `8 + payload.len()` cannot overflow on any 64-bit target without first having allocated more memory than exists. `let mut buf = Vec::with_capacity(8 + payload.len());` reads more directly.

varex83agent (dismissed):
Wires up the full DKG ceremony orchestration (P2P setup, sync protocol, FROST DKG, signing, lock verification, persistence, bundling) on top of the previously-stubbed entrypoint. Overall the structure follows Charon's `Run()` closely and the new `node::setup_p2p` is a clean extraction. A few items need addressing before merge:

**Bugs (must fix):**
1. `frostp2p.rs:637` — duplicate-message log says "round 2" but the surrounding code path is round-1 P2P (`validate_round1_p2p`, `accepted_round1_p2p`). Looks like a typo when editing the message.

**Major (must fix):**
1. `node.rs:104` — `peerinfo::Config::new(...)` is built without `.with_peers(peer_ids)`. Charon passes the peer set (`peerinfo.New(p2pNode, peerIDs, ...)` at `charon/dkg/dkg.go:1286`); without it, the per-peer metrics never publish.
2. `dkg.rs:225` — `From<FrostError>` collapses the source chain via `error.to_string()`. `FrostError` is `thiserror::Error` already; embed as `Frost(#[from] FrostError)` (make `FrostError` `pub` if needed).
3. `dkg.rs:553` — `node_idx` error mapping converts every non-`PeerNotFound` `DefinitionError` variant into `LocalPeerNotInDefinition` with the local peer id, hiding the real cause. Propagate via `DkgError::Definition(source)`.
4. `dkg.rs:387/619/904` — `let _ = ...await` on background tasks swallows `JoinError`. A panic in the lock service, network task, or sync client looks like a clean exit. At minimum, log on `Err`.
5. `dkg.rs:980` — `drive_dkg_network` matches only `Bcast` events; parsigex / sync / peerinfo / frost-p2p / relay-router events fall through silently. Behaviours self-manage internal state, but `parsigex::Event::BroadcastError` and similar are operationally important and Charon surfaces them — please log unmatched variants at debug.
6. `dkg.rs:818` — calling `sync_runtime.next_step().await?` for step 7 *and* `shutdown().await` causes a double round-trip with peers at the very end of the ceremony. Charon sends shutdown after step 7 directly. Verify and remove one.

**Minor / nit:** see inline comments — covers `DEFAULT_SHUTDOWN_DELAY` divergence, `unverified` audit logging, SIGTERM degraded-mode silent fallback, `conf.timeout` not plumbed, duplicated `peer_share_indices` computation, missing test coverage on the new pure functions, `support_partial_deposits` duplicating an existing helper, etc.

**Verdict:** Requesting changes for the bug + the source-chain / error-mapping / event-dispatch / double-shutdown items. The rest is non-blocking.

.github/workflows/dkg-runner.yml
iamquang95: can remove this test

crates/dkg/src/node.rs
```diff
@@ -0,0 +1,211 @@
...
+        bcast::Behaviour::new(peer_ids.clone(), p2p_context.clone(), key.clone());
+    let (sync_comp, sync_server, sync_clients) = sync::new(
+        peer_ids.clone(),
+        p2p_context.clone(),
+        &key,
+        def_hash.clone(),
+        version::VERSION.to_minor(),
+    )?;
+
+    let parsigex_config = parsigex::Config::new(
```
iamquang95, line 85: `with_timeout`

varex83:
@claude

varex83:
@claude review

crates/dkg/src/node.rs
varex83: Didn't get it, can you clarify?

iamquang95: On charon, they gave the `conf.Timeout` to exchanger, this should set timeout also
https://github.com/ObolNetwork/charon/blob/v1.7.1/dkg/dkg.go#L246

crates/dkg/src/dkg.rs
```diff
@@ -259,25 +368,615 @@ fn default_tracing_config() -> TracingConfig {
...
+        exchanger,
+        &mut frost_transport,
+        node_sig_caster,
+        sync_server,
+        sync_clients,
+    )
+    .await;
+
+    network_ct.cancel();
+    let _ = network_task.await;
```
iamquang95: don't swallow error here

crates/dkg/src/dkg.rs
```diff
@@ -259,25 +368,615 @@ fn default_tracing_config() -> TracingConfig {
...
+            client.shutdown(self.cancellation.child_token()).await?;
+        }
+
+        self.server
+            .await_all_shutdown(self.cancellation.child_token())
+            .await?;
+        self.cancellation.cancel();
+
+        for task in self.tasks.drain(..) {
+            let _ = task.await;
```
iamquang95: return error

crates/dkg/src/signing.rs
iamquang95, line 253: remove these

iamquang95 (approved):

PR #384, opened 2026-05-11
feat(bcast): add broadcast_and_wait for bcast

iamquang95:
This change makes DKG broadcast observable by callers that need completion.

Before, `bcast::Component::broadcast()` only enqueued work. Actual send success/failure arrived later as `bcast::Event`, but nodesigs and FROST flows had already moved on or were waiting elsewhere. That meant a failed broadcast could be dropped or ignored, causing DKG to hang until cancellation instead of failing with the real send error.

The fix adds `broadcast_and_wait()`: it enqueues the broadcast and waits for the behaviour’s terminal status. Nodesigs and FROST now use this API before continuing, so broadcast failures propagate immediately through their normal error paths. The old enqueue-only `broadcast()` remains for callers that consume events separately.

### DIFF
--- a/crates/dkg/src/nodesigs.rs
+++ b/crates/dkg/src/nodesigs.rs
@@ -99,8 +99,8 @@ impl NodeSigBcast {
 
     /// Exchanges K1 signatures over the lock hash with all peers.
     ///
-    /// Signs `lock_hash` with `key`, broadcasts the signature to all peers, and
-    /// polls until every peer's signature has been received and verified.
+    /// Signs `lock_hash` with `key`, waits for reliable-broadcast completion,
+    /// then polls until every peer's signature has been received and verified.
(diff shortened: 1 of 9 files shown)


varex83 (approved):

crates/dkg/src/nodesigs.rs
```diff
@@ -432,6 +435,53 @@ mod tests {
         assert_eq!(guard[1], Some(NONE_DATA.to_vec()));
     }
 
+    #[tokio::test]
+    async fn exchange_observes_bcast_failure() -> anyhow::Result<()> {
```
mskrzypkows:
```suggestion
    async fn exchange_observes_bcast_failure_on_peer_unreachable() -> anyhow::Result<()> {
```

mskrzypkows (approved):

PR #388, opened 2026-05-11
feat: add create cluster sem checks

varex83:


scripts/create-cluster-compare.sh
```diff
@@ -0,0 +1,395 @@
...
+    fi
+fi
+
+: "${WORK_DIR:=/tmp/create-cluster-compare}"
+: "${KEEP_WORK:=0}"
+
+FEE_RECIPIENT_ADDRESS="0xDeaDbeefdEAdbeefdEadbEEFdeadbeEFdEaDbeeF"
+WITHDRAWAL_ADDRESS="0xDeaDbeefdEAdbeefdEadbEEFdeadbeEFdEaDbeeF"
+
+EXACT=0
```
iamquang95: Can the `exact` work? since the seed for charon and pluto will be different.

scripts/create-cluster-compare.sh
```diff
@@ -0,0 +1,395 @@
...
+    PLUTO_BIN="${ROOT_DIR}/target/debug/pluto"
+fi
+
+if [[ -z "${CHARON_BIN:-}" ]]; then
+    if [[ -x "${ROOT_DIR}/../charon/charon" ]]; then
+        CHARON_BIN="${ROOT_DIR}/../charon/charon"
+    else
+        CHARON_BIN="charon"
+    fi
+fi
```
iamquang95, line 23: should expect to set the path, the `../charon` seems a bit off. Also the `charon_bin` should be required, otherwise we can run with different version of charon.

iamquang95 (approved):

emlautarom1:
@varex83 Could yo elaborate on what this PR achieves? It's not used on CI or any other place. When should we run this script? Should it be part of the test suite?

PR #392, opened 2026-05-11
refactor: relay router and mut relay reserver

varex83:


crates/p2p/src/relay.rs
```diff
@@ -48,468 +45,607 @@ const RELAY_BACKOFF_MAX: Duration = Duration::from_secs(120);
...
-                    &connected_relays,
-                    &peer,
-                );
-            }
+impl From<&DialError> for RelayDialError {
+    fn from(err: &DialError) -> Self {
+        match err {
+            DialError::LocalPeerId { .. } => Self::LocalPeerId,
+            DialError::NoAddresses => Self::NoAddresses,
+            DialError::DialPeerConditionFalse(_) => Self::Skipped,
```
mskrzypkows, line 176: Will it be correctly handled in the `on_dial_failure`? It'll retry instead of skipping.

crates/p2p/src/relay.rs
```diff
@@ -48,468 +45,607 @@ const RELAY_BACKOFF_MAX: Duration = Duration::from_secs(120);
...
-        let peers = {
-            let Ok(mut queue) = self.subscription_events.lock() else {
-                tracing::warn!("Failed to lock subscription events queue");
-                return;
-            };
-            queue.drain(..).collect::<Vec<_>>()
-        };
+    /// Builds circuit listen addresses for a relay from its transport
+    /// addresses: `/ip4/.../tcp/.../p2p/<relay-id>/p2p-circuit`.
+    fn circuit_addrs(relay_id: PeerId, addrs: &[Multiaddr]) -> Vec<Multiaddr> {
```
mskrzypkows, line 337: Would be good to have some unit test for this function

crates/p2p/src/relay.rs
```diff
@@ -48,468 +45,607 @@ const RELAY_BACKOFF_MAX: Duration = Duration::from_secs(120);
...
-                &mut self.pending_relays,
-                &mut self.pending_circuit_addrs,
-                &mut self.relay_peers,
-                &self.connected_relays,
-                &peer,
-            );
+    /// Extracts the relay peer id from a circuit listen address of the form
+    /// `/.../p2p/<relay-id>/p2p-circuit`. Returns `None` if the address is not
+    /// a relay circuit address.
+    fn relay_id_from_circuit_addr(addr: &Multiaddr) -> Option<PeerId> {
```
mskrzypkows, line 355: ditto

crates/p2p/src/relay.rs
```diff
@@ -48,468 +45,607 @@ const RELAY_BACKOFF_MAX: Duration = Duration::from_secs(120);
...
-        _connection_id: ConnectionId,
-        _peer: PeerId,
-        _local_addr: &Multiaddr,
-        _remote_addr: &Multiaddr,
-    ) -> Result<THandler<Self>, ConnectionDenied> {
-        Ok(dummy::ConnectionHandler)
+    /// Builds circuit dial addresses for reaching `target` through every
+    /// currently reserved relay:
+    /// `/.../p2p/<relay-id>/p2p-circuit/p2p/<target>`.
+    fn peer_circuit_addrs(&self, target: &PeerId) -> Vec<Multiaddr> {
```
mskrzypkows, line 510: ditto

crates/p2p/src/relay.rs
```diff
@@ -48,468 +45,607 @@ const RELAY_BACKOFF_MAX: Duration = Duration::from_secs(120);
...
+        let Some(state) = self.connection_states.get(&relay_id).copied() else {
+            return;
+        };
+        if matches!(state, RelayConnectionState::Reserved) {
+            tracing::info!(
+                relay_peer_id = %relay_id,
+                listen_addr = %addr,
+                "Relay circuit listener expired; demoting to Established"
+            );
+            self.set_relay_state(relay_id, RelayConnectionState::Established);
```
iamquang95, line 687: Do we need to update existing peer dial state?

crates/p2p/src/relay.rs
```diff
@@ -48,468 +45,607 @@ const RELAY_BACKOFF_MAX: Duration = Duration::from_secs(120);
...
-    fn poll_pending_redials(&mut self, cx: &mut Context<'_>) {
-        let retry_due = self
-            .next_retry
-            .as_mut()
-            .is_some_and(|sleep| sleep.as_mut().poll(cx).is_ready());
-        if !retry_due {
+    /// Applies a relay address update from a [`MutablePeer`]: refreshes
+    /// tracked addresses and, if this is the first time we've seen this
+    /// relay, kicks off a new dial campaign.
+    pub fn queue_relay_update(&mut self, relay: Peer) {
```
iamquang95: If a `MutablePeer` changes from one relay peer ID to another, the old relay can stay in the `Reserved` state and still be used for routing. Should we clear the old relay ID state when applying the update?

varex83: This is matches charon implementation, we can keep old relays

crates/p2p/src/relay.rs
varex83: Yes, because our reservation is expired -> libp2p's relay will make a reservation -> we will handle this event properly in relay manager

mskrzypkows (approved):
LGTM, will you add claude to the review?

crates/p2p/src/relay.rs
```diff
@@ -48,287 +45,630 @@ const RELAY_BACKOFF_MAX: Duration = Duration::from_secs(120);
...
+
+        if skipped {
+            tracing::debug!(
+                peer_id = %peer_id,
+                dial_type = ?target,
+                retry_count,
+                %error,
+                "Dial skipped (already connected or dialing); dropping dial state"
+            );
+            self.dial_states.remove(&peer_id);
```
varex83agent: **Bug**: when `DialError::DialPeerConditionFalse` fires for a `RelayDialType::Relay`, the dial state is removed but `connection_states[relay] = Dialing` is left untouched.

Trace: `redial_relay` sets `Dialing` and inserts a fresh `RelayDialState`. The dial fires; in a race with an inbound connection / a concurrent dial from another behaviour, the swarm rejects with `DialPeerConditionFalse`. We drop the dial state but leave `connection_states` at `Dialing`. From here:

- `queue_relay_update` short-circuits at line 354 (`connection_states.contains_key`) and refuses to re-arm.
- `process_relay_dials` has nothing to poll.
- `on_connection_established` only does work when `dial_states.remove(peer_id)` is `Some` (line 472), so even if a connection arrives we never transition to `Established` and never push the `ListenOn` events for the circuit listeners.

The relay is wedged in `Dialing` until process restart.

The peer-type branch is fine to drop the dial state (libp2p owns the existing connection and `on_connection_closed → reroute_peer` re-arms). For relays we need the same care — either also clear `connection_states[peer_id]` here (so the next watch update can recreate the dial campaign), or don't drop the dial state for `RelayDialType::Relay` and let the backoff retry until libp2p surfaces the real state.

crates/p2p/src/relay.rs
```diff
@@ -48,287 +45,630 @@ const RELAY_BACKOFF_MAX: Duration = Duration::from_secs(120);
...
+                    relay_peer_id = %relay_id,
+                    listen_addr = %addr,
+                    "Relay reservation confirmed; routing known peers via this relay"
+                );
+                self.set_relay_state(relay_id, RelayConnectionState::Reserved);
+                self.events
+                    .push_back(ToSwarm::GenerateEvent(RelayManagerEvent::RelayReserved(
+                        relay_id,
+                    )));
+                self.route_known_peers();
```
varex83agent, line 664: **Major**: peer-type `RelayDialState`s snapshot `peer_circuit_addrs` at the moment `route_known_peers` runs. If all currently-reserved relays subsequently demote (`on_expired_listen_addr`) or disconnect (`on_connection_closed`), nothing refreshes or evicts those peer dial states — they keep self-rearming and emitting `Dial` events at circuits through now-unreserved relays, producing a stream of spurious `DialFailed` events that eventually saturate at the 120s backoff cap.

`route_known_peers` only fires from this site (on `RelayReserved`), so between `RelayReservationLost` and the next `RelayReserved` the system burns retries on dead routes. Suggest: when the reserved-relay set shrinks (in `on_expired_listen_addr` / `on_connection_closed` for relays), iterate peer dial states and either rebuild them via `upsert_peer_dial` or drop the ones whose `peer_circuit_addrs` is now empty.

crates/p2p/src/relay.rs
```diff
@@ -48,287 +45,630 @@ const RELAY_BACKOFF_MAX: Duration = Duration::from_secs(120);
...
-                &self.connected_relays,
-                &peer,
+    }
+
+    /// Reacts to a circuit listen address expiring. If the relay was in
+    /// `Reserved`, demote it to `Established` so we stop routing peers through
+    /// it. libp2p's circuit-client will normally refresh the reservation and
+    /// emit `NewListenAddr` again, which promotes us back. If the transport
+    /// connection also drops, `on_connection_closed` will handle the redial.
+    fn on_expired_listen_addr(&mut self, addr: &Multiaddr) {
```
varex83agent, line 674: **Major (parity)**: the only path out of `Established` is `FromSwarm::NewListenAddr` for a circuit addr → promote to `Reserved`. We rely on libp2p's `relay::client::Behaviour` to re-issue RESERVE and emit `NewListenAddr` again after `ExpiredListenAddr` demotes us back to `Established`. If the relay denies the refresh (overloaded, bandwidth quota exhausted, version mismatch) libp2p's relay client typically gives up on that listener: no further `NewListenAddr` is ever emitted, no `RelayDialState` is armed, no fresh `ToSwarm::ListenOn` is issued, and `on_connection_closed` won't fire because the transport TCP is still up.

Charon's `NewRelayReserver` avoids this by explicitly calling `circuit.Reserve` on every loop iteration (`charon/p2p/relay.go:44`) and forcibly tearing down the connection when its 500ms `ConnsToPeer` check returns zero (`relay.go:73-92`). Consider either re-issuing `ToSwarm::ListenOn` after a grace period when stuck in `Established`, or treating consecutive `ExpiredListenAddr`s without an intervening `NewListenAddr` as a fatal reservation error and forcibly closing the relay's transport connection so `on_connection_closed → redial_relay` can recover.

crates/p2p/src/relay.rs
```diff
@@ -427,173 +745,217 @@ impl NetworkBehaviour for MutableRelayReservation {
...
-/// Polls relay peers periodically to detect address updates.
-pub struct RelayRouter {
-    relays: Vec<MutablePeer>,
-    p2p_context: P2PContext,
-    events: VecDeque<ToSwarm<Infallible, Infallible>>,
-    interval: Interval,
-    local_peer_id: PeerId,
-    connected_relays: HashMap<PeerId, Instant>,
-}
+#[cfg(test)]
```
varex83agent, line 913: **Major (test coverage)**: the test module covers only three pure helpers — `circuit_addrs`, `relay_id_from_circuit_addr`, `peer_circuit_addrs`. The most parity-sensitive code is now uncovered, and `backoff_delay` actually regressed (the prior file had `backoff_delay_grows_and_caps`).

Gaps to close:
1. `backoff_delay` — at minimum the retry=0 early-return (no jitter, exact base) and the cap-with-jitter behaviour at large `retry_count`. Inline a Charon-equivalent computation for cross-check.
2. State machine transitions — `on_connection_established → Established`; `on_new_listen_addr → Reserved`; `on_expired_listen_addr` demotes `Reserved → Established`; `on_connection_closed` re-arms relay vs peer differently.
3. `on_dial_failure` Skipped path — the orphaned-`Dialing` case flagged above is exactly the kind of thing a state-machine test would catch.
4. `queue_relay_update` — refresh of an in-flight dial's address list without resetting backoff; no-op when relay is `Established`/`Reserved`.
5. `upsert_peer_dial` — `addr_sets_equal` short-circuit (preserves backoff) vs. address-change reset (resets backoff) is policy worth locking in.
6. `MutablePeer` (peer.rs:130) — the watch-channel rewrite has zero tests. At minimum: `subscribe()` yields current value to a late subscriber; `set()` notifies live subscribers; default-constructed yields `None` until set.

The `RelayManager` is the backstop for cluster connectivity when direct dialing fails — for Charon parity this deserves the same coverage Charon has in `charon/p2p/relay_internal_test.go`.

crates/p2p/src/relay.rs
```diff
@@ -48,287 +45,630 @@ const RELAY_BACKOFF_MAX: Duration = Duration::from_secs(120);
 const RELAY_BACKOFF_JITTER: f64 = 0.2;
 
 /// Mutable relay reservation behaviour.
```
varex83agent: **Stale doc comments.** `/// Mutable relay reservation behaviour.` here understates the type after the refactor — it now does reservation *and* routing of known peers through reserved circuits. e.g. `/// Libp2p NetworkBehaviour that reserves circuits on a configured set of relays and routes known cluster peers through them.`

The unchanged module-level doc at the top of the file is also stale — it references three things that no longer exist as separate entities ("This behaviour…", "Mutable peer is used for…", "Relay router is responsible for…"). The `MutablePeer` ENR-fetching sentence belongs in `bootnode.rs`/`peer.rs`, not here — `RelayManager` only consumes the watch channel, it does not fetch ENRs. Worth updating in this PR since the merge invalidated it. Suggested rewrite: a single paragraph describing `RelayManager`'s three responsibilities — (1) subscribes to `MutablePeer` watch channels for relay address updates, (2) manages reservation lifecycle (`Dialing → Established → Reserved`) with exponential backoff redial, (3) routes known cluster peers through reserved circuits.

crates/p2p/src/relay.rs
```diff
@@ -48,287 +45,630 @@ const RELAY_BACKOFF_MAX: Duration = Duration::from_secs(120);
...
+pub enum RelayDialType {
+    /// Dial a known cluster peer via reserved relay circuits.
+    Peer,
+    /// Dial a relay server directly.
+    Relay,
+}
+
+/// State of an in-flight dial campaign, polled to produce a `ToSwarm::Dial`
+/// event each time its backoff elapses.
+pub struct RelayDialState {
```
varex83agent: `RelayDialState` and its fields `ty`/`peer_id`/`addrs`/`retry_count` are `pub` but are never constructed or read from outside this module. Exposing them leaks internal dial-campaign machinery into `pluto_p2p`'s public API and blocks future restructuring without a semver bump. Recommend dropping `pub` on the struct and the four fields (the `sleep` field is already private, which is the right default), and same for `RelayDialState::new`.

crates/p2p/src/relay.rs
```diff
@@ -48,287 +45,630 @@ const RELAY_BACKOFF_MAX: Duration = Duration::from_secs(120);
...
+                    .join("; "),
+            ),
         }
+    }
+}
+
+/// Whether a [`RelayDialState`] is targeting a relay server or a cluster peer
+/// reached through reserved relay circuits.
+#[derive(Debug, Clone, Copy)]
+pub enum RelayDialType {
```
varex83agent, line 194: `RelayDialType::Peer` / `RelayDialType::Relay` are ambiguous: every dial in this module targets a `PeerId`, so `Peer` doesn't distinguish anything at the use sites. `match dial_state.ty { RelayDialType::Peer => … }` reads as "if the dial target is a peer" which is always true. Prefer e.g. `Cluster` / `RelayServer`, or `Direct` / `ViaRelay`. The same ambiguity propagates into `RelayManagerEvent::DialFailed.target` (line 104) which is part of the public API.

crates/p2p/src/relay.rs
```diff
@@ -48,287 +45,630 @@ const RELAY_BACKOFF_MAX: Duration = Duration::from_secs(120);
...
+        None
+    }
 
-        tracing::debug!(
-            relay_peer_id = %peer.id,
-            "Queued relay dial, will listen on circuit after connection establishes"
+    /// Applies a relay address update from a [`MutablePeer`]: refreshes
+    /// tracked addresses and, if this is the first time we've seen this
+    /// relay, kicks off a new dial campaign.
+    pub fn queue_relay_update(&mut self, relay: Peer) {
```
varex83agent: `queue_relay_update` (here) and `process_relay_dials` (line 382) are `pub` but only called from `RelayManager::poll` in the same file. Making them `pub` advertises them as part of the public API and lets callers bypass the watch-stream subscription model. `process_relay_dials` also takes a `&mut Context<'_>`, which no external caller would have access to. Drop `pub` on both (or `pub(crate)` if a future internal caller needs them).

crates/p2p/src/relay.rs
```diff
@@ -48,287 +45,630 @@ const RELAY_BACKOFF_MAX: Duration = Duration::from_secs(120);
...
+    Dialing,
+    /// Transport connection up; reservation not yet confirmed.
+    Established,
+    /// Reservation confirmed; circuits through this relay are usable.
+    Reserved,
+}
 
-        if pending_relays.contains(&peer.id) || connected_relays.contains(&peer.id) {
-            return;
+impl Future for RelayDialState {
```
varex83agent: `impl Future for RelayDialState` returns `Poll::Ready` and then resets its `Sleep` so the *next* poll starts a new wait. Polling a `Future` after it has returned `Ready` is conventionally unspecified, so this only works because `process_relay_dials` blindly re-polls. The natural abstraction for "yields an item, then re-arms, then yields another" is `Stream`. Change to `impl Stream for RelayDialState { type Item = ToSwarm<…>; }`, then `state.poll_next_unpin(cx)` in `process_relay_dials` — that also lets you drop the manual `Pin::new(state)` dance.

crates/p2p/src/relay.rs
```diff
@@ -48,287 +45,630 @@ const RELAY_BACKOFF_MAX: Duration = Duration::from_secs(120);
...
-            // Poll once to register the waker before returning Pending.
-            let _ = sleep.as_mut().poll(cx);
-            self.next_retry = Some(sleep);
+    /// Reacts to the last connection to `peer_id` closing. Either it's one of
+    /// our relays (queue a fresh re-dial cycle) or a known cluster peer
+    /// (arm a fresh routing dial through the current reserved relays).
+    /// Anything else is ignored.
+    fn on_connection_closed(&mut self, peer_id: PeerId) {
+        if self.connection_states.contains_key(&peer_id) {
+            self.events.push_back(ToSwarm::GenerateEvent(
```
varex83agent, line 715: When a reserved relay's last connection drops, only `RelayDisconnected` is emitted — not `RelayReservationLost`. The `Reserved` state is implicitly lost (subsequent `redial_relay` flips us to `Dialing`) but a subscriber that uses `RelayReservationLost` as the signal "circuits through this relay are no longer routable" will miss this case. Either emit `RelayReservationLost` first in `on_connection_closed` when transitioning out of `Reserved`, or document on the enum that `RelayDisconnected` subsumes a prior reservation loss.

crates/p2p/src/relay.rs
```diff
@@ -48,287 +45,630 @@ const RELAY_BACKOFF_MAX: Duration = Duration::from_secs(120);
...
+                    relay_peer_id = %relay_id,
+                    listen_addr = %addr,
+                    "NewListenAddr for relay in Dialing state; ignoring"
+                );
+            }
+            RelayConnectionState::Reserved => {
+                // Second circuit address from the same relay — already routed.
+            }
+            RelayConnectionState::Established => {
+                tracing::info!(
```
varex83agent, line 654: `info!` fires on every `Established → Reserved` transition, including the routine refresh cycle every time libp2p's relay client renews the reservation (`Reserved → Established → Reserved` via `ExpiredListenAddr` then a fresh `NewListenAddr`). On a long-running node that produces one info-pair per relay per refresh interval. Consider `debug!` for the refresh case (e.g. when the previous state's most-recent value was `Reserved` recently) and keep `info!` for the first acceptance only. Same applies to the `Relay circuit listener expired` log at line 557.

crates/cli/src/commands/test/peers.rs
```diff
@@ -45,17 +45,18 @@ use crate::{
...
 struct TestBehaviour {
     relay: relay::client::Behaviour,
-    reservation: MutableRelayReservation,
+    relay_manager: RelayManager,
 }
 
 #[derive(Debug)]
 enum TestBehaviourEvent {
     Relay(relay::client::Event),
+    RelayManager(#[allow(dead_code)] pluto_p2p::relay::RelayManagerEvent),
```
varex83agent, line 59: `#[allow(dead_code)]` on the variant payload silences a warning by tagging the payload as dead, but nothing in the file ever matches on `TestBehaviourEvent::RelayManager` to read it. The cleaner fix is to drop the payload (`RelayManager` with no fields) and have `From<RelayManagerEvent> for TestBehaviourEvent` discard the event, or — better — actually consume it for a `tracing::debug!`. Same pattern repeats in `crates/parsigex/examples/parsigex.rs:102`.

scripts/dkg-runner/config.sh
```diff
@@ -11,7 +11,7 @@
 : "${CHARON_NODES:=2}"
 : "${RELAY_URL:=https://0.relay.obol.tech}"
 : "${TIMEOUT:=120}"
-: "${SHUTDOWN_DELAY:=30s}"
+: "${SHUTDOWN_DELAY:=5s}"
```
varex83agent, line 14: `SHUTDOWN_DELAY` default lowered from `30s` to `5s` — unrelated to the relay refactor and ships unflagged in the same commit set. A shorter graceful-shutdown window can mask resource-cleanup races (peerstore writes, in-flight DKG packets, log flush). If the change is intentional, please split into its own commit with rationale; otherwise revert.

varex83agent (dismissed):
## Summary
Merges the two prior relay behaviours (`MutableRelayReservation` + `RelayRouter`) into a unified `RelayManager` and rewrites `MutablePeer` from `Arc<Mutex<…>>` + callbacks to `tokio::sync::watch`. The new state machine (`Dialing → Established → Reserved`) is a cleaner mental model and the new `RelayManagerEvent` surface gives consumers visibility into the lifecycle. Overall direction is good and the existing helper tests are well written.

## Bugs / must-fix
1. `on_dial_failure` Skipped path drops the relay's `RelayDialState` but leaves `connection_states[relay] = Dialing` — the relay can wedge with no dial in flight (relay.rs:613).
2. After all reserved relays are lost, peer-type dial states keep firing `Dial` events with stale circuit addrs to now-unreserved relays; `route_known_peers` only fires on a fresh `RelayReserved` so there is no eviction path (relay.rs:449/539).
3. Test coverage regressed: `backoff_delay` (previously tested) and every state-machine transition (`on_connection_established`, `on_new_listen_addr` promotion, `on_expired_listen_addr` demotion, the Skipped dial-failure path, `on_connection_closed` for relays vs peers) are uncovered, as is the new `MutablePeer` watch-based API.

## Major
- No proactive recovery from `Established` when libp2p's relay client silently stops refreshing reservations — Charon's `NewRelayReserver` calls `circuit.Reserve` itself; the Rust port leans entirely on libp2p's relay client (relay.rs:549).

## Minor / nit
Stale module + `RelayManager` doc comments; over-broad `pub` on `RelayDialState`/its fields, `queue_relay_update`, and `process_relay_dials`; `RelayDialType::Peer` / `Relay` naming ambiguity; `Future` impl that self-rearms should be a `Stream`; `RelayDisconnected` semantics vs `RelayReservationLost`; `info!` on every reservation refresh; `#[allow(dead_code)]` payload anti-pattern in two examples; unrelated `SHUTDOWN_DELAY` drop from 30s→5s in `scripts/dkg-runner/config.sh`.

## Verdict
Requesting changes for the Skipped-dial state desync, the stranded peer dial states after relay loss, and the test-coverage regression. The architectural direction is sound — once these are addressed the refactor is a clear improvement over the old split.

iamquang95 (approved):

crates/p2p/src/relay.rs
varex83agent: Fixed in f7caf9e6 (Group A). Rewrote the module-level doc to describe `RelayManager`'s three responsibilities (watch subscriptions, reservation lifecycle, cluster-peer routing) and tightened the struct-level doc to match. Stale references to a separate `MutablePeer` ENR-fetching component and a standalone relay-router are gone.

crates/p2p/src/relay.rs
varex83agent: Fixed in f7caf9e6 (Group A). Renamed `RelayDialType::{Peer,Relay}` → `{ClusterPeer,RelayServer}`. Match arms now read unambiguously at call sites (`RelayDialType::RelayServer` is clearly the direct dial to a relay; `ClusterPeer` is the via-circuit dial). No external callers were affected — `RelayManagerEvent::DialFailed.target` is only consumed via `Debug` outside the crate.

crates/p2p/src/relay.rs
varex83agent: Fixed in f7caf9e6 (Group A). Dropped `pub` on `RelayDialState`, all four fields (`ty`/`peer_id`/`addrs`/`retry_count`), and `::new`. The struct and its constructor are now private to the module, matching the actual usage.

crates/p2p/src/relay.rs
varex83agent: Fixed in f7caf9e6 (Group A). Dropped `pub` on `queue_relay_update` and `process_relay_dials` — both now private. They're only called from `RelayManager::poll` in this file.

crates/p2p/src/relay.rs
varex83agent: Fixed in e956eddb (Group C). `RelayDialState` now implements `futures::Stream<Item = ToSwarm<...>>` instead of `Future`. `process_relay_dials` switched to `state.poll_next_unpin(cx)`, dropping the manual `Pin::new` dance. The repeating-yield semantics now match the trait contract.

crates/p2p/src/relay.rs
varex83agent: Fixed in 50082b47 (Group B). Added `refresh_peer_dials`, called from `on_expired_listen_addr` (after demoting a relay from Reserved → Established) and from `on_connection_closed` when the closing relay was previously Reserved. It walks every cluster-peer dial state and re-runs `upsert_peer_dial`; `upsert_peer_dial` now drops dial state for targets whose `peer_circuit_addrs` came back empty (no reserved relay can route to them). Spurious `DialFailed` storms through dead circuits are gone. Regression coverage in `on_expired_listen_addr_drops_peer_dials_with_no_route_left` and `upsert_peer_dial_drops_stale_state_when_no_route_left` (fa17a28b).

crates/p2p/src/relay.rs
varex83agent: Fixed in c60ec48d (Group D). Added a stuck-`Established` watchdog: `set_relay_state` records an `established_at` timestamp on Dialing→Established transitions (cleared on promotion to Reserved or demotion to Dialing). Every 15s (`ESTABLISHED_WATCHDOG_TICK`) the watchdog scans for relays that have been Established for ≥60s (`ESTABLISHED_STUCK_THRESHOLD`) and emits `ToSwarm::CloseConnection { connection: All }`. The resulting `FromSwarm::ConnectionClosed` drives `on_connection_closed` → `redial_relay` — mirroring Charon's `charon/p2p/relay.go:73-92` recovery path. The watchdog `Sleep` is lazily initialised on first `poll` so `RelayManager::new` still works outside a Tokio runtime (unit tests on pure helpers depend on this).

crates/p2p/src/relay.rs
varex83agent: Fixed in 50082b47 (Group B). `on_connection_closed` now reads the previous state before redialing; if it was `Reserved`, `RelayReservationLost(peer_id)` is pushed onto the event queue before `RelayDisconnected(peer_id)`. Subscribers using `RelayReservationLost` as the "circuits are unroutable" signal see it on relay disconnects too, not just on circuit-listener expiry. Covered by `on_connection_closed_reserved_relay_emits_lost_before_disconnected` (and its negative twin) in fa17a28b.

crates/p2p/src/relay.rs
varex83agent: Fixed in 50082b47 (Group B). `on_dial_failure` now splits on dial type for the Skipped path:

- `RelayDialType::ClusterPeer`: drop the dial state (current behaviour — libp2p owns the existing direct connection and `on_connection_closed → reroute_peer` re-arms).
- `RelayDialType::RelayServer`: **keep** the dial state armed. Dropping it would wedge `connection_states` in `Dialing` exactly as you described. Backoff retries are cheap (libp2p re-rejects with the same error) and `on_connection_established` will tear down the dial state once libp2p surfaces the real connection.

Regression coverage in `on_dial_failure_skipped_relay_keeps_dial_state` (asserts both that the dial state survives and that `connection_states` remains `Dialing`) and `on_dial_failure_skipped_cluster_peer_drops_dial_state` for the other branch (fa17a28b).

crates/p2p/src/relay.rs
varex83agent: Fixed in fa17a28b (Group E). Added 21 unit tests across `relay.rs` and `peer.rs` covering:

1. **`backoff_delay`**: `retry=0` returns base exactly (Charon early-return parity), jittered cap stays inside ±20% envelope at large retries, average growth then plateau.
2. **State machine**: `on_connection_established` (RelayServer → Established + ListenOn events; ClusterPeer drops state + PeerRoutedConnected); `on_new_listen_addr` (Established → Reserved + RelayReserved); `on_expired_listen_addr` (Reserved → Established + RelayReservationLost; also drops peer dials with no route left); `on_connection_closed` orders RelayReservationLost before RelayDisconnected when prev state was Reserved, skips it otherwise.
3. **`on_dial_failure` Skipped path**: ClusterPeer drops, RelayServer keeps (regression coverage for the wedge bug).
4. **`queue_relay_update`**: first-seen arms campaign; in-flight refresh swaps addrs without resetting backoff; no-op when relay already connected.
5. **`upsert_peer_dial`**: addr-set unchanged preserves dial state (and backoff); addr-set changed resets; empty addrs drops stale state.
6. **`MutablePeer`** (peer.rs): default yields `None` until set; late subscribers see current value immediately; `set()` notifies live subscribers.

Total: 96 tests pass (75 before, 21 new).

PR #416, opened 2026-05-14
feat(testutil): add beaconmock

### LINKED ISSUE #37: Implement testutil/beaconmock

varex83:


crates/testutil/src/beaconmock/fuzzer.rs
```diff
@@ -0,0 +1,469 @@
...
+    let mut rng = rand::thread_rng();
+    let count = rng.gen_range(0..=4u64);
+    let data: Vec<Value> = (0..count)
+        .map(|index| {
+            json!({
+                "index": index.to_string(),
+                "balance": rng.r#gen::<u64>().to_string(),
+                "status": random_validator_status(&mut rng),
+                "validator": {
+                    "pubkey": format!("0x{}", hex::encode([rng.r#gen::<u8>(); 48])),
```
varex83agent, line 183: **bug** — `[rng.r#gen::<u8>(); 48]` evaluates the random byte exactly once and copies it 48 times (Rust array-repeat `Copy` semantics). Every emitted pubkey is therefore one of only 256 distinct values (`0xaaaa…aa`, `0xbbbb…bb`, …), and validators generated in the same `(0..count).map(...)` chain typically share the same pubkey. This drastically reduces the value of the fuzzer for any code that branches on pubkey content (lookup by pubkey, hex parsing, BLS validation paths).

The same pattern is repeated in this file at lines 213, 242, and 264.

Fix:

```rust
let mut pk = [0u8; 48];
rng.fill_bytes(&mut pk);
// …
"pubkey": format!("0x{}", hex::encode(pk)),
```

Note the deliberate zero arrays at lines 331/334 (`[0u8; 20]`, `[0u8; 256]`) are *intentional* — only the random-pubkey sites need to change.

crates/testutil/src/beaconmock/headproducer.rs
```diff
@@ -0,0 +1,457 @@
...
+            slot_duration,
+        );
+
+        Self { shutdown }
+    }
+}
+
+impl Drop for HeadProducer {
+    fn drop(&mut self) {
+        self.shutdown.notify_waiters();
```
varex83agent: **major** — `Notify::notify_waiters()` only wakes tasks that have already started polling `notified()`. Between `HeadProducer::spawn` returning and the spawned task at line 140 actually being polled by the scheduler, there is a window where `Drop` can fire `notify_waiters()` and no waiter exists. The signal is then lost and the ticker runs indefinitely (one leaked tokio task per dropped `BeaconMock`).

This happens to be invisible today because `tokio::pin!(shutdown_fut)` is created up-front and the subsequent `Notified` future captures the notify generation at construction. But the invariant is fragile — moving the `pin!` after any `.await`, or freshly constructing `notified()` inside the loop, would silently reintroduce the race.

Recommended: switch to `tokio_util::sync::CancellationToken` (cancellation is persistent and lets you also `cancel_token.is_cancelled()` defensively in the body), or use `Notify::notify_one()` which stores a permit.

As a side benefit, capturing the `JoinHandle` from `tokio::spawn` and `.abort()`-ing it on drop adds belt-and-braces shutdown that doesn't depend on cooperative `select!`.

crates/testutil/src/beaconmock/state.rs
```diff
@@ -0,0 +1,272 @@
...
+            balance: index,
+            status: ValidatorStatus::ActiveOngoing,
+            validator: ValidatorResponseValidator {
+                activation_eligibility_epoch: index.to_string(),
+                activation_epoch: index.checked_add(1).unwrap_or(index).to_string(),
+                effective_balance: index.to_string(),
+                exit_epoch: u64::MAX.to_string(),
+                pubkey,
+                slashed: false,
+                withdrawable_epoch: u64::MAX.to_string(),
```
varex83agent: **major** — `exit_epoch` and `withdrawable_epoch` are set to `u64::MAX`, but Charon's `ValidatorSetA` (charon/testutil/beaconmock/options.go:103-140) leaves these fields zero — only `PublicKey`, `EffectiveBalance`, `ActivationEligibilityEpoch`, `ActivationEpoch`, and `WithdrawalCredentials` are populated. The `u64::MAX` value comes from the Go *fuzzer* (`beaconmock_fuzz.go:44-45`), not from `ValidatorSetA`.

Consequence: any test using `ValidatorSet::validator_set_a()` will see different validator JSON than Charon's mock, with `exit_epoch: "18446744073709551615"` instead of `"0"`. Either set these to 0 to match Charon, or document the intentional deviation.

crates/testutil/src/beaconmock/defaults.rs
```diff
@@ -0,0 +1,620 @@
...
+            "data": [
+                { "previous_version": "0x01017000", "current_version": "0x01017000", "epoch": "0" },
+                { "previous_version": "0x01017000", "current_version": "0x02017000", "epoch": "0" },
+                { "previous_version": "0x02017000", "current_version": "0x03017000", "epoch": "0" },
+                { "previous_version": "0x03017000", "current_version": "0x04017000", "epoch": "0" },
+                { "previous_version": "0x04017000", "current_version": "0x05017000", "epoch": "0" }
+            ]
+        })
+    })
+    .await;
```
varex83agent, line 53: **major** — This hand-rolled `fork_schedule` returns `epoch="0"` for every entry, while Charon serves the response embedded in `static.json` (which has Capella→Deneb at epoch `256` and Deneb→Electra at `29696`). The Rust port reads the embedded snapshot for the spec (line 496) but ignores its fork_schedule entry in favor of this stub.

Clients that branch on fork transition epochs (e.g. to decide which fork's deserializer to use for a slot) will see different transition points under Pluto vs Charon. Either serve `static_endpoint_data("/eth/v1/config/fork_schedule")` directly, or hard-code the same epoch values that the snapshot uses.

crates/testutil/src/beaconmock/defaults.rs
```diff
@@ -0,0 +1,620 @@
...
+}
+
+fn proposer_duties_response(state: &MockState, request: &Request) -> Value {
+    let Some(factor) = *read_lock(&state.deterministic_proposer_duties) else {
+        return duties_response(Vec::new());
+    };
+
+    let epoch = epoch_from_path(request.url.path());
+    let slots_per_epoch = slots_per_epoch(state);
+    let validators = read_lock(&state.validator_set).validators();
```
varex83agent: **major** — Charon's `WithDeterministicProposerDuties` (options.go:363-405) iterates `mock.ActiveValidators(ctx).Indices()` — only validators where `Status.IsActive()`. This handler iterates `state.validator_set.validators()`, returning every validator regardless of status.

Today this is masked because `Validator::active` is the only constructor and always sets `ActiveOngoing`. But `MockState::set_validator_set` accepts arbitrary `ValidatorSet`s, so any caller that inserts a non-active validator will see it receive proposer duties here but not under Charon. Filter on `validator.status.is_active()` (or document this deliberate divergence).

crates/testutil/src/beaconmock/fuzzer.rs
```diff
@@ -0,0 +1,469 @@
...
+    matchers::{method, path, path_regex},
+};
+
+use crate::random::{
+    random_bit_list, random_eth2_signature, random_phase0_attestation, random_root,
+};
+
+/// Priority for fuzzer routes; must be numerically lower (= higher priority)
+/// than `defaults::DEFAULT_MOCK_PRIORITY` so it overrides default mounts.
+const FUZZ_MOCK_PRIORITY: u8 = 10;
```
varex83agent, line 27: **minor** — `FUZZ_MOCK_PRIORITY = 10` is numerically lower than `OVERRIDE_PRIORITY = 50`, and wiremock uses lower priority as higher precedence. Result: when a caller combines `.fuzzer(true)` with `.endpoint_overrides([...])`, the fuzzer routes silently shadow the user's overrides for any overlapping path.

In Charon the two cannot collide (fuzzer mutates `Mock.Func` fields; overrides target static JSON). Either swap fuzzer to a value > `OVERRIDE_PRIORITY` so user overrides win, or document the precedence explicitly on `BeaconMock::builder().fuzzer(...)`.

crates/testutil/src/beaconmock/headproducer.rs
```diff
@@ -0,0 +1,457 @@
...
+    let start = std::time::Instant::now();
+    let mut rx = state.subscribe();
+    while start.elapsed() < budget {
+        if rx.has_changed().unwrap_or(false) {
+            let _ = rx.borrow_and_update();
+        }
+        if state.current_head().is_some() {
+            return;
+        }
+        std::thread::sleep(Duration::from_millis(1));
```
varex83agent: **minor** — `wait_for_first_head` busy-polls with `std::thread::sleep(1ms)` for up to `slot_duration * 2` from inside a wiremock responder. wiremock runs each `MockServer` on a `current_thread` runtime and holds a write lock on `MockServerState` for the responder's duration, so while this sleep is in progress **every other request to the same mock is serialised behind it**.

With the default `slot_duration = 12s` an `/eth/v1/events` request fired before the first slot has elapsed can stall the whole mock for up to 24s. Today's tests pass an explicit 100ms slot, but the cap should not depend on `slot_duration`. Suggestions: clamp `wait_budget` to e.g. `Duration::from_millis(500)`, or convert to an async responder and `tokio::time::timeout` over `state.subscribe().changed()`.

The subscription at line 306 also dead-ends — `has_changed()` / `borrow_and_update()` exist but never gate termination; only `state.current_head().is_some()` does. Either use the subscription (block on `changed().await` via a small async helper) or drop it.

crates/testutil/src/beaconmock/mod.rs
```diff
@@ -0,0 +1,380 @@
...
+use state::{hex_0x, set_object_field, write_lock};
+
+pub use state::{MockState, Validator, ValidatorSet};
+
+/// Errors returned while configuring `BeaconMock`.
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    /// The generated beacon API client could not be created for the mock URL.
+    #[error("create beacon node api client: {0}")]
+    Client(#[source] anyhow::Error),
```
varex83agent, line 37: **minor** — Wrapping a concrete eth2 error in `anyhow::Error` erases the underlying error chain. `EthBeaconNodeApiClient::with_base_url` returns a typed error; propagate it directly:

```rust
#[derive(Debug, thiserror::Error)]
pub enum Error {
    #[error("create beacon node api client: {0}")]
    Client(#[from] pluto_eth2api::EthBeaconNodeApiClientError),
}
```

Matches the rust-style guideline of "no `anyhow` in library crates; use typed errors everywhere".

crates/testutil/src/beaconmock/headproducer.rs
```diff
@@ -0,0 +1,457 @@
...
+    json!({
+        "slot": head.slot.to_string(),
+        "block": hex_0x(head.block),
+        "execution_optimistic": false,
+    })
+}
+
+fn hex_0x(bytes: impl AsRef<[u8]>) -> String {
+    format!("0x{}", hex::encode(bytes.as_ref()))
+}
```
varex83agent: **minor** — `fn hex_0x(...)` is defined here and in `state.rs:203` with identical bodies. The other beaconmock modules (`attestation.rs:21`, `mod.rs:28`) already import `state::hex_0x`. Drop this local copy and add `use super::state::hex_0x;`.

Similarly, `SharedState::set_current_head` / `current_head` (lines 103-118) reimplement `state::read_lock` / `write_lock`. Either reuse those helpers (they're already `pub(crate)`) or factor poisoned-lock handling into one place.

crates/testutil/src/beaconmock/defaults.rs
```diff
@@ -0,0 +1,620 @@
...
+                        "state_root": ZERO_ROOT,
+                        "body_root": ZERO_ROOT
+                    },
+                    "signature": format!("0x{}", "00".repeat(96))
+                }
+            },
+            "execution_optimistic": false,
+            "finalized": false
+        })
+    })
```
varex83agent, line 95: **minor** — Charon's `/eth/v1/beacon/headers/head` returns the minimal `{"data":{"header":{"message":{"slot":"1"}}}}` (server.go:81-83). The Rust port returns a richer payload with `root`, `canonical`, `parent_root`, `state_root`, `body_root`, `proposer_index`, `signature`, `execution_optimistic`, and `finalized`. This is *more* spec-compliant than Charon's stub, but it is a wire-level deviation worth either documenting as an intentional improvement or trimming to match the Go reference exactly.

crates/testutil/src/beaconmock/defaults.rs
```diff
@@ -0,0 +1,620 @@
...
+        ("DOMAIN_APPLICATION_BUILDER", "0x00000001"),
+        ("EPOCHS_PER_SYNC_COMMITTEE_PERIOD", "256"),
+    ];
+    for (key, value) in overrides {
+        spec.insert((*key).to_string(), Value::String((*value).to_string()));
+    }
+    spec.insert(
+        "MIN_GENESIS_TIME".to_string(),
+        Value::String(default_genesis_time().timestamp().to_string()),
+    );
```
varex83agent, line 579: **minor** — Go's `defaultHTTPMock` does not override `MIN_GENESIS_TIME`, so the value from `static.json` (Holesky's 1695902100) is served verbatim. This unconditionally overwrites it with `default_genesis_time().timestamp()` (2022-03-01 = 1646092800). Any caller that relies on `MIN_GENESIS_TIME` will see different values under Pluto vs Charon. Either omit this override or document the divergence.

crates/testutil/src/beaconmock/mod.rs
```diff
@@ -0,0 +1,380 @@
...
+        fuzzer: Option<bool>,
+        #[builder(default)] endpoint_overrides: Vec<(String, Value)>,
+        fork_version: Option<[u8; 4]>,
+        sync_committee_size: Option<u64>,
+        sync_committee_subnet_count: Option<u64>,
+        #[builder(default)] no_proposer_duties: bool,
+        #[builder(default)] no_attester_duties: bool,
+        #[builder(default)] no_sync_committee_duties: bool,
+        deterministic_sync_comm_duties: Option<(u64, u64)>,
+    ) -> Result<Self> {
```
varex83agent, line 78: **minor** — The builder has 17 optional fields with `#[allow(clippy::too_many_arguments)]` admitting the smell. Eight of those are sync-committee-related (`sync_committee_size`, `sync_committee_subnet_count`, `deterministic_sync_comm_duties`, `no_sync_committee_duties`). Cluster related knobs into nested option structs:

```rust
#[derive(Default)]
pub struct SyncCommitteeOptions {
    pub size: Option<u64>,
    pub subnet_count: Option<u64>,
    pub deterministic_duties: Option<(u64, u64)>,
    pub no_duties: bool,
}
```

This keeps `#[bon]` happy without the lint suppression and groups semantically related knobs.

crates/testutil/src/beaconmock/defaults.rs
```diff
@@ -0,0 +1,620 @@
...
+fn static_endpoint_data(endpoint: &str) -> serde_json::Map<String, Value> {
+    let snapshot: Value =
+        serde_json::from_str(STATIC_JSON).expect("static.json validated by build.rs");
+    snapshot
+        .get(endpoint)
+        .and_then(|entry| entry.get("data"))
+        .and_then(Value::as_object)
+        .cloned()
+        .unwrap_or_default()
+}
```
varex83agent, line 537: nit: `static_endpoint_data` re-parses the embedded ~10KB JSON every time it runs (once per `BeaconMock::builder().build()` when no `spec` is supplied). Cache with `std::sync::LazyLock<Value>`:

```rust
static SNAPSHOT: LazyLock<Value> = LazyLock::new(|| {
    serde_json::from_str(STATIC_JSON).expect("static.json validated by build.rs")
});
```

crates/testutil/src/beaconmock/attestation.rs
```diff
@@ -0,0 +1,353 @@
...
+        entries.retain(|_, old| old.slot.saturating_add(PRUNE_AFTER_SLOTS) >= data.slot);
+        entries.insert(root, data);
+    }
+}
+
+/// Computes the epoch for `slot` given `slots_per_epoch`, mirroring
+/// `eth2util.EpochFromSlot` in Charon's Go code.
+fn epoch_from_slot(slot: Slot, slots_per_epoch: u64) -> Epoch {
+    slot.checked_div(slots_per_epoch).unwrap_or(0)
+}
```
varex83agent, line 75: nit: Charon's `eth2util.EpochFromSlot` panics on `slots_per_epoch == 0` (Go integer division). This silently returns 0. Friendlier, but a real semantic divergence — add a doc comment so anyone diffing semantics doesn't get surprised:

```rust
/// Mirrors `eth2util.EpochFromSlot`, but returns 0 when `slots_per_epoch == 0`
/// instead of panicking.
```

crates/testutil/src/beaconmock/state.rs
```diff
@@ -0,0 +1,272 @@
...
+    pub fn active(index: ValidatorIndex, pubkey: BLSPubKey) -> Self {
+        let pubkey = hex_0x(pubkey);
+
+        Self {
+            index,
+            balance: index,
+            status: ValidatorStatus::ActiveOngoing,
+            validator: ValidatorResponseValidator {
+                activation_eligibility_epoch: index.to_string(),
+                activation_epoch: index.checked_add(1).unwrap_or(index).to_string(),
```
varex83agent, line 47: nit: `index.checked_add(1).unwrap_or(index)` quietly produces `activation_epoch == activation_eligibility_epoch` at `u64::MAX`. Charon does `index + 1` and silently wraps. Either use `index.saturating_add(1)` (same result, clearer intent) or add a comment explaining the saturation since this is unreachable in test inputs.

crates/testutil/src/beaconmock/state.rs
```diff
@@ -0,0 +1,272 @@
...
+        Err(poisoned) => poisoned.into_inner(),
+    }
+}
+
+pub(crate) fn write_lock<T>(lock: &RwLock<T>) -> RwLockWriteGuard<'_, T> {
+    match lock.write() {
+        Ok(guard) => guard,
+        Err(poisoned) => poisoned.into_inner(),
+    }
+}
```
varex83agent, line 238: nit: `read_lock`/`write_lock` silently swallow `PoisonError` via `into_inner()`. Reasonable for test code, but worth either (a) adding a `// Poison is ignored: tests own state, so a previous panic doesn't corrupt invariants.` doc comment, or (b) switching to `parking_lot::RwLock` which has no poisoning and an identical API minus the `Result`.

crates/testutil/src/beaconmock/defaults.rs
```diff
@@ -0,0 +1,620 @@
...
+        "genesis_fork_version": DEFAULT_GENESIS_FORK_VERSION,
+    })
+}
+
+pub(crate) fn default_genesis_time() -> DateTime<Utc> {
+    match Utc.with_ymd_and_hms(2022, 3, 1, 0, 0, 0).single() {
+        Some(time) => time,
+        None => Utc::now(),
+    }
+}
```
varex83agent, line 600: nit: The `None => Utc::now()` fallback would silently break tests asserting on `genesis_time` if it ever fired (which it cannot for this literal date). Prefer `.single().expect("2022-03-01 is valid")` or `DateTime::from_timestamp(1646092800, 0).unwrap()`.

crates/testutil/src/beaconmock/fuzzer.rs
```diff
@@ -0,0 +1,469 @@
...
+        .respond_with(move |request: &Request| ResponseTemplate::new(200).set_body_json(f(request)))
+        .with_priority(FUZZ_MOCK_PRIORITY)
+        .mount(server)
+        .await;
+}
+
+fn aggregate_attestation_response() -> Value {
+    json!({
+        "version": "deneb",
+        "data": random_phase0_attestation(),
```
varex83agent, line 128: nit: `aggregate_attestation_response` hardcodes `"version": "deneb"`. Charon's fuzzer randomises the version field via `fuzz.Fuzz(&att)`. Code under test that branches on `version` will only ever see Deneb under the Rust fuzzer. Same observation applies to `signed_beacon_block_response` (line 162).

crates/testutil/src/beaconmock/headproducer.rs
```diff
@@ -0,0 +1,457 @@
...
+    let now = SystemTime::now();
+    let chain_age = now.duration_since(genesis).unwrap_or_default();
+    let nanos = u64::try_from(slot_duration.as_nanos())
+        .unwrap_or(u64::MAX)
+        .max(1);
+    let height = u64::try_from(chain_age.as_nanos())
+        .unwrap_or(0)
+        .checked_div(nanos)
+        .unwrap_or(0);
+    let multiplier = u32::try_from(height).unwrap_or(u32::MAX);
```
varex83agent, line 189: nit: `u32::try_from(height).unwrap_or(u32::MAX)` silently clamps the multiplier passed to `slot_duration.saturating_mul`. For `height > u32::MAX` the computed `start` lands billions of slots off the real boundary. Unreachable today (12s slots + u32::MAX ≈ 1633 years), but the saturation is hidden. Either `debug_assert!(height <= u32::MAX.into())` to make it loud, or compute via u128 nanoseconds.

crates/testutil/src/beaconmock/headproducer.rs
```diff
@@ -0,0 +1,457 @@
...
+        .and(path_regex(r"^/eth/v1/beacon/blocks/[^/]+/root$"))
+        .respond_with(move |request: &Request| {
+            let Some(head) = state.current_head() else {
+                return error_response(500, "head producer not ready".into());
+            };
+
+            let block_id = extract_block_id(request.url.path());
+            if block_id != "head" && block_id != head.slot.to_string() {
+                return error_response(400, format!("Invalid block ID: {block_id}"));
+            }
```
varex83agent, line 282: nit (deviation from Go): Charon returns HTTP 400 but a JSON body with `"code": 500` (charon/testutil/beaconmock/headproducer.go:201-214) — that's a Go bug. Rust returns `"code": 400`, which is internally consistent. Worth noting in the module preamble so wire-level parity tests don't get surprised.

crates/testutil/src/beaconmock/headproducer.rs
```diff
@@ -0,0 +1,457 @@
...
+}
+
+fn head_event_json(head: &HeadEvent) -> Value {
+    json!({
+        "slot": head.slot.to_string(),
+        "block": hex_0x(head.block),
+        "state": hex_0x(head.state),
+        "epoch_transition": false,
+        "current_duty_dependent_root": hex_0x(head.current_duty_dependent_root),
+        "previous_duty_dependent_root": hex_0x(head.previous_duty_dependent_root),
```
varex83agent: nit (deviation from Go): Charon's `headJSON.PreviousDutyDependentRoot` is set to `currentHead.CurrentDutyDependentRoot` (a Go bug — see headproducer.go:125). The Rust port emits the proper `previous_duty_dependent_root` value. Strictly more correct, but anyone diffing wire output against Charon will see a difference at every head event. Document the intentional deviation in the module preamble.

crates/testutil/src/beaconmock/mod.rs
```diff
@@ -0,0 +1,380 @@
...
+
+        let state = Arc::new(MockState::new(spec, genesis, validator_set));
+        *write_lock(&state.deterministic_attester_duties) = deterministic_attester_duties;
+        *write_lock(&state.deterministic_proposer_duties) = deterministic_proposer_duties;
+        *write_lock(&state.deterministic_sync_comm_duties) = deterministic_sync_comm_duties;
+
+        let server = MockServer::start().await;
+
+        // Higher priority (lower number) mounts must register before the defaults
+        // so wiremock falls back to the default routes when no override matches.
```
varex83agent, line 138: nit: This comment claims `wiremock` falls back based on mount order, but wiremock's precedence is determined by `with_priority` first; registration order is only a tie-breaker between equally-priorities mocks. The current ordering matters only for the priority-50 cohort (`endpoint_overrides` mounted before `no_*_duties`). Reword to describe the actual reason (or remove).

crates/testutil/src/beaconmock/attestation.rs
```diff
@@ -0,0 +1,353 @@
...
+    json!({
+        "version": "fulu",
+        "data": {
+            "aggregation_bits": "0x01",
+            "data": attestation_data_json(data),
+            "signature": format!("0x{}", "00".repeat(96)),
+            "committee_bits": hex_0x(committee_bits),
+        }
+    })
+}
```
varex83agent, line 183: nit (parity): Charon's `AggregateAttestationFunc` (charon/testutil/beaconmock/options.go:569-588) returns a `VersionedAttestation` whose top-level `ValidatorIndex: &valIdx` field is set to 0. The Rust response emits only the inner `data` object. If any consumer reads top-level `validator_index` from this endpoint, parity breaks. Probably no consumer does — flagging for awareness.

crates/testutil/build.rs
```diff
@@ -0,0 +1,79 @@
...
+const STATIC_JSON: &str = "src/beaconmock/static.json";
+
+/// Endpoints that must be present in `static.json`. Mirrors `gen_static.sh`.
+const REQUIRED_ENDPOINTS: &[&str] = &[
+    "/eth/v1/beacon/genesis",
+    "/eth/v1/config/deposit_contract",
+    "/eth/v1/config/fork_schedule",
+    "/eth/v1/node/version",
+    "/eth/v1/config/spec",
+];
```
varex83agent, line 19: nit: `REQUIRED_ENDPOINTS` lists 5 endpoints; `scripts/gen_static_beaconmock.sh` fetches 6 (it also adds `/eth/v2/beacon/blocks/0`). The comment claims they mirror each other. Either drop the obsolete endpoint from the script or add it to the required-endpoints list so the snapshot stays self-consistent.

crates/testutil/src/beaconmock/options.rs
```diff
@@ -0,0 +1,316 @@
...
+            route.and(path(endpoint.clone()))
+        };
+
+        route
+            .respond_with(template)
+            .with_priority(OVERRIDE_PRIORITY)
+            .mount(server)
+            .await;
+    }
+}
```
varex83agent, line 41: nit: The regex-form override (`endpoint.starts_with('^')`) is not exercised by any test. Add at least one test mounting e.g. `^/eth/v1/validator/duties/proposer/[0-9]+$` to lock in the dispatch. Also worth documenting that plain-path callers must escape regex specials if their path contains `.` or `?` (they don't today, but it's a footgun).

varex83agent (dismissed):
## Summary

Functionality-rich port of `charon/testutil/beaconmock` to Rust as `pluto-testutil::beaconmock`, with a wiremock-backed `BeaconMock`, deterministic duty handlers, an SSE/head-root producer, an attestation store, a fuzzer mode, an embedded Holesky `static.json` baseline (validated at compile time by `build.rs`), and three golden fixtures ported byte-for-byte from Charon. Existing tests in `app/eth2wrap/valcache`, `cli/test/beacon`, `core/deadline`, `eth2util/eth2exp`, and `eth2util/signing` are migrated to the shared mock. Quality of the port is high overall; the Go semantics are followed closely and the parity gaps that remain are mostly intentional Go-bug fixes worth documenting.

## Bugs (must-fix)

1. **`fuzzer.rs:182,213,242,264` — pubkeys are 48 copies of a single random byte.** `[rng.r#gen::<u8>(); 48]` evaluates the random byte once and copies it 48 times (`Copy` array-repeat semantics). Every fuzzed pubkey collapses to one of 256 values like `0xaaaa...aa`, gutting the fuzz coverage of any code that branches on pubkey content. Use `let mut pk = [0u8; 48]; rng.fill_bytes(&mut pk);`.

## Major / parity

2. **`pluto-testutil` is a normal `[dependencies]` entry of `pluto-eth2util` (`crates/eth2util/Cargo.toml:24`).** Not modified in this PR but materially worsened by it: `wiremock` (added to `testutil`'s `[dependencies]` here) now transitively links into every production crate that depends on `eth2util` (`cli`, `cluster`, `core`, `p2p`, `app`, the `pluto` binary). All current usages inside `eth2util` are `#[cfg(test)]`, so nothing is *called* in release, but the entire mock infrastructure is still compiled and shipped. `crates/cluster/Cargo.toml` already demonstrates the right fix: `pluto-testutil = { workspace = true, optional = true }` plus a feature flag.
3. **`headproducer.rs:83` — `Drop::notify_waiters()` races with the spawned ticker.** `notify_waiters` only wakes tasks already polling `notified()`. If `BeaconMock` is constructed-then-dropped before the spawned task at line 140 first polls, the shutdown signal is lost and the ticker leaks. Switch to `tokio_util::sync::CancellationToken` (persistent + shareable) or `Notify::notify_one()` (stores a permit).
4. **`state.rs:46-49` — `Validator::active` sets `exit_epoch`/`withdrawable_epoch` to `u64::MAX`, but Charon's `ValidatorSetA` leaves them at 0.** These `u64::MAX` defaults come from Go's *fuzzer* (`beaconmock_fuzz.go:44-45`), not from `ValidatorSetA`. Wire JSON for `validator_set_a()` will not match Charon's.
5. **`defaults.rs:42-53` — `fork_schedule` is hand-rolled with `epoch="0"` everywhere**, while the embedded `static.json` (which Charon actually serves) has Capella→Deneb at epoch 256 and Deneb→Electra at 29696. Read the fork_schedule from the snapshot like the spec does.
6. **`defaults.rs:295-341` — `proposer_duties_response` iterates the full validator set**, but Charon's `WithDeterministicProposerDuties` iterates *active* validators only (`mock.ActiveValidators(ctx).Indices()`). Inactive validators inserted via `set_validator_set` will receive duties under the Rust mock but not under Charon's.

## Notable minors

- **`fuzzer.rs:26` — `FUZZ_MOCK_PRIORITY=10` is *lower* (= higher precedence) than `OVERRIDE_PRIORITY=50`**, so user `endpoint_overrides` are silently shadowed by fuzzer routes when both are enabled. In Charon they cannot collide (fuzzer goes through `Mock.Func` fields, overrides go through static JSON). Either flip the priorities or document the precedence.
- **`headproducer.rs:298` — `wait_for_first_head` busy-polls `std::thread::sleep(1ms)` inside a wiremock responder for up to `slot_duration * 2`.** With the default 12s slot, an early `/eth/v1/events` request can stall the entire mock server thread for ~24s. Cap the budget at a small fixed value or convert to async.
- **`mod.rs:37` — `Error::Client(#[source] anyhow::Error)`** wraps a typed eth2 error in `anyhow`. Use `#[from] pluto_eth2api::EthBeaconNodeApiClientError`.
- **`headproducer.rs:347` + `state.rs:203` — `hex_0x` is defined in both files.** All other beaconmock modules already import the `state::hex_0x` version.

## Verdict

The fuzzer pubkey bug is a clear must-fix, and several parity gaps (validator default fields, fork_schedule epochs, proposer-duty active-only filter, fuzz/override priority inversion) should be addressed or explicitly documented before merge. Once the bug and the parity items are resolved, this is a solid baseline.

crates/testutil/src/beaconmock/defaults.rs
```diff
@@ -0,0 +1,623 @@
...
+        ("DOMAIN_SYNC_COMMITTEE_SELECTION_PROOF", "0x08000000"),
+        ("DOMAIN_CONTRIBUTION_AND_PROOF", "0x09000000"),
+        ("DOMAIN_APPLICATION_BUILDER", "0x00000001"),
+        ("EPOCHS_PER_SYNC_COMMITTEE_PERIOD", "256"),
+    ];
+    for (key, value) in overrides {
+        spec.insert((*key).to_string(), Value::String((*value).to_string()));
+    }
+    spec.insert(
+        "MIN_GENESIS_TIME".to_string(),
```
varex83agent, line 577: `default_spec()` unconditionally inserts `MIN_GENESIS_TIME = default_genesis_time().timestamp()` (`1646092800`, 2022-03-01). Charon's `defaultHTTPMock` (`charon/testutil/beaconmock/beaconmock.go:78-169`) does **not** include `MIN_GENESIS_TIME` in its overrides, so Go serves whatever `static.json` carries (`1695902100`, Holesky genesis). Pluto therefore diverges silently for any consumer that reads `MIN_GENESIS_TIME` without first calling `.genesis_time(...)`.

Suggested fix: drop the unconditional override and only set `MIN_GENESIS_TIME` from the builder when `genesis_time` is provided (`mod.rs:98-102` already handles the override case).

crates/testutil/src/beaconmock/fuzzer.rs
```diff
@@ -0,0 +1,467 @@
...
+    };
+
+    route
+        .respond_with(move |request: &Request| ResponseTemplate::new(200).set_body_json(f(request)))
+        .with_priority(FUZZ_MOCK_PRIORITY)
+        .mount(server)
+        .await;
+}
+
+fn aggregate_attestation_response() -> Value {
```
varex83agent, line 125: `aggregate_attestation_response` hardcodes `{"version": "deneb", "data": random_phase0_attestation()}`. The payload shape is phase0 but the tag says deneb — internally inconsistent, and Charon's default attestation mount uses `version=fulu`. Charon's fuzzer (`beaconmock_fuzz.go:121-126`) randomises both the version and the matching body via `eth2spec.VersionedAttestation`.

Either label as the matching version (e.g. `phase0`/`fulu`) or randomise the version per call and shape the body accordingly.

crates/testutil/src/beaconmock/attestation.rs
```diff
@@ -0,0 +1,353 @@
...
+        .respond_with({
+            let state = Arc::clone(&state);
+            move |request: &Request| aggregate_attestation_response(&state, request)
+        })
+        .with_priority(ATTESTATION_PRIORITY)
+        .mount(server)
+        .await;
+}
+
+fn attestation_data_response(state: &MockState, request: &Request) -> Value {
```
varex83agent: If `SLOTS_PER_EPOCH` is missing or unparseable, `attestation_data_response` silently falls back to `16`. Charon's `eth2util.EpochFromSlot` calls `mock.SlotsPerEpoch(ctx)` which surfaces `errors.New("fetch slots per epoch")` (`options.go:635-646`), and the handler returns an error.

A silent fallback can mask misconfiguration in tests. Either propagate an error response or hoist the fallback to a single named constant so the divergence is documented in one place.

crates/testutil/src/beaconmock/defaults.rs
```diff
@@ -0,0 +1,623 @@
...
+    })
+}
+
+fn sync_committee_duties_response(state: &MockState, request: &Request) -> Value {
+    let Some((n, k)) = *read_lock(&state.deterministic_sync_comm_duties) else {
+        return sync_duties_response(Vec::new());
+    };
+
+    let epoch = epoch_from_path(request.url.path());
+    let Some(remainder) = epoch.checked_rem(k) else {
```
varex83agent, line 454: `epoch.checked_rem(k)` returns `None` when `k = 0`, and the handler silently returns an empty list. Charon's `WithDeterministicSyncCommDuties` (`options.go:441`) does `int(epoch) % k` which panics on `k=0`.

Pluto is more robust but diverges. Either add `assert!(k > 0)` (or `debug_assert!`) at the builder seam to match Go's contract, or document the intentional divergence in `deterministic_sync_comm_duties`'s docstring — silent empty duties are the worst-case test bug (passes when broken).

crates/testutil/src/beaconmock/headproducer.rs
```diff
@@ -0,0 +1,423 @@
...
+            next_tick = next_tick.checked_add(slot_duration).unwrap_or_else(|| {
+                SystemTime::now()
+                    .checked_add(slot_duration)
+                    .unwrap_or(SystemTime::now())
+            });
+        }
+    });
+}
+
+fn normalize_slot_duration(slot_duration: Duration) -> Duration {
```
varex83agent, line 171: `normalize_slot_duration` silently substitutes `Duration::from_millis(1)` for a zero `slot_duration`. Charon's `startSlotTicker` (`headproducer.go:272-292`) computes `height := int64(chainAge / slotDuration)` which would panic on division-by-zero.

Pluto is strictly more lenient. Probably fine for a test mock, but worth a one-line comment so a future reader doesn't assume Go parity here.

crates/testutil/src/beaconmock/headproducer.rs
```diff
@@ -0,0 +1,423 @@
...
+            }
+
+            // `HeadProducer::spawn` publishes the initial head before
+            // returning, so the current head is always set here.
+            let Some(head) = state.current_head() else {
+                return error_response(500, "head producer not ready".into());
+            };
+
+            let mut body = String::new();
+            if topics.is_empty() || topics.iter().any(|t| t == TOPIC_HEAD) {
```
varex83agent, line 252: When no `topics` query parameter is supplied, the handler emits a head event by default (`topics.is_empty() || ...`). Charon's `handleEvents` iterates `query["topics"]` and registers no streams when the list is empty, so the SSE connection publishes nothing.

A client probing `/eth/v1/events` with no topics gets a head event in Pluto but nothing in Charon. Either drop the `topics.is_empty()` arm or call out the divergence next to the existing SSE note in the module docstring.

crates/testutil/src/beaconmock/defaults.rs
```diff
@@ -0,0 +1,623 @@
...
+        .unwrap_or(16)
+}
+
+/// Embedded beacon-node snapshot used as the baseline for default responses.
+///
+/// Generated by `scripts/gen_static_beaconmock.sh` against a Holesky beacon
+/// node. Validated at compile time by `build.rs`.
+pub(crate) const STATIC_JSON: &str = include_str!("static.json");
+
+fn static_endpoint_data(endpoint: &str) -> serde_json::Map<String, Value> {
```
varex83agent, line 528: `static_endpoint_data` runs `serde_json::from_str(STATIC_JSON)` (an ~80-key JSON map) on every call, and every `BeaconMock::builder().build()` calls `default_spec()` unconditionally — even when the caller supplies an explicit spec.

Wrap the parsed snapshot in `std::sync::LazyLock<Value>` (or `OnceLock<Value>`) at module scope and clone the relevant subtree. Cheap perf win for tests that spin up many mocks.

crates/testutil/src/beaconmock/mod.rs
```diff
@@ -0,0 +1,426 @@
...
+        for (endpoint, value) in endpoint_overrides {
+            mount_endpoint_override(&server, endpoint, value).await;
```
varex83agent, line 140: Both `mount_endpoint_override` (this loop) and `mount_no_{proposer,attester,sync_committee}_duties` mount routes at `OVERRIDE_PRIORITY = 50` (`options.rs:18`). When a caller passes a custom override for `/eth/v1/validator/duties/proposer/{epoch}` **and** sets `no_proposer_duties(true)`, both routes match the same request at the same priority, and wiremock's tie-break between equal-priority matchers depends on registration order.

Either document which wins, or split the priorities (e.g. user overrides at 40, builtin no-duties at 50).

varex83agent (commented):
## Beaconmock port — overall assessment

This is a substantial, careful port of Charon's `testutil/beaconmock` to Rust. Functional parity with the Go reference is good after the prior round of fixes (zero-value exit/withdrawable epochs, active-validator filter on proposer duties, synchronous initial head publish via `CancellationToken`, mirrored duty-root typo, etc.).

**No `bug` or `major` findings.** What remains are minor parity divergences, a few perf/idiom cleanups, and nits. None of them block merge.

Themes worth addressing before or shortly after merge:
1. **Charon parity tweaks** in defaults/handlers where Pluto is silently more lenient than Go (`MIN_GENESIS_TIME` override, fuzzer aggregate version, `slot_duration=0` and `k=0` fallbacks, empty `topics=` returning a head event). Each one is small but they will add up if not tracked.
2. **Code-dup cleanups** — three near-identical `mount_json*` helpers could collapse, and the duty-init locking on a fresh `Arc` can move into the constructor.
3. **Test-helper hygiene** — `default_spec()` re-parses the embedded ~80-key static.json on every builder call; ticker `JoinHandle` is discarded; `endpoint_overrides` and `no_*_duties` share priority 50 with no documented tie-break.

**Note (out of scope):** This PR also bundles `.claude/skills/loop-review-pr/SKILL.md`, settings, and `.gitignore` tweaks that aren't part of the beaconmock port. Worth splitting into a separate change to keep the port reviewable in isolation.

Verdict: **COMMENT** — leaving the inline notes for author discretion.

crates/testutil/src/beaconmock/attestation.rs
```diff
@@ -0,0 +1,353 @@
...
+    key: &'static str,
+}
+
+impl wiremock::Match for QueryParamPresent {
+    fn matches(&self, request: &Request) -> bool {
+        request.url.query_pairs().any(|(k, _)| k == self.key)
+    }
+}
+
+async fn mount_json_with_priority<F>(
```
varex83agent: `attestation::mount_json_with_priority` (here), `fuzzer::mount_fuzz_json`, and `defaults::mount_json_with_status` are near-identical: build a route, attach `path` or `path_regex` based on a leading `^`, set status, set a JSON body via a closure, apply a priority. Consolidating into a single shared helper (e.g. `mount_json(server, method, endpoint, MountOpts { priority, status }, f)`) would remove ~60 lines of duplication and the copy-paste hazard of keeping them in sync.

crates/testutil/src/beaconmock/headproducer.rs
```diff
@@ -0,0 +1,423 @@
...
+    // The initial head was already published by `HeadProducer::spawn`. Start
+    // the ticker at the next scheduled slot so it advances from there.
+    let mut height = initial_height.wrapping_add(1);
+    let mut next_tick = initial_tick.checked_add(slot_duration).unwrap_or_else(|| {
+        SystemTime::now()
+            .checked_add(slot_duration)
+            .unwrap_or(SystemTime::now())
+    });
+
+    tokio::spawn(async move {
```
varex83agent, line 148: `spawn_slot_ticker` calls `tokio::spawn` and discards the `JoinHandle`. `HeadProducer::drop` signals cancellation via `cancel.cancel()` but never awaits the task — the ticker only exits at its next `tokio::select!` await. In normal test use this is fine; if a test panics while the runtime is draining, the task could linger without a synchronization point.

Storing the `JoinHandle` and `.abort()`-ing it on drop (or awaiting it under a small timeout) would make shutdown deterministic. Alternatively, a comment explaining why detached lifecycle is acceptable for this test helper.

crates/testutil/src/beaconmock/fuzzer.rs
```diff
@@ -0,0 +1,467 @@
...
+    json!({
+        "data": data,
+        "execution_optimistic": false,
+        "finalized": false,
+    })
+}
+
+fn attester_duties_response(epoch: u64) -> Value {
+    let mut rng = rand::thread_rng();
+    let slots_per_epoch = 16u64;
```
varex83agent, line 205: `attester_duties_response` and `proposer_duties_response` both hardcode `let slots_per_epoch = 16u64;` (here and on line 234). If the caller uses `BeaconMock::builder().slots_per_epoch(...)`, the fuzzer's generated slots no longer fall within the requested epoch's slot window.

Either read it from `MockState::spec` (as `defaults::slots_per_epoch` does) or extract a named const with a comment noting the fuzzer intentionally ignores user overrides.

crates/testutil/src/beaconmock/mod.rs
```diff
@@ -0,0 +1,426 @@
...
+        if let Some(count) = sync_committee_subnet_count {
+            set_object_field(&mut spec, "SYNC_COMMITTEE_SUBNET_COUNT", count.to_string());
+        }
+
+        if let Some((n, _)) = deterministic_sync_comm_duties {
+            set_object_field(&mut spec, "EPOCHS_PER_SYNC_COMMITTEE_PERIOD", n.to_string());
+        }
+
+        let state = Arc::new(MockState::new(spec, genesis, validator_set));
+        *write_lock(&state.deterministic_attester_duties) = deterministic_attester_duties;
```
varex83agent, line 131: Lines 131-133 take three `write_lock` write guards on a freshly-constructed `Arc<MockState>` purely to set initial values for the deterministic-duty knobs. The locks aren't actually contended yet (the `Arc` hasn't been shared).

Push these into `MockState::new(..., deterministic_attester_duties, deterministic_proposer_duties, deterministic_sync_comm_duties)` so the constructor produces a fully-initialized value, dropping three superfluous lock acquisitions plus the `write_lock` import.

crates/testutil/src/beaconmock/headproducer.rs
```diff
@@ -0,0 +1,423 @@
...
+        mount_events(server, Arc::clone(&state)).await;
+        mount_block_root(server, Arc::clone(&state)).await;
+
+        let genesis = system_time_from(genesis_time);
+        let slot_duration = normalize_slot_duration(slot_duration);
+        let (initial_height, initial_tick) = initial_slot(genesis, slot_duration);
+
+        // Publish the initial head before handing control back to the caller
+        // so the mounted handlers never see a None current head.
+        update_head(&state, initial_height);
```
varex83agent, line 86: The module doc (lines 19-22) claims the initial head is published before mounted handlers can observe a `None` current head. `update_head(&state, initial_height)` runs here at line 86, **after** `mount_events` / `mount_block_root` are awaited (lines 77-78). `MockServer::start()` accepts connections by the time mounts complete, so a hypothetical concurrent client hitting the URL during this window would observe `None`.

In practice unreachable (the caller doesn't have the URL yet), but the doc comments overstate the guarantee. Either swap the order (publish initial head before mounting handlers) or weaken the comments to "by the time `spawn` returns".

crates/testutil/build.rs
```diff
@@ -0,0 +1,79 @@
...
+//! Catches regressions in the embedded beacon-node snapshot (malformed JSON,
+//! missing endpoints, missing spec keys) before the test crate runs, and
+//! triggers a rebuild whenever the snapshot changes.
+
+use std::path::Path;
+
+const STATIC_JSON: &str = "src/beaconmock/static.json";
+
+/// Endpoints that must be present in `static.json`. Mirrors `gen_static.sh`.
+const REQUIRED_ENDPOINTS: &[&str] = &[
```
varex83agent, line 12: nit: `gen_static_beaconmock.sh` lists `/eth/v2/beacon/blocks/0` in its `ENDPOINTS` array, but `REQUIRED_ENDPOINTS` here omits it. The runtime handler for `/eth/v2/beacon/blocks/...` builds its response from scratch (not from `static.json`), so the snapshot entry is currently dead weight — but the build script won't catch its removal either. Either add the endpoint to `REQUIRED_ENDPOINTS` or drop it from the shell script and `static.json`.

crates/testutil/src/beaconmock/options.rs
```diff
@@ -0,0 +1,316 @@
...
+/// Priority for builder-driven overrides. Lower numeric priority wins in
+/// `wiremock`, so this sits above [`super::defaults::DEFAULT_MOCK_PRIORITY`]
+/// (255) and below any test-supplied overrides mounted directly via
+/// [`BeaconMock::server`](super::BeaconMock::server).
+pub(crate) const OVERRIDE_PRIORITY: u8 = 50;
+
+/// Mounts a static JSON override for `endpoint` returning `value`.
+///
+/// `endpoint` may be either a plain path or a regex prefixed with `^`.
+pub(crate) async fn mount_endpoint_override(server: &MockServer, endpoint: String, value: Value) {
```
varex83agent, line 23: nit: `mount_endpoint_override` takes `endpoint: String` but clones it twice inside (once for `path_regex(endpoint.clone())`, once for `path(endpoint.clone())`) and clones `value` once. Taking `endpoint: &str` and only owning where the matcher needs it (and `value: &Value`) avoids the unnecessary copy; callers in `mod.rs:140` already own the values.

crates/testutil/src/beaconmock/headproducer.rs
```diff
@@ -0,0 +1,423 @@
...
+}
+
+fn random_root(rng: &mut StdRng) -> Root {
+    let mut root = Root::default();
+    rng.fill_bytes(&mut root);
+    root
+}
+
+async fn mount_events(server: &MockServer, state: Arc<SharedState>) {
+    Mock::given(method("GET"))
```
iamquang95, line 237: Do we need to handle stream here? the `/eth/v1/events` in charon create a stream, not a single response

crates/testutil/src/beaconmock/defaults.rs
```diff
@@ -0,0 +1,623 @@
...
+    last_path_segment_u64(path)
+}
+
+fn slots_per_epoch(state: &MockState) -> u64 {
+    read_lock(&state.spec)
+        .get("SLOTS_PER_EPOCH")
+        .and_then(Value::as_str)
+        .and_then(|value| value.parse().ok())
+        .filter(|slots| *slots > 0)
+        .unwrap_or(16)
```
iamquang95: Should not `unwrap_or`, return error instead: https://github.com/ObolNetwork/charon/blob/2c42245a686ed61d44716be52a1e5d4a34529a6b/testutil/beaconmock/options.go#L709

crates/testutil/build.rs
iamquang95, line 19: Missing `/eth/v2/beacon/blocks/0`

iamquang95 (approved):

crates/testutil/build.rs
varex83agent: Done in next commit — added `/eth/v2/beacon/blocks/0` to `REQUIRED_ENDPOINTS`. The static.json snapshot stores it as a 404 envelope (Holesky has no block at slot 0 over the v2 API), so I also relaxed `build.rs` to accept either a `data` payload or an error envelope (`code` + `message`) for required endpoints.

crates/testutil/src/beaconmock/defaults.rs
varex83agent: Done in next commit — `slots_per_epoch` now returns `Result<u64, &'static str>` mirroring Charon's `SlotsPerEpochFunc` (`testutil/beaconmock/options.go:635-647`), and the three handlers that consumed it (`attester_duties_response`, `proposer_duties_response`, `attestation_data_response`) propagate the error as `500 {"code":500,"message":"failed to lookup or invalid SLOTS_PER_EPOCH from spec"}`. Required new `mount_response` / `mount_response_with_priority` helpers since the existing `mount_json*` always wraps in 200.

crates/testutil/src/beaconmock/headproducer.rs
varex83agent: Good catch and you're right about the parity — Charon serves `/eth/v1/events` as a long-lived SSE stream via `r3labs/sse` (`testutil/beaconmock/headproducer.go:230-262`), publishing on every slot tick.

I checked the path and the divergence is intentional: wiremock-rs 0.6 has no streaming response API — every `respond_with` returns a buffered `hyper::Response<Full<Bytes>>`, so a true long-lived stream isn't expressible within the wiremock model. To match Charon we'd need to either (a) stand up a separate axum/hyper task next to wiremock for this single endpoint with port-sharing/proxying, or (b) replace wiremock wholesale.

The current handler returns a single, well-formed SSE record per request (with the current head/block), and no Pluto consumer reads `/eth/v1/events` today — only the mock's own round-trip test does. I documented the divergence in the module doc (`headproducer.rs:9-13`). Happy to upgrade to a real stream if/when a consumer needs it; would do it in a follow-up since it's a bigger architectural change than the rest of this PR.

Issue #393, opened 2026-05-12
Rewrite `core/qbft` to async-Rust-friendly API


varex83:
## Context

The current Rust port of `core/qbft` (`crates/core/src/qbft/mod.rs`) is a faithful translation of the Go reference but built on **synchronous, thread-based primitives**:

- `crossbeam::channel` for transport
- `std::thread::scope` for spawning per-process workers
- A blocking `pub fn run(...)` entry point
- `cancellation::CancellationToken` (third-party crate, not `tokio_util`)

This is at odds with the rest of Pluto, which is **tokio-async end to end**. Every consumer that will eventually wire QBFT in (`core/consensus`, `core/scheduler`, the duty pipeline) is async. The current shape forces either (a) bridging via `spawn_blocking` and ad-hoc channel adapters, or (b) blocking inside async tasks — both bad.

Closes the gap left by #13, which produced the initial sync port.

## Goal

Rewrite `core/qbft` with an **async-Rust-native API** that integrates cleanly with the rest of the workspace and preserves the Byzantine-safety guarantees of the Go reference.

### API
- `run` becomes `pub async fn run(...) -> Result<Decision, QbftError>`
- `Transport` trait methods return `impl Future<Output = Result<...>> + Send` (or are `async fn` via trait async)
- Cancellation uses `tokio_util::sync::CancellationToken` (already used elsewhere in Pluto)
- Channels: `tokio::sync::mpsc` (and `oneshot` where appropriate)
- No `std::thread`, no `crossbeam`, no `std::sync::mpsc` anywhere in the public or internal API

### Internals
- Replace `thread::scope` with `tokio::spawn` + `JoinSet` (or `select!` for per-instance loops)
- Replace blocking sleeps/timers with `tokio::time::sleep` / `Interval`
- Clock abstraction (`fake_clock.rs`) becomes async-aware so deterministic tests still work — likely using `tokio::time::pause()` + `advance()` rather than the current fake clock
- Keep the algorithm shape (rounds, justifications, message bookkeeping) bit-for-bit identical to the Go reference — only the concurrency primitives change

### Safety net (mandatory)
- Port `charon/core/consensus/qbft/strategysim_internal_test.go` (~1040 LOC) as the parity gate. This is the test that catches BFT correctness regressions; without it we cannot trust the rewrite.
- All current `crates/core/src/qbft/internal_test.rs` cases pass (or are replaced by equivalent async versions)

## Acceptance criteria

[...]

PR #417, opened 2026-05-15
fix: QBFT

iamquang95:
Part of: https://github.com/NethermindEth/pluto/issues/393

This PR
- Hardens the QBFT tests. Makes the test parity with charon: Previously, some params are misconfigured.
- Add fuzz tests for QBFT matches with charon.
- Add some more unit tests which helps the fix QBFT stricter.
- Add more chain split test.
- Fix the existing QBFT implementation to pass tests.


iamquang95:
@claude review the changes

crates/core/src/qbft/mod.rs
```diff
@@ -503,7 +527,8 @@ where
 
                                         broadcast_round_change()?;
                                     }
-                                    _ => panic!("bug: expected only {} or {} error", QbftError::CompareError, QbftError::TimeoutError)
+                                    QbftError::ContextCanceled => return Err(QbftError::ContextCanceled),
+                                    _ => return Err(QbftError::UnexpectedCompareError),
```
varex83agent: **Compare callback exit/panic kills the QBFT instance**

When `d.compare` returns/panics without sending on `compare_err_tx`, the spawned thread drops both senders. `recv(compare_err_rx)` at mod.rs:692-699 then returns `Err(Disconnected)`, `compare` returns `QbftError::ChannelError(...)`, and that falls through this `_` arm to `UnexpectedCompareError`, terminating `run`.

Charon Go (qbft.go:438-466) does not have this failure mode — a misbehaving callback simply never signals, the `select` keeps waiting, the round timer eventually fires, and the algorithm recovers via a normal `errTimeout` round change.

This is a robustness regression with security implications: if a malicious leader crafts a PRE-PREPARE whose value triggers a panic in every honest validator's compare callback, the whole cluster aborts consensus deterministically.

Recommendation: either map `ChannelError` here to `TimeoutError` semantics (let the round timer drive recovery), or catch unwinds inside the spawned thread and translate them to `CompareError`. There is also an asymmetry to fix: `recv(compare_value_rx)` at mod.rs:716-721 already treats sender disconnect benignly (`mpmc::never()` and continue), but `recv(compare_err_rx)` at mod.rs:692-699 does not.

crates/core/src/qbft/mod.rs
```diff
@@ -615,46 +661,82 @@ where
...
+    let compare = d.compare.clone();
+    let compare_cts = sync::Arc::new(CancellationTokenSource::new());
+    let compare_ct = compare_cts.token().clone();
+    let msg = msg.clone();
+    let input_value_source_ch = input_value_source_ch.clone();
+
+    // Detached by design, matching Charon's goroutine behavior: if a
+    // caller-provided compare callback ignores cancellation and never reports,
+    // it may outlive this call.
+    thread::spawn(move || {
```
varex83agent, line 682: **Detached compare thread can leak unboundedly**

The doc comment correctly notes this is intentional Charon-parity, but the trade-off in Rust is different: each `UPON_JUSTIFIED_PRE_PREPARE` spawns a fresh OS thread that captures `Arc<CompareFn>`, a cloned `Msg`, the input-value receiver clone, both bounded(1) senders, and the cloned `compare_ct`. If a caller-supplied callback blocks on `input_value_source_ch.recv()` (or any long operation) and ignores `compare_ct`, the thread outlives `compare()` indefinitely and retains all captured state.

A Byzantine peer forcing many round changes — or a long-running validator running many consensus instances — accumulates one leaked thread per round-change/instance. Charon has the same theoretical issue with goroutines, but goroutines are ~2KB stack vs an OS thread's ~8MB default, so the Rust port is much more sensitive to this class of bug.

Mitigations: (a) document on `Definition::compare` that callbacks MUST observe `compare_ct` and exit promptly; (b) add a `tracing::warn!` (or metric) when a detached thread is still alive after `compare()` returns; (c) keep an `Arc<AtomicUsize>` of in-flight workers so leaks are observable in production.

crates/core/src/qbft/internal_test.rs
```diff
@@ -786,5 +1304,825 @@ fn duplicate_pre_prepare_rules() {
...
+    });
+
+    let (result_tx, result_rx) = mpmc::bounded(1);
+    thread::spawn(move || {
+        result_tx
+            .send(compare(&token, &def, &msg, &vs_rx, 0, &timer_rx))
+            .expect(WRITE_CHAN_ERR);
+    });
+
+    thread::sleep(Duration::from_millis(10));
```
varex83agent: **`thread::sleep(Duration::from_millis(10))` is unreliable test synchronization**

The sequence at lines 1685-1692 spawns `compare()` in a thread, sleeps 10ms, then cancels. On a heavily loaded CI runner the spawned thread may not have entered `compare()` within 10ms — if `cts.cancel()` fires before `compare()` starts its loop, `compare()` returns `ContextCanceled` at mod.rs:687-689 *without ever invoking the inner callback*. The `token_cancelled_rx.recv_timeout(100ms)` at line 1702 then times out and panics.

The pattern at lines 1745-1755 (`run_parent_cancel_during_compare_does_not_prepare`) uses a `compare_started_rx` signal to wait for the callback to actually enter before cancelling — apply the same fix here. The same issue affects `idle_run_returns_when_cancelled` (line 1330) and `run_cancels_under_hot_receive_stream` (line 1414), but those are saved by the polling cancellation check at mod.rs:440.

crates/core/src/qbft/mod.rs
```diff
@@ -33,6 +33,19 @@ use std::{
 
 type Result<T> = std::result::Result<T, QbftError>;
 
+const CANCELLATION_POLL_INTERVAL: time::Duration = time::Duration::from_millis(1);
```
varex83agent: **1ms cancellation polling replaces Go's event-driven `ctx.Done()`**

The `default(CANCELLATION_POLL_INTERVAL)` arms at mod.rs:607-611 (`run`) and mod.rs:732-737 (`compare`) cause each idle QBFT instance to wake roughly 1000 times per second. With many concurrent QBFT instances (one per duty per slot per cluster) this aggregates to noticeable wasted CPU.

Charon's main `select` includes `<-ctx.Done()` directly so an idle instance consumes ~0 CPU until cancellation. The `cancellation` crate doesn't expose a channel-shaped wake signal, but you could bridge it once at the top of `run`: spawn a tiny task that sends on a dedicated `cancel_rx: mpmc::Receiver<()>` when `ct` fires, and add `recv(cancel_rx)` to both `select!` blocks. That eliminates the polling entirely.

At minimum, the 1ms constant deserves a comment explaining the trade-off — and probably a much larger value (50-100ms) since cancellation latency on consensus shutdown isn't tight.

crates/core/src/qbft/internal_test.rs
```diff
@@ -309,15 +518,98 @@ fn test_qbft(test: Test) {
...
+    })
+}
+
 /// Construct a leader election function.
 fn make_is_leader(n: i64) -> impl Fn(&i64, i64, i64) -> bool + Clone {
-    move |instance: &i64, round: i64, process: i64| -> bool { (instance + round) % n == process }
+    move |instance: &i64, round: i64, process: i64| -> bool {
+        // Go's helper uses 0-indexed process IDs. These Rust tests use 1..=n,
+        // so translate the same modulo schedule into the local process range.
+        ((instance + round - 1).rem_euclid(n)) + 1 == process
```
varex83agent: **`make_is_leader` doesn't match Charon's leader schedule at rounds multiples of n**

Charon's helper (`charon/core/qbft/qbft_internal_test.go:683-687`) is `(instance+round) % n == process`, with the test loop iterating `i := 1..=n` (line 882). For instance=0, round=4, n=4 it yields `process=0`, which matches no process in 1..=n — so round 4 has **no leader** in Charon, the round must time out.

The Rust translation here reshapes the modulo to always produce 1..=n: `((instance + round - 1) % n) + 1`. For instance=0, round=4 the Rust helper returns process 4, so a leader **is** elected. The comment claiming Go is 0-indexed is incorrect — Go also iterates 1..=n.

This is a hidden parity gap: tests reaching rounds at multiples of n exercise a different leader rotation than Charon and may mask bugs Charon's tests catch. `zz_chain_split_no_consensus_halt` is one such case — it ramps to MAX_ROUND=10 and passes through rounds 4 and 8. Recommend mirroring Go exactly (return a sentinel like 0 that no in-range process equals) so future tests stay faithful.

crates/core/src/qbft/mod.rs
```diff
@@ -834,6 +916,13 @@ where
     let pr = msg.prepared_round();
     let pv = msg.prepared_value();
 
+    // The IBFT paper requires ROUND-CHANGE prepared_round to be lower than the
+    // target round. Go core currently omits this check, but valid Charon traffic
+    // already satisfies it.
+    if !valid_round_change_prepared_round(msg) {
```
varex83agent, line 930: **`valid_round_change_prepared_round` rejects messages Charon would accept — potential one-way interop break**

The comment correctly notes "Go core currently omits this check." That holds for honest Charon peers because `preparedRound` is set to `round` only inside `UponQuorumPrepares` (qbft.go:376) and a ROUND-CHANGE is only broadcast after `changeRound(round+1, ...)` moves the round forward.

However, this is a one-way interop break: if a Charon node ever emits `pr == round` (e.g. future Go regression, or a peer running modified Charon), Pluto silently drops the message. If Pluto and Charon nodes are mixed in a cluster, the divergence is invisible until honest Pluto nodes start ignoring honest Charon traffic.

Recommendation: coordinate with the Charon team to add the same check in `qbft.go` upstream (so the spec is enforced both ways), or at minimum log a warning when the check fires so divergence is observable in production. The same applies to the filters at mod.rs:1040, 1139, and 1290.

crates/core/src/qbft/mod.rs
```diff
@@ -867,6 +956,14 @@ where
     true
 }
 
+fn valid_round_change_prepared_round<I, V, C>(msg: &Msg<I, V, C>) -> bool
```
varex83agent, line 967: **No dedicated unit tests for `valid_round_change_prepared_round`**

This Byzantine-input filter is wired into four call sites (`is_justified_round_change`, `contains_justified_qrc`, `get_justified_qrc`, `quorum_null_prepared`) and is consensus-safety-critical: too strict and honest traffic is dropped (liveness break); too lax and a malicious `prepared_round` leaks through (safety break).

The PR adds no direct boundary-case tests for: `pr < 0`, `pr == 0` with `round == 1`, `pr == round - 1`, `pr == round`, `pr > round`. The existing `round_change_prepared_round_must_be_before_round` test covers one case but not the full matrix. Add a focused table-driven test on `valid_round_change_prepared_round` plus per-call-site tests verifying that each invalid `pr` value is filtered out by `is_justified_round_change` / `contains_justified_qrc` / `get_justified_qrc`.

crates/core/src/qbft/mod.rs
```diff
@@ -41,12 +54,21 @@ pub enum QbftError {
...
+    ContextCanceled,
+
     #[error("Maximum round reached")]
     MaxRoundReached,
 
     #[error("Zero input value not supported")]
     ZeroInputValue,
 
+    #[error("Invalid definition: {0}")]
+    InvalidDefinition(&'static str),
```
varex83agent: **`InvalidDefinition(&'static str)` constrains future detail**

The `&'static str` payload forces every validation message to be a compile-time literal. Today there are only two literals so it works, but future checks (e.g. `"fifo_limit too small: got {got}, need >= {min}"`) would require changing the variant signature, breaking downstream pattern-matching consumers.

A `String` payload, or a typed `InvalidDefinition(InvalidDefinitionKind)` with `#[non_exhaustive]`, is more extensible. Worth deciding now since the variant has just been added — tests at internal_test.rs:1361 and 1370 already pattern-match on the literal string, which couples them to the exact wording.

crates/core/src/qbft/mod.rs
```diff
@@ -579,31 +604,52 @@ where
...
+/// if the callback blocks.
 fn compare<I, V, C>(
     ct: &CancellationToken,
     d: &Definition<I, V, C>,
     msg: &Msg<I, V, C>,
     input_value_source_ch: &mpmc::Receiver<C>,
     input_value_source: C,
     timer_chan: &mpmc::Receiver<time::Instant>,
-) -> Result<C>
+) -> (C, Result<()>)
```
varex83agent, line 653: **`compare` returning `(C, Result<()>)` is awkward**

The tuple-return forces the caller to destructure into two locals (`let (new_input_value_source, compare_result) = compare(...); input_value_source = new_input_value_source;` at mod.rs:501-509) where the `C` is always meaningful but `Result<()>` is the gating signal — an anti-pattern relative to a small typed return like:

```rust
struct CompareOutcome<C> { input_value_source: C, result: Result<()> }
```

Destructuring then becomes `outcome.input_value_source` / `outcome.result` and the contract that the input-value-source is *always* returned (regardless of result) is encoded in the type rather than convention.

varex83agent (commented):
Solid hardening pass on QBFT — the chain-split tests, fuzz cases, and Byzantine-input filters meaningfully tighten the implementation, and the `compare` refactor finally lets the round timer win over a blocked callback. A few items deserve attention before this fully matches Charon's robustness:

**Minor (worth addressing):**
1. `compare` callback exiting/panicking without sending on the err channel kills the QBFT instance (`UnexpectedCompareError` path). Charon Go quietly waits for the round timer instead — Pluto is strictly less resilient.
2. Each `UPON_JUSTIFIED_PRE_PREPARE` now spawns a detached OS thread. A callback that ignores the cancellation token leaks a thread per invocation; over hours this becomes a slow-burn DoS. (Documented as Charon parity, but OS threads aren't goroutines.)
3. `thread::sleep(Duration::from_millis(10))` is used as test synchronization in `compare_parent_cancel_cancels_callback_token` — replace with a `compare_started_rx` signal like the sister test already does.
4. `make_is_leader` reshapes Go's modulo so process 4 wins at round 4 where Charon would elect no leader. Hidden test-only parity gap that may mask bugs in chain-split scenarios.
5. The new `valid_round_change_prepared_round` Byzantine check rejects messages Charon currently accepts — coordinate with the Charon team to add the same check upstream, or log a warning when it fires.

Plus several nits (naming, a dead filter, missing docs, missing boundary tests). Verdict: `COMMENT` — none of these block merge, but the compare-callback robustness deserves a follow-up.

crates/core/src/qbft/mod.rs
```diff
@@ -1178,7 +1285,10 @@ where
     let null_pr = Default::default();
     let null_pv = Some(&Default::default());
 
-    let justification = filter_msgs(all, MSG_ROUND_CHANGE, round, None, Some(null_pr), null_pv);
+    let justification = filter_msgs(all, MSG_ROUND_CHANGE, round, None, Some(null_pr), null_pv)
+        .into_iter()
+        .filter(valid_round_change_prepared_round)
```
varex83agent: nit: the `.filter(valid_round_change_prepared_round)` chained here is dead code. `quorum_null_prepared` already filters via `filter_msgs` with `pr = 0`, so every surviving message has `pr == 0`. For `pr == 0` and `round >= 1` (the only round values this is called with), `valid_round_change_prepared_round` always evaluates true. Either remove the redundant filter or add a comment explaining it's defensive in case future callers pass `round == 0`.

crates/core/src/qbft/mod.rs
```diff
@@ -41,12 +54,21 @@ pub enum QbftError {
     #[error("Compare leader value with local value failed")]
     CompareError,
 
+    #[error("bug: expected only comparison or timeout error")]
+    UnexpectedCompareError,
```
varex83agent: nit: `UnexpectedCompareError` discards the actual underlying variant when fired from the `_` arm at mod.rs:531. If this ever triggers in production (it shouldn't, but the variant exists precisely for the unexpected case), debugging is harder than it needs to be. Consider `UnexpectedCompareError(Box<QbftError>)` with `#[error("bug: expected only comparison or timeout error, got {0}")]` so the original is preserved.

crates/core/src/qbft/fake_clock.rs
```diff
@@ -98,56 +98,86 @@ impl FakeClock {
     }
 }
```
varex83agent, line 183: nit: the previous `impl Drop for FakeClock` auto-cancelled timers when the clock went out of scope. Removing it is correct for the new test pattern (FakeClock is `Clone`; threads hold their own clones, so one clone being dropped must not cancel the others), but no comment explains the design. Add a one-line comment near `cancel` documenting the explicit-cleanup expectation so a future contributor doesn't re-introduce `Drop` — and consider a test-helper macro that wraps the `cts.cancel(); clock.cancel(); panic!(...)` boilerplate, since every panic path in `test_qbft` / `test_qbft_chain_split` must remember to cancel both before panicking.

crates/core/src/qbft/internal_test.rs
```diff
@@ -786,5 +1304,825 @@ fn duplicate_pre_prepare_rules() {
...
+    test_qbft_chain_split(ChainSplitTest {
+        decide_round: 2,
+        value_source: HashMap::from([(1, 3), (2, 1), (3, 1), (4, 1)]),
+        prepared_val: 1,
+        should_halt: false,
+    });
+}
+
+#[test]
+fn zz_chain_split_no_consensus_halt() {
```
varex83agent: nit: the `zz_` prefix on `zz_chain_split_no_consensus_halt` looks like an ordering hack but `cargo test` doesn't guarantee alphabetical execution under parallel scheduling. If the goal is to keep this slow halt scenario from blocking parallel runs of cheaper tests, use `#[ignore = "slow halt scenario"]` and add a CI step that runs it with `cargo test -- --ignored`. Otherwise drop the prefix to `chain_split_no_consensus_halt` for consistency with the other three chain_split tests.

crates/core/src/qbft/mod.rs
```diff
@@ -41,12 +54,21 @@ pub enum QbftError {
...
+
+    #[error("context canceled")]
+    ContextCanceled,
```
varex83agent, line 64: nit: `ContextCanceled` is Go-flavored (mirrors `context.Canceled`). Rust APIs typically use `Cancelled` (or `Canceled` to match the `cancellation` crate's `is_canceled()`). The `Context` prefix is also odd here since there's no `Context` type — the parameter is `ct: &CancellationToken`. Test names below already use `cancelled` (internal_test.rs:1311 `idle_run_returns_when_cancelled`, internal_test.rs:1719 `compare_cancelled_tx`), so the enum variant is inconsistent with its own callers.

crates/core/src/qbft/mod.rs
iamquang95: Paper based

crates/core/src/qbft/mod.rs
iamquang95: That's normal in my opinion.

therustmonk (approved):

crates/core/src/qbft/internal_test.rs
emlautarom1: @iamquang95 WDYT?

iamquang95: Fixed :+1:

crates/core/src/qbft/fake_clock.rs
```diff
@@ -11,41 +14,73 @@ pub struct FakeClock {
...
+    InputValue,
+}
+
 struct FakeClockInner {
     start: Instant,
     now: Instant,
     last_id: usize,
-    clients: HashMap<usize, (mpmc::Sender<Instant>, Instant)>,
+    cancelled: bool,
+    clients: BTreeMap<usize, (mpmc::Sender<Instant>, Instant, TimerPriority)>,
```
therustmonk, line 31: Consider adding a new ClientRecord struct instead of the tuple.

iamquang95: Will do that on the refactoring PR later 👍

iamquang95:
Let's merge this to have more CI and local test runs to see if there is any flanky tests

PR #418, opened 2026-05-15
feat(testutil/validatormock): port Charon validatormock to Rust

### LINKED ISSUE #38: Implement testutil/validatormock

varex83:
## Summary

Ports `charon/testutil/validatormock` (Go) to Rust under `crates/testutil/src/validatormock/`. Mirrors the 5 Go files file-per-concern with full functional parity:

- `meta.rs` — `SpecMeta`/`MetaSlot`/`MetaEpoch` slot/epoch arithmetic
- `sign.rs` — `Sign` trait + `Signer` + `SignFunc = Arc<dyn Sign>` (Go's `SignFunc` + `NewSigner`)
- `validators.rs` — local `ActiveValidators` newtype + helper (avoids a `pluto-testutil → pluto-app` dependency cycle)
- `capture.rs` — `SubmissionCapture` wiremock helper, the test-side replacement for Go's `beaconMock.SubmitAttestationsFunc = ...` callback fields
- `attest.rs` — `SlotAttester` with `Prepare`/`Attest`/`Aggregate`; matches Go goldens byte-for-byte via `assert_json_diff`
- `propose.rs` — `propose_block` + `register`; Bellatrix→Fulu, blinded + full variants
- `synccomm.rs` — `SyncCommMember` + `TestGetSubcommittees`
- `component.rs` — scheduler (`tokio::sync::mpsc` + `CancellationToken` + injectable `Clock`)
- `clock.rs` — `Clock` trait + `SystemClock` + explicit-advance `FakeClock`
- `error.rs` — unified `thiserror::Error`
- `testdata/TestAttest_{0,1}_{attestations,aggregations}.golden` — byte-for-byte from `charon/testutil/validatormock/testdata/`

Based on `bohdan/beaconmockapi` (#416). Will rebase onto `main` once #416 merges.

## Design notes

- Go's `close(chan struct{})` ready signals are modelled per-module: `attest.rs` uses `AtomicBool` + `tokio::sync::Notify`; `synccomm.rs` uses `Arc<tokio::sync::OnceCell<()>>` (lazy per-slot map).
- Go's goroutine scheduler maps to `tokio::spawn` + `mpsc` + `CancellationToken`.
- Go's `clockwork.FakeClock` maps to an injectable `Clock` trait. `tokio::time::pause()` was rejected — it interacts poorly with `wiremock::MockServer`.
- One pre-existing `pluto-eth2util ↔ pluto-testutil` Cargo cycle was fixed: `pluto-testutil` moved from `[dependencies]` to `[dev-dependencies]` in `crates/eth2util/Cargo.toml`. The 3 call sites in eth2util (signing, eth2exp, enr tests) are all `#[cfg(test)]`, so this is a no-op for production code.

## Workflow

Built across 4 phases on parallel git worktrees:
1. Foundation (sequential) — meta/sign/validators/capture/error/mod/testdata
2. Three parallel worktrees — attest / propose / synccomm
3. Component + clock (sequential)
4. Final verification

## Test plan

- [x] `cargo +nightly fmt --all --check` — clean
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings` — clean
- [x] `cargo test --workspace --all-features` — all green; **73 tests pass** in `pluto-testutil` (6 `#[ignore]`d propose tests waiting on missing `RandomCapella/Deneb/Electra/Fulu*Proposal` helpers in `pluto-testutil::random` — follow-up)
- [x] `cargo deny check` — clean
- [x] Goldens match Charon's `testdata/TestAttest_{0,1}_{attestations,aggregations}.golden` byte-for-byte (structural via `assert_json_diff`)

## Known follow-ups (out of scope)

- Add `RandomCapella/Deneb/Electra/Fulu{Versioned,Blinded}Proposal` helpers to `pluto-testutil::random` to un-ignore 6 propose tests.
- Add a default `POST /eth/v1/beacon/states/{state_id}/validators` mount to `BeaconMock` so callers don't have to inline it.
- `register` deviates from Charon's `switch signedRegistration.Version` (always zero-valued V1) and instead switches on the input registration's `Version`. Inline comment notes the Go quirk; please confirm intent.


crates/testutil/src/validatormock/component.rs
```diff
@@ -0,0 +1,565 @@
...
+            // empty — `Some(_)` does not match `None`.
+            Some(_) = duties.join_next() => {}
+        }
+    }
+    // Drain in-flight duties before returning so callers awaiting
+    // `Component::shutdown` see all work settled. Each task observes
+    // `cancel.cancelled()` in its outer select; bodies already running run
+    // to completion (mirroring Go's `Run`, which lets the duty finish
+    // before returning from the loop).
+    while duties.join_next().await.is_some() {}
```
iamquang95, line 327: if `prepare()` fails before closing readiness signals, later attest/aggregate/sync tasks wait forever; `shutdown()` cancels then drains `JoinSet`, so it can block forever

crates/testutil/src/validatormock/attest.rs
```diff
@@ -0,0 +1,986 @@
...
+// ---------------------------------------------------------------------------
+// Raw POST helpers
+// ---------------------------------------------------------------------------
+
+async fn submit_attestations(
+    eth2_cl: &EthBeaconNodeApiClient,
+    atts: &[VersionedAttestationJson],
+) -> Result<()> {
+    const ENDPOINT: &str = "/eth/v2/beacon/pool/attestations";
+    submit_json(eth2_cl, ENDPOINT, atts).await
```
iamquang95, line 629: In Charon this goes through the eth2 client `SubmitAttestations`, so it return the real submit error. This is only json from go golder file, which may not match the actual beacon API shape

crates/testutil/src/validatormock/attest.rs
```diff
@@ -0,0 +1,986 @@
...
+async fn submit_aggregate_attestations(
+    eth2_cl: &EthBeaconNodeApiClient,
+    aggs: &[VersionedSignedAggregateAndProofJson],
+) -> Result<()> {
+    const ENDPOINT: &str = "/eth/v2/validator/aggregate_and_proofs";
+    let body = SubmitAggregateAttestationsOptsJson {
+        common: CommonOpts::default(),
+        signed_aggregate_and_proofs: aggs,
+    };
+    submit_json(eth2_cl, ENDPOINT, &body).await
```
iamquang95, line 641: same as `submit_attestations`

iamquang95 (approved):
Other parts LGTM

crates/testutil/src/validatormock/component.rs
varex83agent: Good catch — fixed in 7a4ced6.

`CloseOnce::wait()` is not cancellable (unlike Go's `wait(ctx, ch)` which selects on `ctx.Done()`), so a duty whose readiness signal was never closed would park forever, and `shutdown()`'s `JoinSet` drain would block indefinitely. The fix wraps `run_duty_via_inner` in an inner `tokio::select!` against `cancel_for_task.cancelled()`, mirroring Go's behavior — `shutdown()` now cancels the token and parked duty bodies wake up immediately.

Also added a regression test (`shutdown_terminates_when_duty_blocked_on_close_once`) that uses `FakeClock` to schedule an `Attester` duty in a non-epoch-boundary slot (so `PrepareAggregator` doesn't fire first), advances time past the attester offset to make the duty actually start running and park on `duties_ok.wait()`, then asserts `shutdown()` completes within 2s. Without the fix it times out.

crates/testutil/src/validatormock/attest.rs
varex83agent: Addressed in 7a4ced6. We can't easily switch to the typed `submit_pool_attestations_v2` because it encodes the Electra+ `SingleAttestation` shape, which doesn't match Charon's `*eth2spec.VersionedAttestation` wire format that the goldens capture. As a middle ground I:

1. Made `submit_json` read the response body on non-success and surface it via a new `Error::SubmitStatus { endpoint, status, body }` variant. Beacon-node validation errors (4xx/5xx with a JSON `{"code":..,"message":..}`) now reach the caller — matching the diagnostic richness of Go's typed `eth2Cl.SubmitAttestations` error rather than the previous "just a transport-level reqwest::Error".
2. Expanded the module-level wire-shape note to flag explicitly that the Go-shaped payload has *not* been validated end-to-end against a real beacon node, and what the follow-up paths look like (teach the typed client about `VersionedAttestation`, or switch to `SingleAttestation` and regenerate the goldens).

Happy to file a follow-up issue if you'd rather track the typed-client migration there.

crates/testutil/src/validatormock/attest.rs
varex83agent: Same fix covers this one — `submit_aggregate_attestations` goes through the same `submit_json` helper, so it now also surfaces the beacon-node response body via `Error::SubmitStatus`. See the reply on the `submit_attestations` thread for the full rationale.

PR #419, opened 2026-05-18
feat(relay): wire loki tracing init and graceful shutdown

### LINKED ISSUE #413: Deploy relay server to production
## Background

The `relay-server` crate (`crates/relay-server/`) provides a libp2p relay used by Pluto nodes for peer discovery and NAT traversal. We need to stand up a deployed instance so cluster operators can rely on it for bootstrapping P2P connectivity.

## Tasks

- [ ] Define hosting target (cloud provider / region / DNS name)
- [ ] Provision infrastructure (compute, networking, TLS as needed)
- [ ] Build and publish a release artifact / container image for `relay-server`
- [ ] Author deployment config (env vars, listen addresses, keys)
- [ ] Deploy the relay server to the chosen environment
- [ ] Wire the deployed multiaddr into default Pluto config / docs
- [ ] Set up basic observability (logs, metrics, health check)
- [ ] Document the deploy + redeploy procedure

## Acceptance Criteria

- A relay server is reachable at a stable public multiaddr.
- Pluto nodes can connect through it (verified end-to-end against a test cluster).
- Deploy/redeploy steps are documented in the repo or ops runbook.
- Basic monitoring is in place so failures are observable.

### LINKED ISSUE #420: Add Loki and tracing initialization for relay server

varex83:
## Summary

- Build the relay's `TracingConfig` directly in `RelayArgs::try_into` so the existing `--log-color`, `--loki-addresses`, and `--loki-service` flags actually flow into the tracing layer (previously parsed but ignored — `common.rs` had a `TODO: Handle loki config`).
- Move `pluto_tracing::init` into `relay::run` and spawn the returned `BackgroundTask` so the `tracing-loki` worker actually ships records (the previous `main.rs` call discarded it).
- Add a 3-second bounded flush window on shutdown so buffered Loki batches drain before the runtime tears the worker down on SIGTERM.
- Tolerate `TryInitError` in `relay::run` so multiple in-process integration tests can share an already-installed global subscriber instead of panicking on re-init.

## Test plan

- [ ] `cargo clippy -p pluto-cli --all-targets --all-features -- -D warnings` (passes locally)
- [ ] `cargo test -p pluto-cli --all-features` — confirm the relay server tests still pass with the new init path
- [ ] Manual: run `pluto relay` with `PLUTO_LOKI_ADDRESSES=http://localhost:3100` against a local Loki, verify logs arrive
- [ ] Manual: SIGTERM mid-run, confirm the last few seconds of logs reach Loki and shutdown completes within ~3s


crates/cli/src/commands/relay.rs
```diff
@@ -60,7 +60,28 @@ impl TryInto<pluto_relay_server::config::Config> for RelayArgs {
...
+                ConsoleColor::Force => builder.console_with_ansi(true),
+                ConsoleColor::Disable => builder.console_with_ansi(false),
+            };
+
+            if let Some(loki_url) = self.loki.loki_addresses.first() {
+                let mut labels = HashMap::new();
+                labels.insert("service".to_string(), self.loki.loki_service.clone());
+
+                builder = builder.loki(pluto_tracing::LokiConfig {
+                    loki_url: loki_url.clone(),
```
varex83agent, line 86: **Loki URL with embedded credentials now leaks through `info!(config = ?config)` at the top of `run`.**

`LokiConfig.loki_url` is a plain `String` and derives `Debug`. Operators commonly paste Grafana Cloud push URLs of the form `https://user:password@logs-prod-xxx.grafana.net/loki/api/v1/push` — Charon's own test fixtures use this shape (`cmd/cmd_internal_test.go`). The line at relay.rs:302 (`info!(config = ?config)`) was harmless before; this PR is what newly populates the field.

The leak has two surfaces:
1. Console / on-disk / journalctl log capture exposes the basic-auth password.
2. By the time line 302 fires, the Loki `tracing_loki::Layer` is already in the global subscriber chain (init happened at line 289), so the very log record carrying the credentials is serialised into a Loki push request and shipped to that same endpoint using those same credentials — the password is exfiltrated through its own auth round-trip and now lives in Grafana, visible to anyone with read access to that workspace.

Mitigations:
- Parse the URL up-front, strip userinfo before storing in `LokiConfig.loki_url`, and pass credentials separately to the HTTP client (cleanest).
- Or: implement a custom `Debug` for `LokiConfig` that redacts the URL's userinfo segment.
- Add a regression test that asserts `format!("{:?}", config)` does not contain `password` for a `LokiConfig` built from `https://u:password@host/...`.

crates/cli/src/commands/relay.rs
```diff
@@ -265,6 +286,18 @@ pub async fn run(
     config: pluto_relay_server::config::Config,
     ct: CancellationToken,
 ) -> Result<(), CliError> {
+    let loki_task = match pluto_tracing::init(&config.log_config) {
+        Ok(Some(task)) => Some(tokio::spawn(task)),
+        Ok(None) => None,
+        Err(pluto_tracing::init::Error::InitError(_)) => {
```
varex83agent: **Swallowing `TryInitError` silently disables all of this PR's behaviour in production.**

The comment frames this as a test-isolation accommodation, but the same arm fires in the real binary if any other component (a panic handler, a build-time `ctor`, an inadvertently-pulled-in logger crate) won the install race. In that case:
- The relay continues without warning.
- All flags wired by this PR — `--log-level`, `--log-color`, `--loki-addresses`, `--loki-service` — are silently discarded.
- `loki_task` returns `None`, so the operator's configured Loki shipping is silently off.
- The credential-bearing `info!(config = ?config)` at line 302 still emits to whatever subscriber was already installed — an even worse exfiltration vector than the in-our-own-Loki case.

Recommend one of:
1. Gate the swallow behind `#[cfg(test)]` so the production binary fails fast.
2. Emit a `tracing::warn!` (or `eprintln!`, since we don't own the subscriber) noting that init was bypassed, and which config fields were discarded.
3. Use a `OnceCell` in the test harness to install the subscriber exactly once, and keep production strict.

Charon's `cmd/relay.go:24-26` propagates the equivalent error unconditionally.

crates/cli/src/commands/relay.rs
```diff
@@ -60,7 +60,28 @@ impl TryInto<pluto_relay_server::config::Config> for RelayArgs {
...
+        let log_config = {
+            let mut builder = pluto_tracing::TracingConfig::builder().with_default_console();
+
+            builder = match &self.log.color {
+                ConsoleColor::Auto => builder.console_with_ansi(std::env::var("NO_COLOR").is_err()),
+                ConsoleColor::Force => builder.console_with_ansi(true),
+                ConsoleColor::Disable => builder.console_with_ansi(false),
+            };
+
+            if let Some(loki_url) = self.loki.loki_addresses.first() {
```
varex83agent: **`--loki-addresses` silently drops all but the first entry.**

`RelayLokiArgs.loki_addresses` is declared `Vec<String>` with `value_delimiter = ','` (line 269-274) and the help text says "these Loki log aggregation server addresses" (plural). The new code uses only `.first()` and discards the rest with no warning.

This is both a parity gap (Charon `app/log/config.go:227-242` iterates the slice and tee's logs to every endpoint via a multi-sink logger) and a configuration-validation gap (an operator listing a primary + failover endpoint sees their failover dropped on the floor).

Options, ranked:
1. **Honour the contract**: extend `pluto_tracing::LokiConfig` / `pluto_tracing::init` to accept multiple URLs (one `tracing_loki::Layer` per URL stacked in the registry). Matches Charon and keeps the CLI surface honest.
2. **Shrink the contract**: change the flag to `--loki-address: Option<String>`, drop the `value_delimiter`, and update help text + env var name. Easy and explicit.

What we have today — accept N, use 1 — is the worst combination: looks like it works, silently doesn't.

crates/cli/src/commands/relay.rs
```diff
@@ -60,7 +60,28 @@ impl TryInto<pluto_relay_server::config::Config> for RelayArgs {
             }
         };
 
-        let log_config = build_console_tracing_config(self.log.level.clone(), &self.log.color);
+        let log_config = {
```
varex83agent: Lines 63-84 are a copy of `commands::common::build_console_tracing_config` (common.rs:27-44) with the Loki block appended — the existing `// TODO: Handle loki config` (common.rs:39) was the natural extension point. Today `dkg.rs:127` still calls the helper, so any future change to console handling (the still-dead `--log-format` and `--log-output-path` flags being the obvious next two) needs to be applied in two places. Recommend extending the shared helper to take an optional `LokiConfig` (or accept the `RelayLokiArgs` directly) so both call sites stay synchronised.

crates/cli/src/commands/relay.rs
```diff
@@ -291,10 +324,21 @@ pub async fn run(
...
         .map(|_| ())
-        .map_err(Into::into)
+        .map_err(Into::into);
+
+    // Give the Loki worker a short window to flush buffered logs before the
+    // runtime aborts it. The task holds a `Layer` clone through the global
+    // subscriber and never completes on its own, so the wait is bounded.
+    if let Some(handle) = loki_task {
+        let abort_handle = handle.abort_handle();
+        let _ = tokio::time::timeout(std::time::Duration::from_secs(3), handle).await;
```
varex83agent: **The 3-second "flush window" is actually a fixed wait followed by an abort, not a graceful shutdown.**

`tracing_loki::BackgroundTask` only resolves when the channel sender drops AND there is no in-flight `send_task` (`tracing-loki-0.2.6/src/lib.rs`). The sender lives inside the `Layer` clone held by the global `Registry`, which lives for the entire process. So `tokio::time::timeout(3s, handle)` always returns `Err(Elapsed)`, and `abort_handle.abort()` then force-cancels the task in whatever state it's in — including mid-HTTP-POST, which drops the in-flight batch.

`tracing_loki` exposes the right API for this case: `Builder::build_controller_url` returns a `(Layer, BackgroundTaskController, BackgroundTask)` triple, and `BackgroundTaskController::shutdown().await` sends the sentinel that flips the task into draining mode — it then finishes any in-flight request and resolves naturally. Plumbing the controller through `pluto_tracing::init` (return both controller and task) would let the relay do:

```rust
if let Some(controller) = loki_controller {
    controller.shutdown().await;
    let _ = tokio::time::timeout(Duration::from_secs(3), loki_task).await;
    abort_handle.abort(); // last-resort fallback
}
```

As written, the bounded wait helps a previously-buffered request finish, but anything queued after the layer stops receiving events is lost because the task never re-enters its `prepare_sending` loop.

crates/cli/src/commands/relay.rs
```diff
@@ -291,10 +324,21 @@ pub async fn run(
         e => e,
     }?;
 
-    pluto_relay_server::p2p::run_relay_p2p_node(&config, key, ct)
+    let result = pluto_relay_server::p2p::run_relay_p2p_node(&config, key, ct)
```
varex83agent: Early `?` exits before this point (key load at line 325, the `auto_p2p_key=false` branch at line 316) drop `loki_task: Option<JoinHandle<()>>` without aborting and without traversing the flush block at lines 335-339. Dropping a `JoinHandle` does NOT cancel the underlying task — the spawned `BackgroundTask` keeps running on the tokio runtime until the process exits. For the CLI binary this is benign (the process is about to die). For `with_relay_server` integration tests that share a tokio runtime, every early-return test would leak a `BackgroundTask` for the lifetime of the test binary if Loki were configured. Today no test sets `loki_addresses` so this is latent, but it's a structural foot-gun: wrap the body in a helper and abort `loki_task` via a `scopeguard`/RAII drop guard, or restructure as `let result = (async { … })().await;` so the flush block always runs.

crates/cli/src/commands/relay.rs
```diff
@@ -265,6 +286,18 @@ pub async fn run(
     config: pluto_relay_server::config::Config,
     ct: CancellationToken,
 ) -> Result<(), CliError> {
+    let loki_task = match pluto_tracing::init(&config.log_config) {
```
varex83agent: Init order is: install global subscriber (and spawn the Loki worker) here, then load/validate the P2P key at lines 304-325. If the user provides `--loki-addresses=https://loki.invalid` AND has a missing/corrupt P2P key with `--auto-p2pkey=false`, the Loki worker gets spun up and starts retry-backoff for an unreachable host, then the relay errors out on the key-load step and the spawned task is leaked (see the line 327 finding). Reversing the order — validate the key first, then install the subscriber — would avoid the spurious worker and the leak. Cost is losing the `info!("{LICENSE}")` banner on key-failure paths. Low-priority polish.

crates/cli/src/commands/relay.rs
```diff
@@ -60,7 +60,28 @@ impl TryInto<pluto_relay_server::config::Config> for RelayArgs {
...
+                labels.insert("service".to_string(), self.loki.loki_service.clone());
+
+                builder = builder.loki(pluto_tracing::LokiConfig {
+                    loki_url: loki_url.clone(),
+                    labels,
+                    extra_fields: HashMap::new(),
+                });
+            }
+
+            builder.override_env_filter(self.log.level.clone()).build()
```
varex83agent: Two CLI flags adjacent to the ones this PR newly honors are still dead: `--log-format` (`RelayLogFlags::format` at line 245, accepts `console`/`logfmt`/`json`) and `--log-output-path` (line 263). Neither is threaded into `TracingConfig` here — the `// TODO: Handle log output path` in common.rs:41 is still the canonical pointer. PR description says "wire loki tracing init" but a reasonable reader will assume the rest of the log flags now work; consider either landing both (`tracing_appender::rolling::Builder` covers `--log-output-path`; `--log-format` needs a small enum on `ConsoleConfig`) or amending the help text on those flags to note they're not yet wired.

crates/cli/src/commands/relay.rs
```diff
@@ -265,6 +286,18 @@ pub async fn run(
     config: pluto_relay_server::config::Config,
     ct: CancellationToken,
 ) -> Result<(), CliError> {
+    let loki_task = match pluto_tracing::init(&config.log_config) {
+        Ok(Some(task)) => Some(tokio::spawn(task)),
```
varex83agent: No test exercises any of the Loki-shipping behaviour this PR introduces. `with_relay_server` hardcodes `loki_addresses: vec![]` (line 559), so every existing test takes the `Ok(None)` branch and the spawn / flush block never runs. The newly-introduced code paths — `LokiConfig` construction in `try_into`, `BackgroundTask` spawn, the 3s flush window, the `InitError` swallow — are entirely untested.

At minimum: a unit test on `RelayArgs::try_into` asserting that a non-empty `loki_addresses` produces a `TracingConfig` with `loki.loki_url == addresses[0]` and `loki.labels["service"] == loki_service`. A second test pointing at a local `httptest::Server` (or `wiremock`) exercising the spawn-and-flush path would lock in the shutdown timing and catch regressions to the abort_handle dance.

crates/cli/src/commands/relay.rs
```diff
@@ -60,7 +60,28 @@ impl TryInto<pluto_relay_server::config::Config> for RelayArgs {
...
+            let mut builder = pluto_tracing::TracingConfig::builder().with_default_console();
+
+            builder = match &self.log.color {
+                ConsoleColor::Auto => builder.console_with_ansi(std::env::var("NO_COLOR").is_err()),
+                ConsoleColor::Force => builder.console_with_ansi(true),
+                ConsoleColor::Disable => builder.console_with_ansi(false),
+            };
+
+            if let Some(loki_url) = self.loki.loki_addresses.first() {
+                let mut labels = HashMap::new();
```
varex83agent: nit: Single-entry map can be built in one expression:

```rust
let labels = HashMap::from([("service".to_string(), self.loki.loki_service.clone())]);
```

Drops the `mut` and one line.

crates/cli/src/commands/relay.rs
```diff
@@ -60,7 +60,28 @@ impl TryInto<pluto_relay_server::config::Config> for RelayArgs {
...
+                labels.insert("service".to_string(), self.loki.loki_service.clone());
```
varex83agent: nit: `try_into` takes `self` by value, so `self.loki.loki_service` and `self.loki.loki_addresses` can be moved instead of cloned. Pull `let RelayLokiArgs { loki_addresses, loki_service } = self.loki;` and `loki_addresses.into_iter().next()` for the URL. Same for `self.log.level.clone()` at line 83. Saves three string allocations on startup.

crates/cli/src/commands/relay.rs
varex83agent: nit: The four-arm match can be reduced to three:

```rust
let loki_task = match pluto_tracing::init(&config.log_config) {
    Ok(maybe_task) => maybe_task.map(tokio::spawn),
    Err(pluto_tracing::init::Error::InitError(_)) => None,
    Err(err) => return Err(err.into()),
};
```

`Option::map(tokio::spawn)` collapses the `Ok(Some)` / `Ok(None)` pair. Also consider `use pluto_tracing::init::Error as TracingInitError;` at the top so the verbose path on line 292 wraps less awkwardly.

crates/cli/src/commands/relay.rs
```diff
@@ -291,10 +324,21 @@ pub async fn run(
...
         .await
         .map(|_| ())
-        .map_err(Into::into)
+        .map_err(Into::into);
+
+    // Give the Loki worker a short window to flush buffered logs before the
+    // runtime aborts it. The task holds a `Layer` clone through the global
+    // subscriber and never completes on its own, so the wait is bounded.
+    if let Some(handle) = loki_task {
+        let abort_handle = handle.abort_handle();
```
varex83agent: nit: `tokio::time::timeout(d, handle)` consumes the `JoinHandle`, which is why `abort_handle` has to be captured first. The pattern is correct but non-obvious — a one-liner comment would help:

```rust
// `timeout` consumes the JoinHandle; grab an abort handle first so we can
// still cancel the task after the timer elapses (dropping a JoinHandle does
// not abort).
let abort_handle = handle.abort_handle();
```

crates/cli/src/commands/relay.rs
varex83agent: nit: `Duration::from_secs(3)` is a magic number sitting inline. Lift to a `const LOKI_FLUSH_TIMEOUT: Duration = Duration::from_secs(3);` near the top of the module (or — better — make it part of `LokiConfig` so operators can tune it). Either way greppable beats inline.

varex83agent (dismissed):
Solid wiring of the previously-unused `--loki-addresses` / `--loki-service` flags into the relay's `TracingConfig`, plus a clean enough shutdown sequence to flush buffered batches before the runtime tears down. Three findings push to request-changes:

1. **`info!(config = ?config)` now leaks Loki URL credentials** — operators routinely embed basic-auth in the Loki push URL, and this PR is what newly populates that field, so the previously-harmless log line is now the credential-exfiltration vector.
2. **`Err(InitError(_)) => None` silently disables Loki in production** — same code path serves the test-isolation case and the production case where any other crate already installed the global subscriber. Operators see no warning that their flags were ignored.
3. **Only the first `--loki-addresses` entry is used** — the flag is `Vec<String>` with plural help text, but `.first()` drops the rest. Charon supports multi-endpoint Loki fan-out. Either honour the contract or shrink the CLI surface.

Minor: duplicates `build_console_tracing_config` instead of extending it; the 3-second flush is an `abort` rather than a graceful shutdown (use `tracing_loki::BackgroundTaskController::shutdown`); `--log-format` / `--log-output-path` flags are still dead; no test exercises the Loki path.

Nothing here is unsafe-Rust or correctness-of-protocol — these are operability / secrets-handling concerns. Verdict: REQUEST_CHANGES on the three majors; the rest are minor/nit.

therustmonk (approved):

crates/tracing/src/init.rs
```diff
@@ -49,22 +65,98 @@ pub fn init(config: &TracingConfig) -> Result<Option<BackgroundTask>> {
...
         Ok(None)
     }
 }
 
+fn extract_basic_auth(url: &Url) -> Option<String> {
+    if url.username().is_empty() && url.password().is_none() {
+        return None;
+    }
+    let user = url.username();
+    let pass = url.password().unwrap_or("");
```
iamquang95: Do we need to handle `%40, %2F, %3A` ?

PR #423, opened 2026-05-19
refactor: improve the deadline module

therustmonk:


crates/core/src/deadline/calculator.rs
```diff
@@ -0,0 +1,102 @@
...
+        let slot_duration = to_chrono_duration(slot_duration)?;
+        Ok(Self {
+            genesis_time,
+            slot_duration,
+        })
+    }
+
+    /// Wall-clock start of the given slot: `genesis_time + slot *
+    /// slot_duration`.
+    fn slot_start(&self, slot: SlotNumber) -> Result<DateTime<Utc>> {
```
mskrzypkows, line 49: would be good to have a unit test for it

crates/core/src/deadline/calculator.rs
```diff
@@ -0,0 +1,102 @@
...
+        let offset = Millis::from(self.slot_duration).checked_mul_slot(slot)?;
+        offset.add_to(self.genesis_time)
+    }
+
+    /// Network-delay margin added to every deadline: `slot_duration /
+    /// MARGIN_FACTOR`.
+    fn margin(&self) -> Result<Millis> {
```
mskrzypkows: ditto

crates/core/src/deadline/mod.rs
```diff
@@ -0,0 +1,625 @@
...
+        use mpsc::error::TrySendError::*;
+        let duty = self.curr_duty.clone();
+        match self.output_tx.try_send(duty) {
+            Ok(()) => {}
+            Err(Full(curr_duty)) => {
+                tracing::warn!(
+                    label = %self.label,
+                    duty = %curr_duty,
+                    "Deadliner output channel full"
+                );
```
mskrzypkows, line 348: Maybe return some small sleep time here to retry in a while?

mskrzypkows (approved):

varex83 (approved):

therustmonk:
@claude review the changes

crates/core/src/deadline/calculator.rs
```diff
@@ -0,0 +1,191 @@
...
+
+use chrono::{DateTime, Duration, Utc};
+use pluto_eth2api::EthBeaconNodeApiClient;
+
+use crate::types::{Duty, DutyType, SlotNumber};
+
+use super::{Result, millis::Millis, to_chrono_duration};
+
+/// Fraction of slot duration to use as a margin for network delays.
+const MARGIN_FACTOR: i32 = 12;
```
iamquang95: This can be `i64` from the begin, no casting later

crates/core/src/deadline/calculator.rs
```diff
@@ -0,0 +1,191 @@
...
+
+    /// Network-delay margin added to every deadline: `slot_duration /
+    /// MARGIN_FACTOR`.
+    fn margin(&self) -> Result<Millis> {
+        Millis::from(self.slot_duration).checked_div(MARGIN_FACTOR.into())
+    }
+
+    /// Duty-type-specific offset from slot start.
+    fn duty_duration(&self, duty_type: &DutyType) -> Result<Millis> {
+        let secs = Millis::from(self.slot_duration);
```
iamquang95: `msecs` is better name i think, it's `Millis`

iamquang95 (approved):

crates/core/src/deadline/mod.rs
therustmonk: We left it like this on purpose to match the reference behavior, but I agree that can be the better choice here.

emlautarom1:
@therustmonk Could you check #430? It makes use of the deadliner and shows some challenges with the existing API that could be taken care of in this PR.

crates/core/src/deadline/mod.rs
```diff
@@ -0,0 +1,682 @@
...
+    /// - The calculator reports the duty has no deadline (`Ok(None)`)
+    /// - The calculator failed to compute the deadline (`Err(_)`)
+    async fn add(&self, duty: Duty) -> bool;
+
+    /// Returns the channel for receiving deadlined duties.
+    ///
+    /// This method may only be called once and returns `None` on subsequent
+    /// calls. The returned channel should only be used by a single task.
+    fn c(&self) -> Option<mpsc::Receiver<Duty>>;
+}
```
emlautarom1, line 113: The `Deadliner` interface is hard to use as is:

- It's not intended to be cloned, so `Arc<dyn Deadliner>` is unnecessary, that is, a `Deadliner` instance is never shared by multiple "services" (ex. `aggsigdb`), instead it's owned.
- In a service, there might be multiple places where `Duty` are added
- The channel `c` is used only once by a service and it's always expected to be `Some`.

If you squint a bit you can see how a `Deadliner` is a `mpsc` channel, where there are multiple places where a `Duty` can be registered, while there is a single place where the deadlined duties are read. Also, consider that a `Deadliner` behaves like an Actor: https://rust-lang-nursery.github.io/rust-cookbook/concurrency/actor.html

I roughly suggest:

- Make the construction of a deadliner return `(handle, c)`, where `c` is the current `Deadliner.c()`, a background task is listening for messages, and `handle` allows submitting `Duty` to the background task (through `mpsc` channel, thus we support multiple handles by cloning)
- The parameters to the constructor remain the same (ct, label and calculator), they're good.
- Submitting through a handle waits for a response as we're doing now (through `oneshot`), but the response is more explicit, separating the "successfully scheduled, "already expired", "no deadline", "failed to compute deadline".
- The background task emits values on `c` when the deadline is reached.

crates/core/src/deadline/calculator.rs
```diff
@@ -0,0 +1,190 @@
...
+/// Computes deadlines for duties.
+///
+/// `Ok(Some(deadline))` — duty expires at the given wall-clock time.
+/// `Ok(None)`           — duty never expires (e.g. Exit, BuilderRegistration).
+/// `Err(_)`             — arithmetic or conversion failure.
+pub trait DeadlineCalculator: Send + Sync + 'static {
+    /// Computes the deadline for the given duty. See trait docs for return
+    /// semantics.
+    fn deadline(&self, duty: &Duty) -> Result<Option<DateTime<Utc>>>;
+}
```
emlautarom1, line 87: Looks like this interface is the important one for testing. I would explore if we can make `Deadliner` a concrete type and control the logic through the calculator instead. If we can do that we can simplify quite the usages (no need to `Box<dyn Deadliner>`, etc.)

emlautarom1 (changes requested):
This refactor looks promising but it does not improve usages in a meaningful way.

I've left a comment on how I would address it based on the needs of modules like `aggsigdb` (#430), but I suggest to also explore how we can make existing modules better to use (ex. `parsigdb`).

Lastly, regarding the internals, some time ago I explored a different approach that uses tokio's scheduling for most of the work: https://github.com/NethermindEth/pluto/tree/emlautarom1/deadline_alt . Maybe something is worth salvaging from that experiment, but the main goal should now be to make it easier for callers to use this module.

crates/core/src/deadline/mod.rs
therustmonk: @emlautarom1 Yes, I thought about what we can do here, and overall I came to similar conclusions, fully agree with you.

The only thing is the `(handle, c)` pair, I think it's a good design. However, in that case we would also need to pass `c` around, OR switch to a stricter `c()` that returns `Result`, but an explicit `c` looks better to me.

Here's what I was thinking. Maybe we should merge the refactoring first, since it's the most compatible with the current changes, but already gives us flexibility for future improvements.

Then I can make a follow-up PR for the `(handle, c)` pair, since it looks like everyone will need to rebase for that change anyway, and it will be easier if the PR is small.

So basically:

1. Smoothly migrate to this PR, since it's well compatible with the current implementation.
2. Add a series of PRs (or one PR) that introduce more radical changes.

What do you think? Or do you still think it's better to include the API changes here as well?

emlautarom1: Passing the explicit `c` around is better than the `c()`, too many issues on callers otherwise.

No issues on having it as a stacked PRs (as long as we actually look into it and does not get forgotten :face_with_diagonal_mouth:)

therustmonk: Okay, I’ll open a separate (follow-up) PR based on this one, but with the updates, so you can use them and the team can rebase their branches on top of the changes from this PR.

therustmonk: The next PR with API changes (in progress): https://github.com/NethermindEth/pluto/pull/434

emlautarom1 (approved):
Internal changes LGTM, waiting for a follow-up PR to address this module's API.

varex83 (approved):

therustmonk:
@emlautarom1 The next PR with API changes: https://github.com/NethermindEth/pluto/pull/434

PR #424, opened 2026-05-19
fix(relay-server): pass dummy port to lookup_host

varex83:
## Summary
- `tokio::net::lookup_host` requires a `host:port` input, so passing a bare hostname (e.g. `pluto-relay-0-p2p.ovh.dev-nethermind.xyz`) failed with `invalid socket address` and the relay never resolved the configured `external_host` for its ENR.
- Pass a dummy port of `0` so the bare hostname resolves, and iterate the resolved addresses to find the first IPv4 instead of silently giving up if the first record happens to be IPv6.

## Test plan
- [ ] Deploy a relay with `external_host` set to a hostname and verify the warning `Failed to resolve external host ... invalid socket address` no longer appears
- [ ] Confirm the relay's `/enr` response contains the resolved IPv4 in the ENR `ip` field
- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings`

### DIFF
--- a/crates/cli/src/commands/relay.rs
+++ b/crates/cli/src/commands/relay.rs
@@ -406,23 +406,23 @@ mod tests {
                 // Resolution happens asynchronously on a tick, so poll until the
                 // ENR reflects a non-loopback IP (mirrors the Go test using
                 // `assert.Eventually`).
-                let deadline = time::Instant::now() + time::Duration::from_secs(10);
-                loop {
-                    let response = relay_server_get(cfg.clone(), "/enr").await.unwrap();
-                    let body = response.text().await.unwrap();
-                    let enr = pluto_eth2util::enr::Record::try_from(body.as_str()).unwrap();
-                    let ip = enr.ip().unwrap();
+                tokio::time::timeout(time::Duration::from_secs(10), async {
+                    loop {
+                        let response = relay_server_get(cfg.clone(), "/enr").await.unwrap();
+                        let body = response.text().await.unwrap();
+                        let enr =
+                            pluto_eth2util::enr::Record::try_from(body.as_str()).unwrap();
+                        let ip = enr.ip().unwrap();
 
-                    if !ip.is_loopback() {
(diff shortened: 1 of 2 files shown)


crates/cli/src/commands/relay.rs
```diff
@@ -403,11 +403,26 @@ mod tests {
...
             async |cfg| {
-                let response = relay_server_get(cfg, "/enr").await.unwrap();
-                let body = response.text().await.unwrap();
-                let enr = pluto_eth2util::enr::Record::try_from(body.as_str()).unwrap();
-
-                assert!(enr.ip().unwrap().is_loopback());
+                // Resolution happens asynchronously on a tick, so poll until the
+                // ENR reflects a non-loopback IP (mirrors the Go test using
+                // `assert.Eventually`).
+                let deadline = time::Instant::now() + time::Duration::from_secs(10);
```
emlautarom1: nit: Could use `tokio::time::timeout` to make it clearer (test should take at most 10 seconds)

emlautarom1 (approved):
LGTM

crates/relay-server/src/web.rs
```diff
@@ -329,13 +329,22 @@ async fn resolve_external_host_periodically(
...
+                .filter_map(|a| match a.ip() {
+                    IpAddr::V4(v4) => Some(v4),
+                    IpAddr::V6(_) => None,
+                })
+                .next();
+
+            if let Some(ipv4) = ipv4 {
                 debug!("Resolved external host {external_host} to {ipv4}");
                 state.set_external_host_ip(Some(ipv4)).await;
+            } else {
```
iamquang95, line 346: Should we invalidate the cache here `state.set_external_host_ip(None).await;`? The DNS may no longer return IPv4 for the host, so a later query could fail

varex83: I would keep it as it is, since it's the original charon behavior

PR #426, opened 2026-05-20
feat: sigagg

mskrzypkows:


mskrzypkows:
@claude review

crates/core/src/sigagg.rs
```diff
@@ -0,0 +1,847 @@
...
+        /// The out-of-range index value.
+        idx: u64,
+    },
+}
+
+/// Convenience alias for [`std::result::Result`] with [`SigAggError`].
+pub type Result<T> = std::result::Result<T, SigAggError>;
+
+/// Per-duty output: one aggregated [`SignedData`] per validator public key.
+pub type AggSignedDataSet = HashMap<PubKey, Box<dyn SignedData>>;
```
varex83, line 75: This is the same type introduced in https://github.com/NethermindEth/pluto/pull/430, I think after merging those two PRs we can make a shared type for this

crates/core/src/sigagg.rs
```diff
@@ -0,0 +1,847 @@
...
+        + 'static,
+>;
+
+/// Verify callback — checks the aggregated signature against the beacon chain.
+pub type VerifyFn = Arc<
+    dyn Fn(&PubKey, &dyn SignedData) -> Pin<Box<dyn Future<Output = Result<()>> + Send>>
+        + Send
+        + Sync
+        + 'static,
+>;
```
varex83, line 91: I think it would be great to replace this with `async_trait`

crates/core/src/sigagg.rs
```diff
@@ -0,0 +1,847 @@
...
+                        source: e,
+                    })?;
+            bls_sigs.insert(par_sig.share_idx, sig);
+        }
+
+        if (bls_sigs.len() as u64) < self.threshold {
+            return Err(SigAggError::InsufficientDistinctSignatures { pubkey: *pubkey });
+        }
+
+        let bls_map: HashMap<u8, [u8; 96]> = bls_sigs
```
varex83:
```suggestion
        let bls_map: HashMap<u8, Signature> = bls_sigs
```

crates/core/src/sigagg.rs
```diff
@@ -0,0 +1,847 @@
...
+            .map(|(idx, sig)| {
+                let idx_u8 = u8::try_from(*idx).map_err(|_| SigAggError::InvalidShareIndex {
+                    pubkey: *pubkey,
+                    idx: *idx,
+                })?;
+                Ok((idx_u8, *sig.as_ref()))
+            })
+            .collect::<Result<_>>()?;
+
+        let agg_bytes = info_span!("tbls.ThresholdAggregate")
```
varex83:
```suggestion
        let agg_bytes = info_span!("BlstImpl::threshold_aggregate")
```

crates/core/src/sigagg.rs
```diff
@@ -0,0 +1,847 @@
...
+        if (bls_sigs.len() as u64) < self.threshold {
+            return Err(SigAggError::InsufficientDistinctSignatures { pubkey: *pubkey });
+        }
+
+        let bls_map: HashMap<u8, [u8; 96]> = bls_sigs
+            .iter()
+            .map(|(idx, sig)| {
+                let idx_u8 = u8::try_from(*idx).map_err(|_| SigAggError::InvalidShareIndex {
+                    pubkey: *pubkey,
+                    idx: *idx,
```
varex83: I think it would be good to change the `ParSignedData` to have `share_idx` of type `u8`

crates/core/src/sigagg.rs
```diff
@@ -0,0 +1,847 @@
...
+        debug!("Threshold aggregated partial signatures");
+
+        for sub in &self.subs {
+            sub(duty, &output).await?;
+        }
+
+        Ok(())
+    }
+
+    async fn aggregate_one(
```
varex83, line 153:
```suggestion
	#[tracing::instrument(skip_all, fields(pubkey = %pubkey))]
    async fn aggregate_one(
```

crates/core/src/sigagg.rs
```diff
@@ -0,0 +1,847 @@
...
+                Some(ps.signed_data.as_ref())
+            })
+            .unwrap_or_else(|| par_sigs[0].signed_data.as_ref());
+
+        let agg_signed = template
+            .set_signature_boxed(Signature::new(agg_bytes))
+            .map_err(|e| SigAggError::SetSignature {
+                pubkey: *pubkey,
+                source: e,
+            })?;
```
varex83:
```suggestion
        let agg_signed = template
            .set_signature_boxed(Signature::new(agg_bytes))
            .map_err(|e| {
            	tracing::error!(error = %e, "set_signature failed");
            	SigAggError::SetSignature {
	                pubkey: *pubkey,
	                source: e,
            	}
            )?;
```

crates/core/src/sigagg.rs
```diff
@@ -0,0 +1,847 @@
...
+
+        (self.verify_fn)(pubkey, agg_signed.as_ref()).await?;
```
varex83:
```suggestion
        (self.verify_fn)(pubkey, agg_signed.as_ref()).await.map_err(|e| {
        	tracing::error!(error = %e, "verify failed");
        	e
        })?;
```

varex83: This is needed to follow the same logic as go:
```go
if err := a.verifyFunc(ctx, pubkey, aggSig); err != nil {
		span.RecordError(err)
		span.SetStatus(codes.Error, err.Error())

		return nil, err
	}
```

crates/core/src/sigagg.rs
varex83: Go reference: 
```go
// Inject signature into one of the parSigs resulting in aggregate signed data.
	aggSig, err := fullSig.SetSignature(tblsconv.SigToCore(sig))
	if err != nil {
		span.RecordError(err)
		span.SetStatus(codes.Error, err.Error())

		return nil, err
	}

```

crates/core/src/sigagg.rs
```diff
@@ -0,0 +1,847 @@
...
+        types::Signature as CoreSig,
+    };
+
+    fn noop_verify() -> VerifyFn {
+        Arc::new(|_, _| Box::pin(async { Ok(()) }))
+    }
+
+    #[derive(Debug, Clone, PartialEq, Eq)]
+    struct MockSignedData {
+        sig: [u8; 96],
```
varex83: Use `pluto_crypto::types::Signature`

crates/core/src/sigagg.rs
```diff
@@ -0,0 +1,847 @@
...
+        }
+
+        fn message_root(&self) -> std::result::Result<[u8; 32], SignedDataError> {
+            Ok([0u8; 32])
+        }
+    }
+
+    #[derive(Debug, Clone, PartialEq, Eq)]
+    struct FailSetSignatureMock {
+        sig: [u8; 96],
```
varex83: same

crates/core/src/sigagg.rs
```diff
@@ -0,0 +1,847 @@
...
+    fn mock_par_sigs(count: usize, share_idx: u64) -> Vec<ParSignedData> {
+        (0..count)
+            .map(|_| ParSignedData::new(MockSignedData { sig: [0u8; 96] }, share_idx))
+            .collect()
+    }
+
+    struct BLSContext {
+        pubkey: [u8; 48],
+        sigs: Vec<(u8, [u8; 96])>,
+        expected_agg: [u8; 96],
```
varex83: use `pluto-crypto` types

varex83 (changes requested):

crates/core/src/sigagg.rs
mskrzypkows: I thought about it, but what would it give us? A bit simpler definition of the callback which is not very complicated.

crates/core/src/sigagg.rs
mskrzypkows: Good idea, changed

crates/core/src/types.rs
```diff
@@ -553,7 +559,7 @@ pub struct ParSignedData {
     pub signed_data: Box<dyn SignedData>,
 
     /// Threshold BLS share index.
-    pub share_idx: u64,
+    pub share_idx: u8,
```
iamquang95: Any reason to change this from `u64` -> `u8`? `share_idx` is 1-based and max can be 256

crates/core/Cargo.toml
```diff
@@ -46,6 +46,7 @@ prost-types.workspace = true
 hex.workspace = true
 chrono.workspace = true
 test-case.workspace = true
+pluto-crypto.workspace = true
```
iamquang95: No need, it's on `dependencies` already

crates/core/src/types.rs
mskrzypkows: This is the reason: https://github.com/NethermindEth/pluto/pull/426#discussion_r3288946103
Hmm, why is it 1-based? That breaks possibility to keep it as u8 🤔

mskrzypkows: Or we can keep it in u8 but we need to remember to `+1` every time we use it.

mskrzypkows: It looks like this should be u64 instead of u8 to mach charon implementation: https://github.com/NethermindEth/pluto/blob/de42a3208ff8e37a31f6d398bb9e122aea725d0f/crates/crypto/src/types.rs#L23 

So instead of changing to u8 I've should change to u64 the index.

iamquang95: So in Definition
https://github.com/NethermindEth/pluto/blob/main/crates/cluster/src/definition.rs#L72
```
/// Charon nodes in the cluster and their operators.
/// Max 256 operators.
pub operators: Vec<Operator>,
```

On `node_idx()`
```
/// Returns the node index for a given peer ID.
    pub fn node_idx(&self, pid: &PeerId) -> Result<NodeIdx, DefinitionError> {
        let peers = self.peers()?;

        for (i, peer) in peers.iter().enumerate() {
            if peer.id == *pid {
                return Ok(NodeIdx {
                    peer_idx: i,
                    share_idx: peer.share_idx(),
                });
            }
        }

        Err(DefinitionError::PeerNotFound { peer_id: *pid })
    }
```

That's call `share_idx()`

```
 /// Returns share index of this Peer. ShareIdx is 1-indexed while peer index
    /// is 0-indexed.
    pub fn share_idx(&self) -> usize {
        self.index.wrapping_add(1)
    }
```

So it will overflow if you keep it as `u8`. Since in `dkg`, we map the `share_idx` from definition to core type, this will fail

https://github.com/NethermindEth/pluto/pull/426/changes#diff-f8e65bf8b68144c18a39c840773e15d737ffe51d055e3605b5cb5d074a9a4beeR219

iamquang95: Also this should be appear in the test, the boundary check

crates/core/src/sigagg.rs
```diff
@@ -0,0 +1,833 @@
...
+            return Err(SigAggError::InsufficientDistinctSignatures { pubkey: *pubkey });
+        }
+
+        let span = info_span!("BlstImpl::threshold_aggregate");
+        let agg_bytes = {
+            let _enter = span.enter();
+            BlstImpl.threshold_aggregate(&bls_sigs)
+        }
+        .map_err(|e| {
+            tracing::error!(parent: &span, error = %e, "threshold aggregate failed");
```
varex83: In the code we import `debug` from tracing, but here we use `tracing::error`, consider using only one approach around the codebase

varex83 (approved):

iamquang95 (approved):

PR #428, opened 2026-05-20
fix(relay-server): advertise configured external multiaddrs on / and /enr

varex83:
## Summary

Fixes both `/` and `/enr` returning no useful data in private-network / K8s deployments where libp2p only sees private listen addresses and `filter_private_addrs=true`.

**Root cause:** `AppState.addrs` was populated exclusively from `SwarmEvent::NewListenAddr`. External addresses configured via `--p2p-external-ip` / `--p2p-external-hostname` reach the swarm through `add_external_address(...)` but never entered that Vec, so:

- `/` returned `[]` once libp2p's listen addrs were filtered out as private.
- `/enr` short-circuited with `500 "no addresses"` before `apply_ip_override` ran, making `external_ip` and the DNS resolver dead code in this deployment shape.

**Fix:** compute `external_tcp_multiaddrs` + `external_udp_multiaddrs` from config at relay startup, thread them through `enr_server` into `AppState` as an immutable `Vec<Multiaddr>`, and union them with the live listeners (externals first, deduped) when serving both endpoints. Mirrors Go charon's `AddrsFactory` + `filterAdvertisedAddrs` shape.

For `/enr` specifically, the TCP/UDP scan loop now has a DNS fallback: `/dns/<host>/tcp/<port>` and `/dns/<host>/udp/<port>/quic-v1` multiaddrs are resolved via the cached `external_host_ip` produced by the existing resolver loop. The `/ip4/...` + `apply_ip_override` path is unchanged, so `external_ip` keeps priority over `external_host` when both are set.

## Files

- `crates/p2p/src/utils.rs` — `external_tcp_multiaddrs` / `external_udp_multiaddrs` visibility bumped from `pub(crate)` to `pub`.
- `crates/relay-server/src/utils.rs` — new `extract_dns_and_{tcp,udp}_port` helpers.
- `crates/relay-server/src/web.rs` — `AppState` gains `external_addrs` + `advertised_addrs()`; both handlers use the union.
- `crates/relay-server/src/p2p.rs` — compute `external_addrs` and pass to `enr_server`.

## Verification

Ran locally with the same env shape as the affected K8s deployment:

```bash
PLUTO_HTTP_ADDRESS=127.0.0.1:43640 \
PLUTO_P2P_ADVERTISE_PRIVATE_ADDRESSES=false \
PLUTO_P2P_EXTERNAL_HOSTNAME=example.com \
PLUTO_P2P_TCP_ADDRESS=127.0.0.1:43610 \
PLUTO_P2P_UDP_ADDRESS=127.0.0.1:43610 \
PLUTO_AUTO_P2PKEY=true \
./target/debug/pluto relay
```

`/`:
```json
[
  "/dns/example.com/tcp/43610/p2p/16Uiu2HAm...",
  "/dns/example.com/udp/43610/quic-v1/p2p/16Uiu2HAm..."
]
```

`/enr` returned `200 OK` with a valid ENR containing example.com's resolved IPv4 address.

### DIFF
--- a/crates/relay-server/src/web.rs
+++ b/crates/relay-server/src/web.rs
@@ -185,6 +185,7 @@ pub async fn monitoring_server(bind_addr: SocketAddr, ct: CancellationToken) {
 }
 
 /// Error response for HTTP handlers.
+#[derive(Debug)]
 pub struct HandlerError {
     status: StatusCode,
     message: String,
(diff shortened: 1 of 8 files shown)


crates/relay-server/src/web.rs
```diff
@@ -49,16 +53,34 @@ impl AppState {
...
     }
 
+    /// Returns the union of configured external multiaddrs and the live
+    /// libp2p listen addresses, externals first, deduped while preserving
+    /// order. Mirrors Go charon's `filterAdvertisedAddrs(externalAddrs,
+    /// internalAddrs, …)` — listeners are already filtered for private
+    /// addresses at ingest time when `filter_private_addrs` is set.
+    async fn advertised_addrs(&self) -> Vec<Multiaddr> {
+        let listeners = self.addrs.read().await;
+        let mut union: Vec<Multiaddr> = self.external_addrs.clone();
```
iamquang95: `external_addrs` can have duplicates. Better use `HashSet` here to simplify the code

iamquang95 (approved):

crates/relay-server/src/web.rs
```diff
@@ -174,11 +206,7 @@ pub async fn enr_handler(
 ) -> std::result::Result<String, HandlerError> {
     debug!("Getting ENR for node {}", state.peer_id);
 
-    let addrs = state.addrs.read().await;
-
-    // Sort addresses with public addresses first
-    let mut sorted_addrs: Vec<Multiaddr> = addrs.clone();
-    drop(addrs);
+    let mut sorted_addrs = state.advertised_addrs().await;
```
emlautarom1, line 209: question: We no longer sort addresses here and `advertised_addrs` makes no guarantee about ordering. Is this correct?

PR #430, opened 2026-05-21
feat: implement `core/aggsigdb`

### LINKED ISSUE #169: Implement `core/aggsigdb`
### Description
Port the module `core/aggsigdb` from the reference Go implementation.

### Acceptance Criteria

emlautarom1:
Closes #169

---

There are two implementations in Charon: V1 uses an Actor model with commands and queries, and V2 uses a Mutex alongside Go runtime trickery. The implementation in use in Charon is V1 due to concurrency issues in V2.

This implementation is closer to V2, and we use tokio's tools (ex. `Notify`) to replicate the Go runtime features.

Tests have been ported when appropriate, and some new ones were added to increase coverage.


emlautarom1:
@claude

emlautarom1:
@claude

emlautarom1:
There is a bug on Charon regarding immediately expired duties, which result in readers getting `SignedData` associated to an expired duty:

1. Goroutine A calls Await(duty, pubKey) and blocks waiting for data.
2. Goroutine B calls Store(duty, pubKey, data) for the same key.
3. While executing the write, the deadliner determines the duty is already expired.
4. The Await goroutine gets resolved with data tied to an expired duty.

Both V1 and V2 showcase this behaviour. The cause is that the result of `deadliner.Add` is ignored while also being imprecise:

1. If the duty can be scheduled, proceed with the write.
2. If the duty never expires, also proceed with the write.
3. If the duty is already expired skip the write.

Right now, cases 2 and 3 are treated under `false` (note the existing TODO in Charon).

crates/core/src/aggsigdb/memory.rs
```diff
@@ -0,0 +1,424 @@
...
+                        inserted = true;
+                        Ok(())
+                    }
+                    Entry::Occupied(slot) if slot.get() != &signed_data => {
+                        Err(Error::MismatchingData)
+                    }
+                    Entry::Occupied(_) => Ok(()),
+                });
+
+        if inserted {
```
mskrzypkows: The `self.0.data` can be unlocked before notifying

crates/core/src/aggsigdb/memory.rs
```diff
@@ -0,0 +1,424 @@
...
+            tokio::pin!(notified);
+            notified.as_mut().enable();
+
+            {
+                if let Some(data) = self.get(&duty, &pub_key).await {
+                    return data.clone();
+                }
+            }
+
+            notified.await;
```
mskrzypkows, line 123: Do we want to handle cancellation token here? I did it similarly in dutydb/memory.rs

crates/core/src/aggsigdb/memory.rs
```diff
@@ -0,0 +1,424 @@
...
+///
+/// Share an instance by cloning. Cloning is cheap and creates a new reference
+/// to the same underlying data.
+#[derive(Clone)]
+pub struct MemDB(Arc<MemDBInner>);
+
+type SignedDataByPubKey = HashMap<types::PubKey, Box<dyn types::SignedData>>;
+
+struct MemDBInner {
+    data: Mutex<HashMap<types::Duty, SignedDataByPubKey>>,
```
mskrzypkows: consider rwlock instead of mutex

varex83: I think it would be reasonable to use `DashMap` here instead of `Mutex<HashMap<...>>`

varex83 (approved):

crates/core/src/aggsigdb/memory.rs
emlautarom1: - No need to pass a `CancellationToken` to this method since callers can take the future from `wait_for` and use it in a a `select` or `timeout` safely.
- The background task exists as long as the `Deadliner` is emitting deadlines, and it already takes care of cancellation.

In summary, I don't think we need to handle cancellation here.

crates/core/src/aggsigdb/memory.rs
emlautarom1: Explored this for a bit and it seems like it would not work:
- We need to lock before adding the duty to the `Deadliner` to avoid data races.
- Using `DashMap` we can achieve the same by holding a reference to an entry first.
- `DashMap` uses sync locks, so we end up holding a sync lock across an await point (`deadliner.add()`)

Seems like this is a subtle, tricky and common bug: https://github.com/xacrimon/dashmap/issues/150

crates/core/src/aggsigdb/memory.rs
```diff
@@ -0,0 +1,426 @@
...
+/// to the same underlying data.
+#[derive(Clone)]
+pub struct MemDB(Arc<MemDBInner>);
+
+type SignedDataByPubKey = HashMap<types::PubKey, Box<dyn types::SignedData>>;
+
+struct MemDBInner {
+    data: RwLock<HashMap<types::Duty, SignedDataByPubKey>>,
+    deadliner: Arc<dyn Deadliner>,
+    notify: Notify,
```
iamquang95, line 29: https://github.com/ObolNetwork/charon/blob/v1.7.1/core/aggsigdb/memory.go#L24

Charon has the `chan quit` to mark that the MemDB is completed, every store or await later will be return an error.

With the current design in pluto, there is no "terminated" status for the MemDB, so user can still use `store` and `wait_for`

crates/core/src/aggsigdb/memory.rs
```diff
@@ -0,0 +1,426 @@
...
+                return;
+            };
+            inner.data.write().await.remove(&duty);
+        }
+    }
+
+    /// Stores aggregated signed duty data set.
+    pub async fn store(&self, duty: types::Duty, set: types::SignedDataSet) -> Result<(), Error> {
+        let mut should_notify = false;
+
```
iamquang95, line 68: If set is empty, should be no-op

mskrzypkows (approved):

emlautarom1:
Waiting for #434 before merging

crates/core/src/aggsigdb/memory.rs
emlautarom1: As mentioned, this is not strictly necessary since cancellation is handled when no longer needed.
Still, for the sake of compatibility I've added it

varex83 (approved):

emlautarom1:
There are a couple of issues on this implementation with no trivial fixes. Reworked in #452

PR #433, opened 2026-05-22
feat(core): implement consensus/timer

### LINKED ISSUE #158: Implement `core/consensus/timer`
### Description
Port the module `core/consensus/timer` from the reference Go implementation.

### Acceptance Criteria

iamquang95:
Fix: https://github.com/NethermindEth/pluto/issues/158

This PR:
- Port the consensus/timer to pluto.
- Use cancellable Tokio sleep futures as the Rust timer primitive.
- Cover timer behavior with parity-focused unit tests, including proposal timeout and eager timer deadlines.

### DIFF
--- a/crates/core/src/consensus/timer.rs
+++ b/crates/core/src/consensus/timer.rs
@@ -134,16 +134,12 @@ pub struct IncreasingRoundTimer {
 impl IncreasingRoundTimer {
     /// Creates an increasing round timer.
     pub fn new() -> Self {
-        Self::from_parts(None)
+        Self { duty: None }
     }
 
     /// Creates an increasing round timer for a duty.
     pub fn with_duty(duty: Duty) -> Self {
-        Self::from_parts(Some(duty))
-    }
-
-    fn from_parts(duty: Option<Duty>) -> Self {
-        Self { duty }
+        Self { duty: Some(duty) }
     }
 }
 
@@ -195,17 +191,16 @@ pub struct EagerDoubleLinearRoundTimer {
 impl EagerDoubleLinearRoundTimer {
     /// Creates an eager double linear round timer.
     pub fn new() -> Self {
-        Self::from_parts(None)
+        Self {
+            duty: None,
+            first_deadlines: Mutex::new(HashMap::new()),
+        }
     }
 
     /// Creates an eager double linear round timer for a duty.
     pub fn with_duty(duty: Duty) -> Self {
-        Self::from_parts(Some(duty))
-    }
-
-    fn from_parts(duty: Option<Duty>) -> Self {
         Self {
-            duty,
+            duty: Some(duty),
             first_deadlines: Mutex::new(HashMap::new()),
         }
     }
@@ -262,16 +257,12 @@ pub struct LinearRoundTimer {
 impl LinearRoundTimer {
     /// Creates a linear round timer.
     pub fn new() -> Self {
-        Self::from_parts(None)
+        Self { duty: None }
     }
 
     /// Creates a linear round timer for a duty.
     pub fn with_duty(duty: Duty) -> Self {
-        Self::from_parts(Some(duty))
-    }
-
-    fn from_parts(duty: Option<Duty>) -> Self {
-        Self { duty }
+        Self { duty: Some(duty) }
     }
 }
 
@@ -478,6 +469,24 @@ mod tests {
             "round 2 second timer did not fire",
         )
         .await;
+
+        assert_fires_after(
+            must_timer(timer.timer(3)),
+            Duration::from_millis(3_000),
+            "round 3 first timer did not fire",
+        )
+        .await;
+
+        let timeout = spawn_timeout(must_timer(timer.timer(3)));
+        advance(Duration::from_millis(2_500)).await;
+        tokio::task::yield_now().await;
+        assert!(!timeout.is_finished(), "round 3 second timer fired early");
+        assert_fires_after(
+            join_timeout(timeout),
+            Duration::from_millis(500),
+            "round 3 second timer did not fire",
+        )
+        .await;
     }
 
     #[test_case(1, Duration::from_millis(1_000) ; "round_1")]
@@ -636,11 +645,20 @@ mod tests {
 
     #[test]
     fn negative_round_returns_error() {
-        let timer = IncreasingRoundTimer::new();
+        let timers: Vec<(&str, Box<dyn RoundTimer>)> = vec![
+            ("increasing", Box::new(IncreasingRoundTimer::new())),
+            (
+                "eager_double_linear",
+                Box::new(EagerDoubleLinearRoundTimer::new()),
+            ),
+            ("linear", Box::new(LinearRoundTimer::new())),
+        ];
 
-        match timer.timer(-4) {
-            Ok(_) => panic!("negative round must fail"),
-            Err(err) => assert_eq!(Error::InvalidRound { round: -4 }, err),
+        for (name, timer) in timers {
+            match timer.timer(-4) {
+                Ok(_) => panic!("{name} negative round must fail"),
+                Err(err) => assert_eq!(Error::InvalidRound { round: -4 }, err),
+            }
         }
     }
 
@@ -660,6 +678,15 @@ mod tests {
 
     fn spawn_timeout(timeout: RoundTimerFuture) -> JoinHandle<Instant> {
         tokio::spawn(timeout)
+    }
+
+    fn join_timeout(timeout: JoinHandle<Instant>) -> RoundTimerFuture {
+        Box::pin(async move {
+            match timeout.await {
+                Ok(deadline) => deadline,
+                Err(err) => panic!("timer task failed: {err}"),
+            }
+        })
     }
 
     async fn assert_fires_after(timeout: RoundTimerFuture, duration: Duration, message: &str) {
--- a/crates/core/src/consensus/timer.rs
+++ b/crates/core/src/consensus/timer.rs
@@ -143,6 +143,12 @@ impl IncreasingRoundTimer {
     }
 }
 
+impl Default for IncreasingRoundTimer {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl RoundTimer for IncreasingRoundTimer {
     fn timer_type(&self) -> TimerType {
         TimerType::Increasing
@@ -200,6 +206,12 @@ impl EagerDoubleLinearRoundTimer {
     }
 }
 
+impl Default for EagerDoubleLinearRoundTimer {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl RoundTimer for EagerDoubleLinearRoundTimer {
     fn timer_type(&self) -> TimerType {
         TimerType::EagerDoubleLinear
@@ -251,6 +263,12 @@ impl LinearRoundTimer {
     /// Creates a linear round timer for a duty.
     pub fn with_duty(duty: Duty) -> Self {
         Self { duty: Some(duty) }
+    }
+}
+
+impl Default for LinearRoundTimer {
+    fn default() -> Self {
+        Self::new()
     }
 }
 
--- a/crates/core/src/consensus/timer.rs
+++ b/crates/core/src/consensus/timer.rs
@@ -370,7 +370,8 @@ fn linear_subsequent_round_timeout(round: i64) -> Result<Duration> {
 }
 
 fn duration_from_nanos(timeout_nanos: i64, round: i64) -> Result<Duration> {
-    let timeout_nanos = u64::try_from(timeout_nanos).map_err(|_| Error::InvalidRound { round })?;
+    let timeout_nanos =
+        u64::try_from(timeout_nanos).map_err(|_| Error::DurationOverflow { round })?;
 
     Ok(Duration::from_nanos(timeout_nanos))
 }
--- a/crates/core/src/consensus/timer.rs
+++ b/crates/core/src/consensus/timer.rs
@@ -281,7 +281,7 @@ impl RoundTimer for LinearRoundTimer {
         let timeout = match proposal_timeout_duration(self.duty.as_ref(), round) {
             Some(timeout) => timeout,
             None if round == 1 => Duration::from_secs(1),
-            None => linear_timer_timeout(round)?,
+            None => linear_subsequent_round_timeout(round)?,
         };
 
         timeout_from_now(timeout, round)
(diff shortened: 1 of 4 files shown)


iamquang95:
@claude review this PR

iamquang95:
All good, no need to fix any findings

mskrzypkows (approved):
LGTM

varex83 (approved):
LGTM

crates/core/src/consensus/timer.rs
```diff
@@ -0,0 +1,756 @@
...
+fn linear_round_timeout(round: i64) -> Result<Duration> {
+    ensure_non_negative_round(round)?;
+
+    let rounds = u32::try_from(round).map_err(|_| Error::DurationOverflow { round })?;
+    LINEAR_ROUND_INC
+        .checked_mul(rounds)
+        .ok_or(Error::DurationOverflow { round })
+}
+
+fn linear_timer_timeout(round: i64) -> Result<Duration> {
```
varex83agent: Confusingly similar helper names: `linear_round_timeout` (line 345, used by `EagerDoubleLinearRoundTimer::timer`) and `linear_timer_timeout` (line 354, used by `LinearRoundTimer::timer`) differ by one word but compute very different things (`LINEAR_ROUND_INC * round` vs the `200*(round-1) + 200` nanoseconds formula). A future refactor could easily swap them at the two call sites (lines 223 and 284) with no compile error. Suggest renaming the second to something like `linear_subsequent_round_timeout` or moving it onto `LinearRoundTimer` as an inherent method. Charon avoided this by inlining the expression directly inside `linearRoundTimer.Timer`.

crates/core/src/consensus/timer.rs
```diff
@@ -0,0 +1,756 @@
...
+    fn timer(&self, round: i64) -> Result<RoundTimerFuture> {
+        let timeout = match proposal_timeout_duration(self.duty.as_ref(), round) {
+            Some(timeout) => timeout,
+            None => linear_round_timeout(round)?,
+        };
+
+        let mut first_deadlines = self
+            .first_deadlines
+            .lock()
+            .map_err(|_| Error::TimerStatePoisoned)?;
```
varex83agent, line 217: Inconsistent poison handling: this line maps the per-instance `Mutex` poison to `Error::TimerStatePoisoned`, but `feature_enabled` (line 315) — also called inside `timer()` via `proposal_timeout_duration` — `.expect`s on `GLOBAL_STATE` poisoning. So one poisoned-lock path returns a recoverable error while the other panics, inside the same function. Per workspace convention (see `crates/dkg/src/sync/client.rs:247,254`), poisoned-lock = `.expect`. Suggest dropping `TimerStatePoisoned` (line 68-70) and using `.expect("timer state lock poisoned")` here for consistency; the critical section is too small to plausibly poison the lock anyway.

crates/core/src/consensus/timer.rs
```diff
@@ -0,0 +1,756 @@
...
+        .ok_or(Error::DurationOverflow { round })?;
+    let timeout_nanos = increment_nanos
+        .checked_add(200)
+        .ok_or(Error::DurationOverflow { round })?;
+
+    duration_from_nanos(timeout_nanos, round)
+}
+
+fn duration_from_nanos(timeout_nanos: i64, round: i64) -> Result<Duration> {
+    let timeout_nanos = u64::try_from(timeout_nanos).map_err(|_| Error::InvalidRound { round })?;
```
varex83agent: Wrong error variant. By the time `duration_from_nanos` is reached, `ensure_non_negative_round(round)?` (line 355) has already proven `round >= 0`. The `u64::try_from(timeout_nanos)` conversion here can only fail if the *computed* `timeout_nanos` is negative — which is an arithmetic-overflow condition, not an invalid-round one. Should be `Error::DurationOverflow { round }`. (Note: today this branch is actually unreachable for `round >= 0` since `(round-1)*200 + 200 >= 0`, but using the wrong variant if it ever becomes reachable would be misleading.)

crates/core/src/consensus/timer.rs
```diff
@@ -0,0 +1,756 @@
...
+pub enum Error {
+    /// Round produced a negative duration.
+    #[error("invalid consensus round: {round}")]
+    InvalidRound {
+        /// Invalid round.
+        round: i64,
+    },
+    /// Timer duration arithmetic overflowed.
+    #[error("timer duration overflow for round: {round}")]
+    DurationOverflow {
```
varex83agent, line 58: `DurationOverflow` (line 58) and `DeadlineOverflow` (line 64) both carry only `round: i64` and originate from interchangeable arithmetic paths — `DurationOverflow` from round arithmetic (`Duration::checked_mul`/`checked_add`/`u32::try_from`), `DeadlineOverflow` from `Instant::checked_add` (line 402). From a caller's perspective the failure is the same: "timer arithmetic overflowed for round N". Consider collapsing into a single `Overflow { round }` (or `TimerArithmeticOverflow { round }`).

crates/core/src/consensus/timer.rs
```diff
@@ -0,0 +1,756 @@
...
+        Self { duty: None }
+    }
+
+    /// Creates an increasing round timer for a duty.
+    pub fn with_duty(duty: Duty) -> Self {
+        Self { duty: Some(duty) }
+    }
+}
+
+impl Default for IncreasingRoundTimer {
```
varex83agent: All three timer structs have a manual `Default` impl (lines 146, 209, 269) that just calls `Self::new()`, where `new()` itself produces `duty: None` + (for EagerDoubleLinear) an empty `HashMap`. The whole boilerplate can be replaced with `#[derive(Default)]` on each struct — `Option`, `Mutex<HashMap<_,_>>`, and `HashMap` all implement `Default`. Either drop `new()` in favour of `T::default()` / `T::with_duty(...)`, or keep `new()` and derive `Default` to remove the three manual impls.

varex83agent (commented):
## Overall

Clean, well-documented port of Charon's `consensus/timer` package. The cancellable Tokio sleep model is a nice Rust-native fit, parity with the Go reference (including the deliberate preservation of Charon's nanoseconds-vs-milliseconds quirk in `linear_timer_timeout`) is faithful, and tests are well-structured around tokio's paused clock plus a serialised featureset guard. No `bug` or `major` severity issues. Findings below are mostly cleanup/consistency suggestions:

- **Error model** — `TimerStatePoisoned` is reachable only on programmer error and is inconsistent with how `GLOBAL_STATE` poisoning is handled (`.expect`); `duration_from_nanos` reports `InvalidRound` for what is really a `DurationOverflow`; `DurationOverflow`/`DeadlineOverflow` carry identical payloads and could merge.
- **Naming** — `linear_round_timeout` (used by `EagerDoubleLinear`) vs `linear_timer_timeout` (used by `LinearRoundTimer`) differ only by one word but compute very different things; `with_duty` shadows the workspace's builder-mutator convention; the trait method `RoundTimer::timer` collides with locals named `timer`.
- **API surface** — `#[derive(Default)]` would replace three hand-written `Default` impls that just call `new()`; `Clone` on `IncreasingRoundTimer`/`LinearRoundTimer` is never used through the trait object.
- **Tests** — missing coverage for `round = 0`, `round = i64::MAX`, the doubling-with-ProposalTimeout combination, and `FEATURESET_TEST_LOCK` cascades on test-panic poisoning.

Verdict: **COMMENT** — leave to author discretion. None of the findings block merge; the major ones are worth addressing in this PR, the rest are polish.

crates/core/src/consensus/timer.rs
```diff
@@ -0,0 +1,756 @@
...
+
+    /// Returns a timeout for the round.
+    ///
+    /// The caller owns cancellation by dropping the returned future before it
+    /// fires.
+    fn timer(&self, round: i64) -> Result<RoundTimerFuture>;
+}
+
+/// Implements a linearly increasing round timer.
+#[derive(Debug, Clone)]
```
varex83agent: `IncreasingRoundTimer` (line 129) and `LinearRoundTimer` (line 252) derive `Clone`, but `EagerDoubleLinearRoundTimer` cannot (it holds a `Mutex`). All three are consumed exclusively as `Box<dyn RoundTimer>` (via `RoundTimerFunc`), where the `Clone` bound is unreachable. Consider dropping `Clone` from the two concrete types for consistency, or document why an external consumer would clone a concrete timer.

crates/core/src/consensus/timer.rs
```diff
@@ -0,0 +1,756 @@
...
+        assert_fires_after(
+            timeout,
+            wake_duration(want),
+            "round 3 proposer timer did not fire",
+        )
+        .await;
+    }
+
+    #[test]
+    fn negative_round_returns_error() {
```
varex83agent, line 643: Test coverage gaps for the new error branches that have no Go counterpart:
- `round = i64::MAX` should hit `DurationOverflow` from each timer (via `u32::try_from` on lines 335, 348, or `checked_mul` on line 363).
- `round = 0` is a legal-but-pathological case: linear/eager-double-linear collapse to a 0-duration timer that fires immediately. Worth a test asserting current behaviour, even if just to lock it in.
- For `EagerDoubleLinearRoundTimer` + `ProposalTimeout` enabled, calling `timer(1)` *twice* exercises the doubling-with-proposal-timeout path (second call should yield `now + 3000ms`). Currently `proposal_timeout_optimization_double_eager_linear_round_timer` (line 590) only calls `timer(1)` once and then `timer(2)`.

crates/core/src/consensus/timer.rs
```diff
@@ -0,0 +1,756 @@
...
+
+    use pluto_featureset::{Config, FeatureSet};
+    use test_case::test_case;
+    use tokio::{task::JoinHandle, time::advance};
+
+    use super::*;
+    use crate::types::SlotNumber;
+
+    // Feature state is process-global.
+    static FEATURESET_TEST_LOCK: StdMutex<()> = StdMutex::new(());
```
varex83agent, line 399: `FEATURESET_TEST_LOCK: StdMutex<()>` will become poisoned if any test panics while holding the `FeatureSetGuard` (e.g. an assertion failure inside the `test()` closure of `with_featureset`). Every subsequent test that constructs a `FeatureSetGuard` will then panic at line 732 with `"featureset test lock poisoned"`, cascading the failure and hiding the original assertion error. Consider recovering via `.lock().unwrap_or_else(|e| e.into_inner())`, or switching to `parking_lot::Mutex` if it becomes a workspace dep.

Separately: this lock is private to this file but writes to a process-global `GLOBAL_STATE`. If any test in `crates/core` (or any test sharing the binary) ever writes to `GLOBAL_STATE` without acquiring this lock, the timer tests will race and flake. Worth promoting the lock into `pluto_featureset` as a `pub` helper (`with_global_state(...)`) so all writers naturally serialise.

crates/core/src/consensus/timer.rs
```diff
@@ -0,0 +1,756 @@
...
+/// It is eager, meaning it starts at an absolute time before proposal values
+/// are present. This aligns round start times across peers, which matters for
+/// leader election.
+///
+/// It is linear, meaning the round duration increases linearly with the round
+/// number: 1s, 2s, 3s, etc.
+#[derive(Debug)]
+pub struct EagerDoubleLinearRoundTimer {
+    duty: Option<Duty>,
+    first_deadlines: Mutex<HashMap<i64, Instant>>,
```
varex83agent, line 182: `std::sync::Mutex` is the right choice here — the critical section (lines 226-240) is short, synchronous, and the guard is dropped before the returned future is constructed. A future contributor reviewing async code might be tempted to "fix" this to `tokio::sync::Mutex`, which would be both unnecessary and would force `timer()` to become `async`. A one-line comment on the field — e.g. `// std::sync::Mutex: critical section is bounded and no guard escapes this function` — would prevent that.

Also worth a brief note that `first_deadlines` grows monotonically with rounds (matching Charon's `firstDeadlines map[int64]time.Time` — same parity-preserved property) so a future reader doesn't add eviction logic prematurely.

crates/core/src/consensus/timer.rs
```diff
@@ -0,0 +1,756 @@
...
+}
+
+impl IncreasingRoundTimer {
+    /// Creates an increasing round timer.
+    pub fn new() -> Self {
+        Self { duty: None }
+    }
+
+    /// Creates an increasing round timer for a duty.
+    pub fn with_duty(duty: Duty) -> Self {
```
varex83agent, line 141: `with_duty` shadows the workspace's `with_*` convention, which is overwhelmingly a *consuming builder mutator* (e.g. `with_timeout`, `with_relays`, `with_external_ip` in `crates/p2p`, `crates/peerinfo`, `crates/parsigex`). The three `with_duty` constructors here are alternative-`new` associated functions, not mutators on an existing instance. Rust API Guidelines (C-CTOR) suggest naming after the input or distinguishing role. Consider `IncreasingRoundTimer::for_duty(duty)` (and same for the other two) to avoid clashing with workspace conventions.

crates/core/src/consensus/timer.rs
```diff
@@ -0,0 +1,756 @@
...
+        tokio::task::yield_now().await;
+
+        assert!(timeout.is_finished(), "{message}");
+        match timeout.await {
+            Ok(_) => {}
+            Err(err) => panic!("timer task failed: {err}"),
+        }
+    }
+
+    fn wake_duration(duration: Duration) -> Duration {
```
varex83agent: nit: `wake_duration` clamps any duration to a 1ms minimum. The reason — that tokio's paused clock cannot reliably wake a sub-microsecond `sleep_until` — isn't obvious from the function name or body. A one-line doc comment explaining the paused-time edge case (and why the helper exists for the `linear_round_timer` test with its 400/600/800-nanosecond timeouts) would help future readers.

crates/core/src/consensus/timer.rs
```diff
@@ -0,0 +1,756 @@
...
+
+const PROPOSAL_TIMEOUT: Duration = Duration::from_millis(1_500);
+
+/// Timer errors.
+#[derive(Debug, thiserror::Error, PartialEq, Eq)]
+pub enum Error {
+    /// Round produced a negative duration.
+    #[error("invalid consensus round: {round}")]
+    InvalidRound {
+        /// Invalid round.
```
varex83agent, line 53: nit: the `round: i64` field-level doc comments on each `Error` variant (lines 53, 59, 65) restate the variant title and add no information — the variant-level `#[error("...round: {round}")]` already conveys what `round` means. Either tighten to a single-token description ("The offending round number.") or accept the duplication.

crates/core/src/consensus/timer.rs
```diff
@@ -0,0 +1,756 @@
...
+/// Provides the duration for each consensus round.
+pub trait RoundTimer: Send + Sync {
+    /// Returns the timer type.
+    fn timer_type(&self) -> TimerType;
+
+    /// Returns a timeout for the round.
+    ///
+    /// The caller owns cancellation by dropping the returned future before it
+    /// fires.
+    fn timer(&self, round: i64) -> Result<RoundTimerFuture>;
```
varex83agent, line 125: nit: the trait method `RoundTimer::timer` shares its name with both the type and common local variable names — call sites read `let timer = ...; timer.timer(round)`. Charon also uses `Timer` for the method, so this preserves a 1:1 mapping with upstream (which has value for porters), but a Rust-idiomatic alternative like `start(round)` or `deadline(round)` would read more clearly. Up to author discretion.

PR #435, opened 2026-05-22
fix: dkg use different ports and adjust timeout

iamquang95:
### Finding 1: Smoke port bind flake

Phase: Phase 7 (https://github.com/NethermindEth/pluto/actions/runs/26281190079/job/77357497425#step:11:104)
Error: ValidatorAPI ... listen tcp 127.0.0.1:39002: bind: address already in use
Cause: smoke ports used 39000+, inside Linux ephemeral port range. Readiness probes can collide with later node listener ports.
Fix: move smoke port base to 19000;

### Finding 2: Charon shutdown relay flake

Phase: Phase 4 (https://github.com/NethermindEth/pluto/actions/runs/26106918358/job/76773227878)
Error: NO_RESERVATION then Fatal error: sync shutdown
Cause: --shutdown-delay=5s too short. One node can exit/drop relay reservation before slower peers finish final sync shutdown.
Fix: raise SHUTDOWN_DELAY=120s; raise NODE_EXIT_TIMEOUT=180; log effective shutdown delay.


.github/workflows/dkg-runner.yml
```diff
@@ -105,8 +105,11 @@ jobs:
           CHARON_BIN: ${{ github.workspace }}/bin/charon
           PLUTO_BIN: ${{ github.workspace }}/target/debug/pluto
           TIMEOUT: ${{ inputs.timeout || '120' }}
+          SHUTDOWN_DELAY: "120s"
```
varex83: I think 30 seconds more the enough

varex83 (approved):

.github/workflows/dkg-runner.yml
emlautarom1: Agree, 2 minutes is a lot

emlautarom1 (approved):

Issue #179, opened 2026-01-28
Implement `core/tracker`


emlautarom1:
### Description
Port the module `core/tracker` from the reference Go implementation.

### Acceptance Criteria
All unit tests from the reference Go implementation must be ported and must pass.

PR #436, opened 2026-05-22
refactor: corrected DutyDefinitionSet to match Charon implementation

mskrzypkows:
Preparation for the: #179
In charon DutyDefinitionSet is keyed by pubilc key


emlautarom1 (commented):
Could you elaborate on why the change is needed? Why the `struct -> trait` conversion?

Remember to include a small summary for reviewers, or point to a specific ticket if appropriate.

mskrzypkows:
Ah, good point that's not really needed. I just need different key for the hashmap: public key instead of duty type.

emlautarom1 (approved):
I'm fine with this PR, it's an existing issue that was not caught since we haven't ported the clients of this code (ex. `core/scheduler`). Will probably require further attention though based on usages.

crates/core/src/types.rs
```diff
@@ -996,14 +1025,24 @@ mod tests {
         assert_eq!(pk.abbreviated(), "2a2_a2a");
     }
 
+    #[test]
+    fn duty_type_all() {
+        let all = DutyType::all();
+        assert_eq!(all.len(), 13);
+        assert!(all.iter().all(DutyType::is_valid));
+        assert!(!all.contains(&DutyType::Unknown));
+    }
```
iamquang95, line 1026: Need to check the uniqueness also

https://github.com/ObolNetwork/charon/blob/v1.7.1/core/types_test.go#L82

iamquang95 (approved):

PR #437, opened 2026-05-22
chore: switch default relays to pluto-relay nethermind endpoints

varex83:
## Summary
- Replaces the public Obol relays (`{0,1,2}.relay.obol.{tech,dev}`) with the new self-hosted endpoints `https://pluto-relay-{0,1}.ovh.dev-nethermind.xyz` across `DEFAULT_RELAYS`, example doc-comments, the mixed-DKG docker-compose, and the `dkg-runner` scripts.
- `DEFAULT_RELAYS` shrinks from 3 to 2 entries since only two endpoints are provisioned.

## Test plan
- [x] `cargo check --workspace`
- [ ] DKG smoke run against the new relays (`./scripts/dkg-runner/run.sh`)

### DIFF
--- a/crates/p2p/src/config.rs
+++ b/crates/p2p/src/config.rs
@@ -9,10 +9,9 @@ use std::{
 use libp2p::{Multiaddr, multiaddr, ping};
 
 /// Shared default relay endpoints used by commands and P2P-facing configs.
-pub const DEFAULT_RELAYS: [&str; 3] = [
-    "https://0.relay.obol.tech",
-    "https://2.relay.obol.dev",
-    "https://1.relay.obol.tech",
+pub const DEFAULT_RELAYS: [&str; 2] = [
+    "https://pluto-relay-0.ovh.dev-nethermind.xyz",
+    "https://pluto-relay-1.ovh.dev-nethermind.xyz",
 ];
 
 /// P2P configuration error.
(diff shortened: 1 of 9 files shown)


iamquang95 (approved):

crates/p2p/src/config.rs
```diff
@@ -9,10 +9,9 @@ use std::{
 use libp2p::{Multiaddr, multiaddr, ping};
 
 /// Shared default relay endpoints used by commands and P2P-facing configs.
-pub const DEFAULT_RELAYS: [&str; 3] = [
-    "https://0.relay.obol.tech",
-    "https://2.relay.obol.dev",
-    "https://1.relay.obol.tech",
+pub const DEFAULT_RELAYS: [&str; 2] = [
+    "https://pluto-relay-0.ovh.dev-nethermind.xyz",
+    "https://pluto-relay-1.ovh.dev-nethermind.xyz",
```
emlautarom1: I'm not sure we want to change the default relays to ours. I'd leave it as is, or even extend them.

Docs and examples should use ours though.

varex83: I think it would be reasonable to use our relays, at least we will test it as developing

emlautarom1: Let's check with the Obol team and confirm before merging.

**Update**: According to the Obol team we should have both (Charon defaults + Pluto relays)

emlautarom1 (approved):
Let's have all relays: Charon defaults + Pluto relays.

emlautarom1:
@varex83 Anything blocking us from merging? We would like to start using the relays on CI.

Issue #157, opened 2026-01-28
Implement `core/consensus/qbft`


emlautarom1:
### Description
Port the module `core/consensus/qbft` from the reference Go implementation.

### Acceptance Criteria
All unit tests from the reference Go implementation must be ported and must pass.

PR #442, opened 2026-05-26
feat(core): implement consensus/qbft/msg

iamquang95:
Part of: https://github.com/NethermindEth/pluto/issues/157


iamquang95:
@claude review this for me

crates/core/src/consensus/qbft/msg.rs
```diff
@@ -0,0 +1,717 @@
...
+    prepared_value_hash: [u8; 32],
+    values: sync::Arc<ValueMap>,
+    justification_protos: Vec<pbconsensus::QbftMsg>,
+    justification: Vec<qbft::Msg<ConsensusQbftTypes>>,
+}
+
+impl fmt::Debug for Msg {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Msg")
+            .field("type", &MessageType::from_wire(self.msg.r#type).to_string())
```
varex83agent, line 127: **Bug: Debug-formatting a `Msg` panics when `msg.r#type` is outside `0..=5`.**

This line calls `MessageType::from_wire(self.msg.r#type).to_string()`. The new `MessageType::from_wire` (qbft/mod.rs:208) is documented and tested to preserve any wire `i64` without clamping. But `Display for MessageType` (qbft/mod.rs:228) panics on its catch-all arm: `_ => panic!("bug: invalid message type")`. Combined, any `Msg` whose protobuf carries a malformed or unrecognised `type` field will abort the task as soon as anything `Debug`-prints it — `tracing::debug!(msg = ?m, ...)`, the `QbftLogger` callbacks that receive `&Msg<T>` (qbft/mod.rs:115-122), an `assert!` in a test that compares two `Msg`s, or even a `Vec<Msg>` printed at a higher layer. `Msg::new` does not validate `r#type`, so a peer can trigger this just by sending a QBFT message with `type=99`.

`Debug` should never panic. The minimal fix is to format the raw integer here instead of going through `Display`:
```rust
.field("type", &self.msg.r#type)
```
A more durable fix is to make `MessageType::Display` total (e.g. fall back to `unknown({n})` for invalid values) — see the companion comment on qbft/mod.rs:228. Either way, please also add a test that constructs a `Msg` with `type=99` and `Debug`-prints it.

crates/core/src/qbft/mod.rs
```diff
@@ -200,6 +204,11 @@ pub const MSG_DECIDED: MessageType = MessageType(5);
 const MSG_SENTINEL: MessageType = MessageType(6); // intentionally not public
 
 impl MessageType {
+    /// Converts a stable wire integer into a message type without clamping.
+    pub fn from_wire(value: i64) -> Self {
```
varex83agent, line 208: **Root cause of the `Msg::fmt` panic flagged on `consensus/qbft/msg.rs:119`.** `from_wire` is explicitly designed to preserve any wire `i64` (and is tested for that in `message_type_from_wire_preserves_unknown_wire_value`), but `Display for MessageType` further down in this file (`_ => panic!("bug: invalid message type")`) aborts on any value outside `0..=5`. The combination makes a panic reachable any time a `MessageType` flows into a format string. The same critique applies to `UponRule::Display` just below — that one is constructed only internally so the risk is lower, but the inconsistency is worth fixing alongside.

Suggested fix (in the existing Display impl):
```rust
impl Display for MessageType {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        match self.0 {
            0 => write!(f, "unknown"),
            1 => write!(f, "pre_prepare"),
            2 => write!(f, "prepare"),
            3 => write!(f, "commit"),
            4 => write!(f, "round_change"),
            5 => write!(f, "decided"),
            n => write!(f, "unknown({n})"),
        }
    }
}
```
(also: please switch the existing `write!(f, "{}", s)` to named-arg form `write!(f, "{s}")` per the rust-style guide.)

crates/core/src/consensus/qbft/msg.rs
```diff
@@ -0,0 +1,717 @@
...
+impl Msg {
+    /// Wraps a raw QBFT protobuf message for the generic core.
+    ///
+    /// Non-zero `value_hash` and `prepared_value_hash` fields must both exist
+    /// in `values`. Invalid hash encodings, including zero hashes, are
+    /// treated as the nil value and do not require a map entry.
+    ///
+    /// Justifications are raw protobuf messages from the same consensus
+    /// envelope. They are recursively wrapped with the same shared value map.
+    pub fn new(
```
varex83agent: **Defence-in-depth: `Msg::new` accepts an unknown wire `type` without validation.** The function validates value-hash and prepared-value-hash encodings against the `ValueMap`, but never checks `msg.r#type`. A peer can submit a QbftMsg with `type = 0` (MSG_UNKNOWN), `type = 6+`, or any negative `i64`, and the wrapper happily constructs a `Msg`. The QBFT runner downstream may assume `type_().valid()` was checked by the wrapper. Combined with the `Debug` panic flagged on line 119, this is the trigger for the remotely-reachable abort. Even without that, garbage flows through to consumers that may not filter.

Suggested guard:
```rust
if !MessageType::from_wire(msg.r#type).valid() {
    return Err(Error::InvalidMessageType(msg.r#type));
}
```

crates/core/src/consensus/qbft/msg.rs
```diff
@@ -0,0 +1,717 @@
...
+fn duty_from_proto(duty: Option<&pbcore::Duty>) -> Duty {
+    let Some(duty) = duty else {
+        return Duty::new(SlotNumber::new(0), DutyType::Unknown);
+    };
+
+    // Message receive validation rejects invalid duty types before this adapter
+    // is used by the consensus runner. If an invalid value reaches this local
+    // projection, Rust's closed enum maps it to Unknown instead of preserving
+    // the raw wire value.
+    let duty_type: DutyType = DutyType::try_from(duty.r#type).unwrap_or(DutyType::Unknown);
```
varex83agent, line 340: **Documented divergence from Charon `DutyFromProto`.** Charon (`core/proto.go:42`) preserves the raw wire value via `DutyType(duty.GetType())`. Two distinct invalid wire types (e.g. 99 and 100) round-trip distinctly in Go. The Rust closed-enum mapping collapses every unknown wire value to `DutyType::Unknown`. The inline comment acknowledges this and notes that upstream message-receive validation rejects invalid duty types before this adapter is invoked.

Worth promoting that note into the module-level documentation (currently only the inner-function comment mentions it), and adding a test verifying that whatever validation is expected upstream actually rejects unknown duty types. Otherwise a peer can craft two semantically distinct invalid duties that compare equal under `Msg::instance()` — possibly affecting per-instance buffering in the QBFT core.

crates/core/src/consensus/qbft/msg.rs
```diff
@@ -0,0 +1,717 @@
...
+    Sign(pluto_k1util::K1UtilError),
+}
+
+/// Wrapped consensus message consumed by the generic QBFT core.
+///
+/// The raw protobuf remains available for re-broadcasting. The hash fields are
+/// cached as `[u8; 32]` because the core treats consensus values as comparable
+/// hashes, not full protobuf payloads.
+#[derive(Clone)]
+pub struct Msg {
```
varex83agent, line 115: **Perf: `#[derive(Clone)]` deep-copies protobufs that Charon shares by pointer.** Each `Msg::clone()` deep-copies `msg: pbconsensus::QbftMsg` and `justification_protos: Vec<pbconsensus::QbftMsg>`. Charon's Go counterpart uses pointer fields (`*pbv1.QBFTMsg`, `[]*pbv1.QBFTMsg`), so multiple `Msg` instances share allocations. `Msg` is cloned every time it flows into the QBFT core's per-source FIFO buffers / justification vectors / decided sets, so the cost compounds.

Suggested shape:
```rust
pub struct Msg {
    msg: Arc<pbconsensus::QbftMsg>,
    value_hash: [u8; 32],
    prepared_value_hash: [u8; 32],
    values: Arc<ValueMap>,
    justification_protos: Arc<Vec<pbconsensus::QbftMsg>>,
    justification: Vec<qbft::Msg<ConsensusQbftTypes>>,
}
```
`values` already uses `Arc`, and the generic `justification: Vec<qbft::Msg<...>>` is already `Arc`-backed via the `Msg<T>` alias in `qbft/mod.rs:260`.

crates/core/src/consensus/qbft/msg.rs
```diff
@@ -0,0 +1,717 @@
...
+    #[error("value not found")]
+    ValueNotFound,
+
+    /// Callers must hash the concrete inner message, not `Any`.
+    #[error("cannot hash any proto, must hash inner value")]
+    CannotHashAnyProto,
+
+    /// Protobuf marshal failed.
+    #[error("marshal proto: {0}")]
+    MarshalProto(prost::EncodeError),
```
varex83agent: **Wrapped errors lose the error chain.** `MarshalProto(prost::EncodeError)`, `HashProto(HasherError)`, `RecoverPubkey(K1UtilError)`, and `Sign(K1UtilError)` capture an inner error solely to format it via `{0}`, but do not annotate it as `#[source]` (or `#[from]`). `std::error::Error::source()` therefore returns `None`, so anyhow's `.chain()`, tracing's chained Display, and any other walker can't see the underlying cause.

`MarshalProto` and `HashProto` wrap distinct types and can use `#[from]`, letting the `.map_err(...)` in `hash_proto` collapse to `?`. `RecoverPubkey` and `Sign` share `K1UtilError`, so keep the manual `.map_err` at the call site but mark the payload `#[source]`:
```rust
#[error("marshal proto")]
MarshalProto(#[from] prost::EncodeError),
#[error("hash proto")]
HashProto(#[from] HasherError),
#[error("recover pubkey")]
RecoverPubkey(#[source] pluto_k1util::K1UtilError),
#[error("sign")]
Sign(#[source] pluto_k1util::K1UtilError),
```

crates/core/src/consensus/qbft/msg.rs
```diff
@@ -0,0 +1,717 @@
...
+        self
+    }
+}
+
+/// Returns a deterministic SSZ hash root of a concrete protobuf message.
+///
+/// The hash input is deterministic protobuf encoding, then SSZ `PutBytes`
+/// merkleization. `Any` is rejected because the consensus value hash must bind
+/// to the inner message bytes, not the transport envelope.
+pub fn hash_proto<M>(msg: &M) -> Result<[u8; 32]>
```
varex83agent: **Wider `pub` visibility than Charon.** Charon's `msg.go` keeps `newMsg`, `hashProto`, `signMsg`, `verifyMsgSig`, and `toHash32` package-private; only `Msg` and its methods are exported because the type implements the interface used by the upper consensus runner. The Rust port instead exposes everything with `pub`: `hash_proto`, `sign_msg`, `verify_msg_sig`, `Msg::new`, `Msg::value_source`, `Msg::msg`, `Msg::values`, `Msg::to_consensus_msg`, `ConsensusQbftTypes`. None of these helpers is referenced outside `pluto-core` today (verified by grep).

For the free helpers (`hash_proto`, `sign_msg`, `verify_msg_sig`) prefer `pub(crate)` until a downstream crate genuinely needs them — preserves refactor freedom and matches Charon intent. `Msg` and `ConsensusQbftTypes` probably need `pub` (the consensus runner consuming this adapter will live elsewhere), but `Msg::new` itself can stay tighter as well.

crates/core/src/consensus/qbft/msg.rs
```diff
@@ -0,0 +1,717 @@
...
+
+    fn round(&self) -> i64 {
+        self.msg.round
+    }
+
+    fn value(&self) -> [u8; 32] {
+        self.value_hash
+    }
+
+    fn value_source(&self) -> std::result::Result<Any, qbft::QbftError> {
```
varex83agent, line 237: **Trait `value_source` double-wraps the lookup and loses error context.** The trait impl calls `Msg::value_source(self).map_err(|_| qbft::QbftError::ValueNotFound)`. The inherent method already does the exact `HashMap::get(...).cloned().ok_or(...)` lookup; this constructs and immediately discards an `Error::ValueNotFound` only to translate it.

Also, the `Error` enum has many variants — today only `ValueNotFound` is reachable from `Msg::value_source`, but any future change (e.g. lazy `hash_proto` decoding, cache lookup errors) would be silently rewritten to `QbftError::ValueNotFound` and masked in logs/metrics. Inline the lookup so the mapping is direct:
```rust
fn value_source(&self) -> std::result::Result<Any, qbft::QbftError> {
    self.values
        .get(&self.value_hash)
        .cloned()
        .ok_or(qbft::QbftError::ValueNotFound)
}
```

crates/core/src/consensus/qbft/msg.rs
```diff
@@ -0,0 +1,717 @@
...
+        let mut signed = sign_msg(&fixed_qbft_msg(), &key).unwrap();
+        signed.round += 1;
+
+        let ok = verify_msg_sig(&signed, &key.public_key()).unwrap();
+
+        assert!(!ok);
+    }
+
+    #[test]
+    fn verify_msg_sig_errors_on_empty_signature() {
```
varex83agent, line 701: **Test gap: no test for a non-empty but malformed signature.** Coverage today: empty signature (errors), wrong-key (returns `Ok(false)`), tampered message (returns `Ok(false)`). Missing: a non-empty signature with invalid length/shape (e.g. 64 random bytes instead of 65, or 65 bytes of garbage). That branch hits `pluto_k1util::recover`, which surfaces as `Error::RecoverPubkey`. Add a `verify_msg_sig_errors_on_malformed_signature` test so any future change to `pluto_k1util::recover` (e.g. accepting weird lengths) cannot silently change the contract.

crates/core/src/consensus/qbft/msg.rs
```diff
@@ -0,0 +1,717 @@
...
+    /// Justifications are raw protobuf messages from the same consensus
+    /// envelope. They are recursively wrapped with the same shared value map.
+    pub fn new(
+        msg: Option<pbconsensus::QbftMsg>,
+        justification: Vec<pbconsensus::QbftMsg>,
+        values: sync::Arc<ValueMap>,
+    ) -> Result<Self> {
+        let msg = msg.ok_or(Error::NilQbftMessage)?;
+
+        let value_hash = match to_hash32(&msg.value_hash) {
```
varex83agent, line 159: nit: the `match { Some(_) if guard => ..., Some(_) => err, None => default }` shape reads as if the second `Some(_)` arm depends on guard semantics (it does — that's why it works). The pattern is duplicated immediately for `prepared_value_hash`. A small helper deduplicates and makes the intent obvious:
```rust
fn lookup_hash(
    raw: &[u8],
    values: &ValueMap,
    missing: impl FnOnce() -> Error,
) -> Result<[u8; 32]> {
    match to_hash32(raw) {
        Some(hash) if values.contains_key(&hash) => Ok(hash),
        Some(_) => Err(missing()),
        None => Ok([0u8; 32]),
    }
}
```

crates/core/src/consensus/qbft/msg.rs
```diff
@@ -0,0 +1,717 @@
...
+    /// Returns the `Any` payload for this message's `value_hash`.
+    pub fn value_source(&self) -> Result<Any> {
+        self.values
+            .get(&self.value_hash)
+            .cloned()
+            .ok_or(Error::ValueNotFound)
+    }
+
+    /// Rebuilds the protobuf consensus envelope for transport.
+    pub fn to_consensus_msg(&self) -> pbconsensus::QbftConsensusMsg {
```
varex83agent, line 207: nit: `self.values.values().cloned().collect()` iterates a `HashMap`, so the resulting `QbftConsensusMsg.values` order varies between calls/processes. This matches Charon (`for _, v := range m.values`), and the envelope is not signed, but the `to_consensus_msg_preserves_raw_message_justification_and_values` test only papers over it by sorting both sides. If any downstream consumer (sniffer golden files, debug snapshots, future signature-over-envelope) ever assumes stable ordering, this silently regresses. Consider either documenting the non-determinism on `to_consensus_msg` or sorting the `Vec<Any>` by `(type_url, value)` before returning.

crates/core/src/consensus/qbft/msg.rs
```diff
@@ -0,0 +1,717 @@
...
+    // is used by the consensus runner. If an invalid value reaches this local
+    // projection, Rust's closed enum maps it to Unknown instead of preserving
+    // the raw wire value.
+    let duty_type: DutyType = DutyType::try_from(duty.r#type).unwrap_or(DutyType::Unknown);
+    Duty::new(SlotNumber::new(duty.slot), duty_type)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
varex83agent, line 346: nit: test-module imports are not grouped consistently with the production file (`std::*` / external / `crate::*` separated by blank lines on lines 23-33). Trivial to align: `use super::*;` and `use crate::qbft::{...};` in one group, `use prost::bytes::Bytes; use prost_types::Timestamp; use test_case::test_case;` in another.

varex83agent (dismissed):
## Summary

This PR ports `core/consensus/qbft/msg.go` from Charon v1.7.1 to Pluto Rust as `crates/core/src/consensus/qbft/msg.rs`. The functional parity is solid: `Msg::new`, `hash_proto`, `sign_msg`, `verify_msg_sig`, and `to_hash32` all map cleanly to their Go counterparts, golden-vector tests (`UNSIGNED_DATASET_HASH`, `QBFT_MSG_HASH`, `QBFT_MSG_SIGNATURE`) lock in cross-language byte parity, and the test surface meaningfully expands on Charon's (parametric hash cases, deterministic-encoding regression, sign/verify roundtrip with fixed keys). All 27 new tests pass locally.

## Bugs (must-fix)

1. **Debug-formatting a `Msg` with an unknown wire `type` panics the task** (`crates/core/src/consensus/qbft/msg.rs:119`). `Msg::fmt` calls `MessageType::from_wire(self.msg.r#type).to_string()`, and the new public `from_wire` constructor preserves any `i64` without clamping — but `MessageType`'s `Display` impl panics for any value outside `0..=5`. Since `Msg::new` does not validate `r#type`, a peer can construct a `Msg` whose `Debug` representation aborts the process. This is reachable from any `tracing` site or logger that touches a wrapped message, and `Debug` should never panic.

## Major findings

None beyond the bug above.

## Minor / nit findings

Closely related to the bug: `MessageType::Display` itself is the root cause (panic on the catch-all arm), and `Msg::new` should arguably reject `!MessageType::valid()` rather than relying on upstream filters. Other findings cover documented divergences from Charon (closed-enum `DutyType` projection, empty-vs-nil signature semantics), Rust-style polish (wider `pub` visibility than Go, missing `#[source]` on wrapped errors, double-wrap in trait `value_source`), an efficiency concern (`#[derive(Clone)]` deep-copies protobufs that Go shares by pointer), and a small test gap (no malformed-but-non-empty signature case).

## Verdict

Request changes for the `Debug`/`Display` panic — easy fix (have `Display` fall back to `unknown({n})` or branch on `valid()` in `Msg::fmt`). The rest is non-blocking and at author discretion.

crates/core/src/consensus/qbft/msg.rs
iamquang95: No need

crates/core/src/consensus/qbft/msg.rs
iamquang95: Not worth for now

crates/core/src/consensus/qbft/msg.rs
iamquang95: not worth

crates/core/src/consensus/qbft/msg.rs
```diff
@@ -0,0 +1,760 @@
...
+    /// Wraps a raw QBFT protobuf message for the generic core.
+    ///
+    /// Non-zero `value_hash` and `prepared_value_hash` fields must both exist
+    /// in `values`. Invalid hash encodings, including zero hashes, are
+    /// treated as the nil value and do not require a map entry.
+    ///
+    /// Justifications are raw protobuf messages from the same consensus
+    /// envelope. They are recursively wrapped with the same shared value map.
+    pub(crate) fn new(
+        msg: Option<pbconsensus::QbftMsg>,
```
mskrzypkows, line 153: I wonder if we need optional msg here, if passing without it returns an error in the first line of the function. I might be better to not call `new` at all in this case.

mskrzypkows (approved):

crates/core/src/consensus/qbft/msg.rs
iamquang95: Agree that we can change this to `msg: pbconsensus::QbftMsg`, the option is just charon compability API (they use pointer). I will change this in https://github.com/NethermindEth/pluto/pull/448

PR #444, opened 2026-05-26
feat: tracker module skeleton

mskrzypkows:
Main loop of tracker module, done similarly to deadliner with a handler. Missing functions will be implemented in next PR.


mskrzypkows:
@claude review

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+/// Holds the send-half of the event channel and implements the [`Tracker`]
+/// trait so core workflow components can submit events. The background loop
+/// that consumes those events lives in [`TrackerService`].
+pub struct TrackerHandle {
+    input_tx: mpsc::Sender<Event>,
+}
+
+impl TrackerHandle {
+    fn send_event(&self, event: Event) {
+        if let Err(e) = self.input_tx.try_send(event) {
```
varex83agent: **major: `try_send` silently drops telemetry events; diverges from Charon's blocking semantics.**

Charon's tracker uses an unbuffered channel with `select { case <-t.quit: return; case t.input <- event{...}: }` (core/tracker/tracker.go ~line 688), giving producers blocking back-pressure — events are never dropped, only delayed until either the loop accepts them or shutdown fires.

Pluto's `send_event` calls `try_send` and on `Full` only logs `warn!` and discards the event. The tracker is the sole source of duty-failure analysis and per-peer participation metrics; silent drops skew `duty_failed` / `duty_success` counters and can mask real failures during bursts. A single slot fanning a `ParSignedDataSet` across many validators can produce hundreds of events; exceeding 1024 in a burst quietly corrupts metrics rather than slowing the producer.

Compounding this, the per-drop `warn!` (line 118) fires once per dropped event with no rate-limit, so a sustained flood produces a wall of warn-level logs that itself drowns out other signal.

Recommendation: either (a) make this an async `send().await` matching Go, or (b) keep `try_send` but increment a `tracker_events_dropped_total` counter so drops are observable in metrics (not just logs), distinguish `Closed` (shutdown — silent) from `Full` (real drop — warn/count), and document the deliberate divergence in the PR description.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+
+                Some(e) = self.input_rx.recv() => {
+                    if e.duty.slot.inner() < self.from_slot {
+                        continue;
+                    }
+
+                    // Ignore expired or never-expiring duties.
+                    if !self.deleter.add(e.duty.clone()).await
+                        || !self.analyser.add(e.duty.clone()).await
+                    {
```
varex83agent: **major: Sequential `deleter.add().await` then `analyser.add().await` stalls the event loop and amplifies dropped events.**

Each `Deadliner::add` (deadline/mod.rs:132-148) round-trips through an internal channel (buffer = 1) plus a oneshot reply. While those two awaits are in flight, the tracker is **not** draining `input_rx`. Combined with the `try_send` drop policy from producers, the receiver-side stall translates directly into channel saturation and dropped telemetry for whichever duty is being added.

Charon uses unbuffered channels throughout, so producers naturally back-pressure during this period rather than drop. Two options:

1. Run the two `add` calls concurrently via `tokio::try_join!`/`tokio::join!` (their results are AND-ed, so concurrency is safe) — halves the stall window.
2. Only call `add(...)` when the duty is new to the local `events` map. Today the loop calls `add` on **every** event, paying the round-trip even when the duty was already added on a prior event. The `Deadliner` contract says `add` is idempotent and returns `true` if the duty was previously added, so correctness is preserved, but we're paying the latency every time.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,348 @@
+/// Failure reason definitions for duty analysis.
```
varex83agent, line 13: **major: No unit tests included.**

Charon's `core/tracker/tracker_internal_test.go` covers `TestStepString`, `TestTrackerFailedDuty` (FailAtConsensus, Success subtests), `TestAnalyseDutyFailed`, etc. This Rust PR adds zero tests. Step `Display` labels and Reason `code` strings end up as Prometheus labels — silent drift breaks dashboards downstream — and the trait fan-out determines event counts that drive every metric the tracker will emit.

At minimum this PR should add:

1. `step::Display` round-trip — every variant `Zero..=Sentinel` produces the exact Go label (`unknown`, `fetcher`, `consensus`, `duty_db`, `validator_api`, `parsig_db_local`, `parsig_ex`, `parsig_db_external`, `sig_aggregation`, `aggsig_db`, `bcast`, `chain_inclusion`, `sentinel`).
2. Each `pub const REASON_*` has the expected `code` field (this catches subtle cases like the intentional `REASON_ZERO_AGGREGATOR_SELECTIONS` → `"zero_aggregator_prepares"` name/code mismatch).
3. Trait fan-out — `par_sig_db_stored_internal` with a 3-pubkey `ParSignedDataSet` produces exactly 3 events, all with the same `duty` and `step = Step::ParSigDBInternal`; pubkeys-only methods produce exactly N events; `inclusion_checked` produces exactly 1.
4. Slot gate — events with `duty.slot < from_slot` are dropped *before* `deleter.add` / `analyser.add` are touched.

Flagging as `major` because all four properties affect downstream Prometheus labels or event counts and are easy to silently break in later refactors — without a regression net here, the skeleton becomes a load-bearing port without verification.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+                        continue;
+                    }
+
+                    events.entry(e.duty.clone()).or_default().push(e);
+                }
+
+                Some(duty) = self.analyser_rx.recv() => {
+                    // TODO: extract par sigs, analyse failed duty, report participation.
+                    let _ = &events;
+                    tracing::debug!(duty = %duty, "Duty analysis triggered (not yet implemented)");
```
varex83agent: **minor: Analyser arm is a stub that silently discards accumulated events.**

When `analyser_rx` fires, the arm only logs `Duty analysis triggered (not yet implemented)` at `debug!` and falls through. The accumulated `events` for that duty stay in the map until the deleter arm later removes them — but the duty is never analysed, no failure is reported, no participation metrics are emitted. Net effect of merging this PR alone: every event is stored and then dropped without analysis.

Acceptable as a deliberately staged skeleton (PR description says missing functions come next), but please:

1. Add a top-of-module doc comment stating that `tracker` is currently a no-op skeleton and analysis/reporting are not yet wired.
2. Update `TrackerService::start` doc to say the same — otherwise the wiring layer in `app/lifecycle` may assume tracker metrics are live as soon as `start` returns.
3. Optionally bump the `debug!` to `warn!` *once* (e.g. via `std::sync::Once`) so deployments don't silently rely on the stub.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+            input_rx,
+            analyser,
+            analyser_rx,
+            deleter,
+            deleter_rx,
+            from_slot,
+            peers,
+        };
+
+        tokio::spawn(task.run());
```
varex83agent: **minor: Fire-and-forget `tokio::spawn` discards the `JoinHandle`, preventing supervision and panic detection.**

If the loop ever panics — a future change introduces a bug, an `expect` is added, etc. — the caller has no way to observe the failure. The input channel will simply close and producers will see `TrySendError::Closed` (logged once per call by the existing `warn!`), masking the panic.

Charon's `Run(ctx) error` propagates failure back to its caller via the `lifecycle` package. Pluto today has no equivalent. Consider returning `(Arc<TrackerHandle>, JoinHandle<()>)` so `app` can `.await` the handle during shutdown and observe panics. The neighbouring `DeadlinerTask::start` (deadline/mod.rs ~line 205) has the same fire-and-forget pattern, so this is a broader project gap — but adding the supervision hook here makes it easy to fill in later.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+///
+/// Methods that only need validator pubkeys (fetcher, consensus, dutydb,
+/// sigagg, aggsigdb, bcast) accept `&[PubKey]` for object safety. Methods
+/// that also carry partial-signature data accept `&ParSignedDataSet`.
+///
+/// `err` is `Option<StepError>` (passed by value) so the caller's `Arc` can
+/// be cheaply cloned per event inside the implementation.
+pub trait Tracker: Send + Sync {
+    /// Called when the fetcher fetches duty data.
+    fn fetcher_fetched(&self, duty: Duty, pubkeys: &[PubKey], err: Option<StepError>);
```
varex83agent: **minor: Trait reduces Charon's typed sets to `&[PubKey]`, dropping payload that future analysis may want.**

Charon's `core.Tracker` (core/interfaces.go) passes typed sets: `DutyDefinitionSet`, `UnsignedDataSet`, `map[PubKey][]ParSignedData`, `SignedDataSet`. The tracker today only iterates keys, so reducing to `&[PubKey]` is functionally equivalent at v1.7.1. But this trait will appear in `app/lifecycle` wiring; every caller will have to materialise a `Vec<PubKey>` from a typed set they already hold, and the next PR's `analyseParticipation` may need set size or per-pubkey data (e.g. `expected_per_pubkey`). Future analysis additions would then require a trait change.

Consider mirroring the Go signatures (typed-set references) or accepting an iterator. If the reduced shape is intentional, please call it out in the trait doc-comment alongside the existing rationale for `Option<StepError>` by value.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+    pub fn start(
+        cancel: CancellationToken,
+        analyser: Arc<dyn Deadliner>,
+        deleter: Arc<dyn Deadliner>,
+        peers: Vec<PeerInfo>,
+        from_slot: u64,
+    ) -> Arc<TrackerHandle> {
+        let (input_tx, input_rx) = mpsc::channel(INPUT_BUFFER);
+
+        let analyser_rx = analyser.c().expect("analyser receiver already taken");
```
varex83agent: **minor: `.expect()` on `Deadliner::c()` panics on misuse instead of returning a `Result`.**

`analyser.c().expect("analyser receiver already taken")` and the matching `deleter.c()` call panic the spawning thread if `c()` was previously called. Because both `Deadliner`s are passed in as `Arc<dyn Deadliner>` (shareable), accidentally passing the same `Arc` to a second `start` invocation — e.g. in tests, integration wiring, or a reload path — hard-panics rather than surfacing as a recoverable error.

Consider `Result<Arc<TrackerHandle>, StartError>` so misconfiguration surfaces during app start-up rather than as a tokio task panic. The existing `# Panics` doc is good for now but the panic-on-misuse pattern locks in a tighter contract than necessary.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+
+    /// Called when aggregated data is broadcast to the beacon node.
+    fn broadcaster_broadcast(&self, duty: Duty, pubkeys: &[PubKey], err: Option<StepError>);
+
+    /// Called when chain inclusion is checked for a duty.
+    fn inclusion_checked(&self, duty: Duty, pubkey: PubKey, err: Option<StepError>);
+}
+
+/// Buffer capacity for the internal event channel.
+const INPUT_BUFFER: usize = 1024;
```
varex83agent: **minor: `INPUT_BUFFER = 1024` magic number without rationale.**

Charon's channel is unbuffered (`make(chan event)`); choosing 1024 here is a Pluto-specific design choice that affects worst-case memory (1024 × `sizeof(Event)`) and drop behaviour under load. The neighbouring `deadline/mod.rs` explains its `INPUT_BUFFER = 1` choice inline.

Please add a comment explaining (a) why 1024, (b) what fan-out this supports (peak event rate × loop processing latency), and (c) the failure mode when it overflows (current `try_send` drops them — see comment on line 117). This will help reviewers and operators understand the trade-off.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+        Arc::new(TrackerHandle { input_tx })
+    }
+
+    async fn run(mut self) {
+        let mut events: HashMap<Duty, Vec<Event>> = HashMap::new();
+
+        loop {
+            tokio::select! {
+                biased;
+
```
varex83agent, line 390: **minor: `biased;` diverges from Charon's unbiased select.**

Charon's `tracker.Run` uses an unbiased `select { case <-ctx.Done(); case e := <-t.input; case duty := <-t.analyser.C(); case duty := <-t.deleter.C() }`. Pluto adds `biased;` with cancellation first.

Subtle divergence: at shutdown Go may still drain a few pending events before observing `ctx.Done`; Pluto won't. Fine for telemetry, but either document the intentional change with a `// Diverges from Go: ...` comment or drop `biased;` — `select!`'s default fair polling still cancels pending futures correctly when the cancel arm fires.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+pub struct TrackerService {
+    cancel: CancellationToken,
+    input_rx: mpsc::Receiver<Event>,
+    analyser: Arc<dyn Deadliner>,
+    analyser_rx: mpsc::Receiver<Duty>,
+    deleter: Arc<dyn Deadliner>,
+    deleter_rx: mpsc::Receiver<Duty>,
+    from_slot: u64,
+    #[allow(dead_code)]
+    peers: Vec<PeerInfo>,
```
varex83agent, line 316: **minor: `peers` field is stored but never consumed.**

`TrackerService` accepts and stores `peers: Vec<PeerInfo>` (gated by `#[allow(dead_code)]`), but the analyser arm is a stub so the field is pure baggage today. Charon feeds peers into `newParticipationReporter(peers)` from the analyser case (tracker.go ~line 600), which is the next PR's responsibility.

Either (a) add a TODO comment referencing the follow-up PR and the participation reporter so the reservation is explicit, or (b) drop the parameter from `start` and re-add it when the reporter lands — leaving unused parameters in a public constructor invites callers to pass meaningless values.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+
+                    // Ignore expired or never-expiring duties.
+                    if !self.deleter.add(e.duty.clone()).await
+                        || !self.analyser.add(e.duty.clone()).await
+                    {
+                        continue;
+                    }
+
+                    events.entry(e.duty.clone()).or_default().push(e);
+                }
```
varex83agent, line 433: nit: `events.entry(e.duty.clone()).or_default().push(e)` clones `e.duty` to obtain the HashMap key even though `e` (which contains its own copy of `duty`) is moved into the `Vec` on the same line. `Duty` isn't `Copy`, so this is a real allocation per accepted event. Once analysis lands, consider hoisting the key extraction (`if let Some(v) = events.get_mut(&e.duty) { v.push(e) } else { events.insert(e.duty.clone(), vec![e]); }`) or storing duty once per bucket and elide it from `Event`.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+///
+/// Defined here to avoid a circular dependency with `pluto-p2p`
+/// (which already depends on `pluto-core`). Callers convert their
+/// `pluto_p2p::Peer` values before passing them to [`TrackerService::start`].
+#[derive(Debug, Clone)]
+pub struct PeerInfo {
+    /// Human-readable peer name.
+    pub name: String,
+    /// 1-indexed share index (`peer.index + 1`).
+    pub share_idx: usize,
```
varex83agent, line 48: nit: `share_idx: usize` — please document the `share_idx >= 1` invariant inline. Charon's `participatedShares map[int]int` uses `int` (and indices are 1-based: `Peer.Index + 1`). When `analyseParticipation` is ported in the next PR, the conversion should happen in one place (likely a `From<p2p::Peer>` impl) and the invariant in the doc will keep future callers from accidentally passing a 0-based index.

crates/core/src/tracker/reason.rs
```diff
@@ -0,0 +1,228 @@
...
+    code: "no_consensus",
+    short: "consensus algorithm didn't complete",
+    long: "Reason `no_consensus` indicates a duty failed in consensus step. This could indicate that insufficient honest peers participated in consensus or p2p network connection problems.",
+};
+
+/// Local validator client did not submit a partial signature for the duty.
+pub const REASON_NO_LOCAL_VC_SIGNATURE: Reason = Reason {
+    code: "no_local_vc_signature",
+    short: "signed duty not submitted by local validator client",
+    long: "Reason `no_local_vc_signature` indicates that partial signature we never submitted by the local validator client. This could indicate that the local validator client is offline, or has connection problems with charon, or has some other problem. See validator client logs for more details.",
```
varex83agent: nit: This long string is a verbatim copy of Charon's `reasonNoLocalVCSignature` long string, including the grammatical typo (`"partial signature we never submitted"` — should read `"was never submitted"`). Preserving the typo is the right call for log/metric parity with Charon (downstream alerting may grep on this exact text), but worth flagging so the team can choose whether to upstream a fix to Charon and propagate.

crates/core/src/tracker/step.rs
```diff
@@ -0,0 +1,57 @@
...
+    /// Partial signed data aggregated.
+    SigAgg = 8,
+    /// Aggregated signed data stored in aggsigdb.
+    AggSigDB = 9,
+    /// Aggregated data submitted to beacon node.
+    Bcast = 10,
+    /// Aggregated data included in canonical chain.
+    ChainInclusion = 11,
+    /// Sentinel — must always be last.
+    Sentinel = 12,
```
varex83agent, line 35: nit: `Step::Zero` (the "no event for this step" sentinel) and `Step::Sentinel` (the upper bound for backward-scan analysis matching Charon's `for step := sentinel - 1; step > zero; step--`) are defined correctly but not constructed anywhere in this PR. Please add a doc-comment referencing the follow-up PR / issue that will use them in `analyseDutyFailed`, otherwise a reviewer following only this PR will think they are dead variants.

varex83agent (changes requested):
Solid skeleton port — the Step enum, Reason constants, and trait shape all match Charon v1.7.1 verbatim (ordering, labels, codes; even the `"signature we never submitted"` typo is preserved for log/metric parity).

Three blocking items before merge:

1. **Telemetry drop semantics diverge from Charon.** Charon uses an unbuffered channel with `select { case <-quit: ...; case input <- event: }`, giving producers blocking back-pressure. Pluto uses a 1024-slot buffer plus `try_send` that silently drops events when full. Combined with the sequential `deleter.add().await; analyser.add().await` head-of-line in the loop, a single slot's fan-out can saturate the buffer and silently corrupt failed-duty analysis and participation metrics — exactly the signal the tracker exists to surface.
2. **No unit tests.** Step `Display` labels and Reason `code` strings end up as Prometheus labels; trait fan-out determines event counts. All three are silently breakable in later refactors and have zero coverage today. At least the Go test suite's `TestStepString` and a fan-out test should land with this PR.
3. **The analyser arm is a no-op stub** (`let _ = &events; tracing::debug!(...)`). Acceptable as a staged skeleton, but the public `TrackerService::start` doc must state plainly that the spawned task does not yet analyse or report — otherwise the wiring layer may assume metrics are live.

Minor items: prefer `Result` over `expect()` on `Deadliner::c()`; return a `JoinHandle` from `start` for supervision; document the `INPUT_BUFFER = 1024` choice; either remove `biased;` or call out the intentional divergence; either drop the unused `peers` argument or note it as reserved for the participation reporter.

Nits on per-event clones, the `&[PubKey]` trait shape vs Go's typed sets, and a few small comments are left inline for author discretion.

Verdict: **REQUEST_CHANGES** — the three majors should be addressed (or explicitly waived in the PR description) before merge.

mskrzypkows:
@varex83agent review again

mskrzypkows:
@varex83agent /review-pr

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,409 @@
...
+    }
+
+    async fn run(mut self) {
+        let mut events: HashMap<Duty, Vec<Event>> = HashMap::new();
+
+        loop {
+            tokio::select! {
+                // Cancellation is checked first so shutdown is never delayed by
+                // a busy event or deadliner channel.
+                biased;
```
varex83agent, line 389: The event loop uses `tokio::select! { biased; ... }` with branch order (cancel, `input_rx`, `analyser_rx`, `deleter_rx`). With `biased`, tokio polls branches strictly top-down and takes the first ready one. If `input_rx` stays ready (producers fanning out one event per pubkey under load, with `INPUT_BUFFER = 1024` of slack), the loop takes the input branch every iteration and never polls `analyser_rx`/`deleter_rx`. Awaiting the inner `tokio::join!` yields the task but does **not** re-poll the other select branches — they only get a turn when `select!` is re-entered and `input_rx` happens to be empty.

This diverges from Go's `select`, which picks fairly among ready cases so the deleter/analyser cases keep making progress.

The consequence is more than delayed cleanup: the deleter deadliner's output channel is bounded (`OUTPUT_BUFFER = 10`) and `handle_expired` uses `try_send`, **dropping** the duty and removing it from its own `duties` set when the channel is full (`deadline/mod.rs:333-346`). So if `deleter_rx` is starved long enough to fill its 10 slots, expired duties are silently dropped and the matching `events.remove(&duty)` (line 404) never runs — `events: HashMap<Duty, Vec<Event>>` then grows without bound. Suggest draining `deleter_rx`/`analyser_rx` with priority over `input_rx` (reorder them above input in the biased list, or bias only the cancel branch) so cleanup can't be starved.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,409 @@
...
+                }
+
+                Some(e) = self.input_rx.recv() => {
+                    if e.duty.slot.inner() < self.from_slot {
+                        continue;
+                    }
+
+                    // Run both deadliner adds concurrently to avoid stalling
+                    // the loop on two sequential channel round-trips.
+                    let (deleter_outcome, analyser_outcome) = tokio::join!(
```
varex83agent: Go gates accumulation with a short-circuiting OR: `if !t.deleter.Add(e.duty) || !t.analyser.Add(e.duty) { continue }` — so `analyser.Add` is **not** called when `deleter.Add` returns false (`tracker.go:144`). This PR runs both concurrently with `tokio::join!` and then checks both outcomes, so `analyser.add` is always invoked. (This was changed in the last commit `f6a46fb`; the prior commit had the Go-faithful sequential short-circuit.)

In correctly-configured setups this is benign: the deleter deadline is documented to be well after the analyser's, so a non-`Scheduled` deleter implies the analyser is also expired and its `add` is side-effect-free. The divergence only bites under an inverted/misconfigured deadline pairing, where Rust would arm a spurious analyser timer for a duty that was never inserted into `events`. The analyser branch is a stub today so the impact is nil — but consider restoring the short-circuit to match Go exactly; the cost of one extra `oneshot` round-trip is negligible.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,409 @@
...
+}
+
+impl TrackerService {
+    /// Builds the [`TrackerHandle`] and spawns the background event loop.
+    ///
+    /// `analyser` triggers duty analysis at deadline; `deleter` triggers
+    /// cleanup well after analysis (matching Go's contract that the deleter
+    /// deadline must be well after the analyser's). `from_slot` sets the
+    /// minimum slot to track — events for earlier slots are ignored.
+    pub fn start(
```
varex83agent, line 330: `start(cancel, analyser, analyser_rx, deleter, deleter_rx, peers, from_slot)` takes each `DeadlinerHandle` and its paired `mpsc::Receiver<Duty>` as separate args. Both receivers are `mpsc::Receiver<Duty>`, so passing `deleter_rx` where `analyser_rx` is expected (or vice-versa) type-checks but silently inverts behavior: analysis would fire on deleter deadlines and deletion on analyser deadlines — and since the deleter deadline is much later, analysis would run too late and deletion before analysis. Consider bundling each `(DeadlinerHandle, Receiver<Duty>)` into one newtype/struct so the pairing is un-swappable at the type level.

Related lifecycle note: `DeadlinerHandle::add` checks the **deadliner's own** `cancel_token`, which is independent of `TrackerService.cancel`. If a caller wires the deadliners with a different token than the tracker, cancelling one leaves the others (and the detached spawned tasks) running. Worth documenting in `start()` that the tracker and both deadliners must share the same cancellation token (or child tokens).

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,409 @@
...
+            from_slot,
+            peers,
+        };
+
+        tokio::spawn(task.run());
+
+        Arc::new(TrackerHandle { input_tx })
+    }
+
+    async fn run(mut self) {
```
varex83agent, line 381: This PR adds no unit tests, while Charon has an extensive `core/tracker/tracker_internal_test.go`. The analysis/participation logic is intentionally stubbed, so that can't be tested yet — but the behaviors this skeleton **does** implement are testable now and should be pinned before the next PR builds on them:
- `from_slot` filtering (line 377): `slot < from_slot` dropped; boundary `slot == from_slot` kept (Go uses strict `<`).
- add-gating (lines 383-393): a non-`Scheduled` duty is not accumulated; a `Scheduled` one is.
- per-duty accumulation (line 395) and removal on the deleter channel (line 404).
- per-pubkey fan-out: N pubkeys → N events with the right `Step` and `par_sig` presence.
- cancellation (line 372): `run()` returns promptly and later sends log-and-drop instead of hanging.

These can be driven with a mock/stub `DeadlinerHandle` (or `NeverExpiringCalculator` vs an always-expired calculator).

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,409 @@
...
+impl TrackerHandle {
+    async fn send_event(&self, event: Event) {
+        if let Err(e) = self.input_tx.send(event).await {
+            tracing::warn!(error = %e, "Tracker input channel closed; dropping event");
+        }
+    }
+}
+
+impl Tracker for TrackerHandle {
+    async fn fetcher_fetched(&self, duty: Duty, pubkeys: &[PubKey], err: Option<StepError>) {
```
varex83agent: The nine pubkey-slice / `ParSignedDataSet` trait impls (lines 168-298) are identical except for the `Step` constant and `par_sig` presence — ~140 lines of copy-paste where a wrong `Step` would silently mislabel events and the compiler wouldn't catch it. Two private async helpers collapse all of it and keep the only per-method-distinct fact (the `Step`) right at the call site:
```rust
async fn fan_out_pubkeys(&self, duty: Duty, step: Step, pubkeys: &[PubKey], err: Option<StepError>) {
    for pubkey in pubkeys {
        self.send_event(Event { duty: duty.clone(), step, pubkey: *pubkey, step_err: err.clone(), par_sig: None }).await;
    }
}
async fn fan_out_par_sigs(&self, duty: Duty, step: Step, set: &ParSignedDataSet, err: Option<StepError>) {
    for (pubkey, par_sig) in set.inner() {
        self.send_event(Event { duty: duty.clone(), step, pubkey: *pubkey, step_err: err.clone(), par_sig: Some(par_sig.clone()) }).await;
    }
}
```
Each method then becomes e.g. `self.fan_out_pubkeys(duty, Step::Fetcher, pubkeys, err).await`. Prefer a plain helper over a macro here.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,409 @@
...
+    /// Human-readable peer name.
+    pub name: String,
+    /// 1-indexed share index (`peer.index + 1`).
+    pub share_idx: usize,
+}
+
+/// Tracker receives events from core workflow components for duty analysis and
+/// participation reporting, matching Go's `core.Tracker` interface.
+///
+/// Methods that only need validator pubkeys (fetcher, consensus, dutydb,
```
varex83agent, line 54: This doc says methods "accept `&[PubKey]` for object safety", but every method returns `impl Future<Output = ()> + Send` (RPITIT), which already makes `Tracker` non-dyn-compatible — `dyn Tracker` can't be formed, and it's used nowhere (`TrackerService` holds a concrete `DeadlinerHandle` and `start()` returns `Arc<TrackerHandle>`). So the slice-vs-`Vec` choice has nothing to do with object safety. Recommend dropping that rationale and stating the real reason (avoid forcing callers to allocate, since the impl only iterates).

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,409 @@
...
+/// trait so core workflow components can submit events. The background loop
+/// that consumes those events lives in [`TrackerService`].
+pub struct TrackerHandle {
+    input_tx: mpsc::Sender<Event>,
+}
+
+impl TrackerHandle {
+    async fn send_event(&self, event: Event) {
+        if let Err(e) = self.input_tx.send(event).await {
+            tracing::warn!(error = %e, "Tracker input channel closed; dropping event");
```
varex83agent: nit: `send_event` logs at `warn!` on a closed channel. The channel only closes after `TrackerService::run` returns (shutdown), at which point every in-flight event from every producer hits this — a duty with N validators can emit up to N identical warnings during a clean shutdown. Consider `debug!` (shutdown is expected, not a warning) or logging once.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,409 @@
...
+/// Sized to absorb a full epoch's worth of events across all duty types and
+/// validators without back-pressuring producers while the loop is busy with a
+/// deadliner round-trip.
+const INPUT_BUFFER: usize = 1024;
+
+/// A single event emitted by a core workflow component.
+///
+/// `par_sig` is only set by `ValidatorAPI`, `ParSigDBInternal`, and
+/// `ParSigEx` events, matching Go's `event.parSig`.
+#[allow(dead_code)]
```
varex83agent, line 153: nit: the pinned toolchain supports `#[expect(dead_code)]` (stable since 1.81), which is preferable for known-temporary skeleton placeholders like this `Event` and `peers` (line 325): it fires if the item *stops* being dead, prompting removal of the attribute once a follow-up wires it up, whereas `#[allow]` lingers silently. The crate currently uses `#[allow]` throughout, so this is a consistency-vs-correctness call — `#[expect(dead_code, reason = "wired up in follow-up PR")]` fits a skeleton well.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,409 @@
...
+/// Buffer capacity for the internal event channel.
+///
+/// Sized to absorb a full epoch's worth of events across all duty types and
+/// validators without back-pressuring producers while the loop is busy with a
+/// deadliner round-trip.
+const INPUT_BUFFER: usize = 1024;
+
+/// A single event emitted by a core workflow component.
+///
+/// `par_sig` is only set by `ValidatorAPI`, `ParSigDBInternal`, and
```
varex83agent: nit: this says `par_sig` is set by "`ValidatorAPI`, `ParSigDBInternal`, and `ParSigEx`" — but `par_sig_db_stored_external` also sets it (line 255), which is correct and matches Go's `ParSigDBStoredExternal` (`tracker.go:771`). Also Go names the source `parSigExReceive`, not `parSigEx`. Worth aligning the comment so it doesn't under-list the external case while claiming to "match Go". (The same comment exists in Go and is equally imprecise there.)

crates/core/src/tracker/reason.rs
```diff
@@ -0,0 +1,228 @@
...
+    /// Short machine-readable code used as a metrics label.
+    pub code: &'static str,
+    /// One-line human-readable summary.
+    pub short: &'static str,
+    /// Full explanation shown in logs and documentation.
+    pub long: &'static str,
+}
+
+/// Unknown error occurred.
+pub const REASON_UNKNOWN: Reason = Reason {
```
varex83agent: nit: verified all 30 `REASON_*` consts field-by-field against Charon `core/tracker/reason.go` — every `code` and `short` (the values used as metric labels / log fields) matches verbatim, none missing or extra. The only differences are six `long` strings that intentionally rebrand "charon"→"pluto"; since `long` is never used as a label/wire value that's fine. One small thing: Go keeps these reasons **unexported**; here they're `pub`. Unless a cross-crate consumer is planned, `pub(crate)` would match Go's visibility and keep the crate's public surface smaller.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,409 @@
...
+        err: Option<StepError>,
+    ) -> impl Future<Output = ()> + Send;
+}
+
+/// Buffer capacity for the internal event channel.
+///
+/// Sized to absorb a full epoch's worth of events across all duty types and
+/// validators without back-pressuring producers while the loop is busy with a
+/// deadliner round-trip.
+const INPUT_BUFFER: usize = 1024;
```
varex83agent: nit: two notes on this constant. (1) The name collides conceptually with `deadline/mod.rs`'s local `INPUT_BUFFER` (=1); since they size very different channels, `EVENT_BUFFER` here would disambiguate when reading both modules. (2) Charon's tracker input is an **unbuffered** channel, so producers block until the loop accepts each event; the 1024 buffer is strictly more forgiving but changes the failure mode — under the starvation scenario above it lets producers run far ahead while cleanup stalls, amplifying memory growth rather than throttling. Worth re-evaluating the size once the (more expensive) analyser branch lands.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,409 @@
+/// Failure reason definitions for duty analysis.
```
varex83agent, line 13: nit: `tracker/mod.rs` has no top-of-file `//!` module doc, whereas `deadline/mod.rs` opens with a rich overview (including a runnable example). Given the channel topology here (handle vs service vs trait, two deadliners, a background task), a short `//!` block describing the architecture and its Charon counterpart would help readers.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,409 @@
...
+            input_rx,
+            analyser,
+            analyser_rx,
+            deleter,
+            deleter_rx,
+            from_slot,
+            peers,
+        };
+
+        tokio::spawn(task.run());
```
varex83agent: nit (forward-looking): `tokio::spawn(task.run())` detaches the loop with no retained `JoinHandle`. If `run()` ever panics, the task dies silently while `TrackerHandle` keeps accepting events (sends succeed until the receiver drops), so a failure of this monitoring component would be invisible. No panic is reachable in the current skeleton, and this mirrors the existing `deadline` module's spawn pattern — but the analyser branch (line 399 TODO) is where Charon does slice indexing / map lookups, so consider supervising or at least logging task exit before that logic lands.

varex83agent (changes requested):
## Summary

This PR ports the skeleton of Charon's `core/tracker` — the duty-lifecycle tracker that detects where a duty got stuck and (eventually) reports failures and peer participation. Scope is deliberately limited: the event loop, the `Tracker` trait + `TrackerHandle`, the `Step` enum, and the `reason` constants are implemented; the analysis/participation logic is stubbed for a follow-up PR (per the description). Overall quality is high — the code builds clean, `clippy -D warnings` and `cargo +nightly fmt --check` both pass, and parity with Charon v1.7.1 is strong: I verified the `Step` discriminants/labels, all 30 reason codes/shorts, and the trait→step mapping all match the Go reference exactly.

**Bugs (must-fix): none.**

The one blocking item is a **major** concurrency concern (1): the `biased` `select!` polls `input_rx` ahead of `analyser_rx`/`deleter_rx`, so under sustained input load the cleanup branch can be starved; combined with the deleter deadliner's bounded output channel + `try_send` (which drops duties when full), this can leak the `events` map unboundedly and silently. This diverges from Go's fair `select`.

The remaining findings are minor/nit: the `tokio::join!` departure from Go's short-circuit `deleter.Add() || analyser.Add()`; the swappable `(handle, rx)` arguments to `start()`; absence of unit tests for the already-implemented loop behaviors; ~140 lines of copy-paste across the nine fan-out impls; and a stale "for object safety" doc rationale (the trait is non-dyn-compatible due to RPITIT). Nits cover shutdown log spam, `#[allow]`→`#[expect]`, doc accuracy, reason visibility, the input-buffer naming/semantics, a missing module doc, and the detached spawn.

**Verdict: REQUEST_CHANGES** — solely for the starvation/leak risk in the core loop, which should be addressed (or explicitly justified) before further analysis logic is layered on top. Everything else is non-blocking.

mskrzypkows:
@varex83agent /review-pr

crates/core/src/tracker/reason.rs
```diff
@@ -0,0 +1,228 @@
...
+    /// Short machine-readable code used as a metrics label.
+    pub code: &'static str,
+    /// One-line human-readable summary.
+    pub short: &'static str,
+    /// Full explanation shown in logs and documentation.
+    pub long: &'static str,
+}
+
+/// Unknown error occurred.
+pub(crate) const REASON_UNKNOWN: Reason = Reason {
```
varex83agent, line 15: This crate fails the mandatory quality gate `cargo clippy --workspace --all-targets --all-features -- -D warnings` (AGENTS.md). All 30 `pub(crate)` reason constants are unused, so `dead_code` fires and — promoted by `-D warnings` — becomes a hard error (31 errors total, incl. the `task` field). `pub(crate)` items aren't externally reachable, so the lint applies.

The author already used `#[allow(dead_code)]` on `Event` and the `peers` field; the same is needed here. Simplest fix: add `#![allow(dead_code)]` at the top of `reason.rs` (these consts will be consumed by the analyser next PR), or `#[allow(dead_code)]` per const.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,547 @@
...
+
+/// Public-facing handle returned by [`TrackerService::start`].
+///
+/// Holds the send-half of the event channel and implements the [`Tracker`]
+/// trait so core workflow components can submit events. The background loop
+/// that consumes those events lives in [`TrackerService`].
+pub struct TrackerHandle {
+    input_tx: mpsc::Sender<Event>,
+    /// Kept so callers can detect task panics; also aborts on drop if desired.
+    pub(crate) task: tokio::task::JoinHandle<()>,
```
varex83agent, line 183: Two issues on this field:

1. **Build break:** like the reason constants, `task` is never read in non-test code, so `dead_code` fires and fails `cargo clippy ... -- -D warnings` (the mandatory gate). Add `#[allow(dead_code)]`, or actually expose/use it.

2. **Incorrect doc:** the comment says the handle "also aborts on drop if desired," but dropping a `tokio::task::JoinHandle` *detaches* the task — it keeps running to completion; it does **not** abort. Only `handle.abort()` (or `tokio_util::task::AbortOnDropHandle`) aborts. Since shutdown here is driven by the `CancellationToken`, reword to e.g.: "Kept so callers can `.await` the task to observe panics/completion; dropping it detaches the task — shutdown is driven by `cancel`."

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,547 @@
...
+/// Tracker receives events from core workflow components for duty analysis and
+/// participation reporting, matching Go's `core.Tracker` interface.
+///
+/// Methods that only need validator pubkeys (fetcher, consensus, dutydb,
+/// sigagg, aggsigdb, bcast) accept `&[PubKey]` for object safety. Methods
+/// that also carry partial-signature data accept `&ParSignedDataSet`.
+///
+/// `err` is `Option<StepError>` (passed by value) so the caller's `Arc` can
+/// be cheaply cloned per event inside the implementation.
+pub trait Tracker: Send + Sync {
```
varex83agent, line 60: The trait doc (lines 53-54) says non-parsig methods take `&[PubKey]` "for object safety," but this trait can never be dyn-compatible: every method returns `-> impl Future<Output = ()> + Send` (RPITIT / async-fn-in-trait), which precludes `dyn Tracker` regardless of argument types. So `&[PubKey]` doesn't buy object safety, and `Arc<dyn Tracker>` is impossible — consumers must be generic over `T: Tracker` (static dispatch).

Charon uses `core.Tracker` as a stored *interface* (dynamic dispatch) wired into many components (fetcher, consensus, dutydb, sigagg, …). With this shape each becomes generic over `T: Tracker`, propagating type params through the wiring, or you commit to a single concrete type. Worth deciding *now*, before consumers exist: if dynamic dispatch is wanted, return `Pin<Box<dyn Future<Output=()> + Send + '_>>` explicitly or use the `#[trait_variant::make]` / `async-trait` pattern. At minimum, fix the doc rationale.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,547 @@
...
+        let raw = Arc::try_unwrap(handle)
+            .unwrap_or_else(|_| panic!("single Arc owner in test"));
+        tokio::time::timeout(Duration::from_secs(1), raw.task)
+            .await
+            .expect("task did not exit within timeout")
+            .expect("task panicked");
+    }
+
+    #[tokio::test]
+    async fn events_accumulated_per_duty() {
```
varex83agent: This test — and `from_slot_filters_old_events` / `fan_out_sends_one_event_per_pubkey` — asserts only that the background task doesn't panic and exits on cancel, never the behavior its name claims. The `events` map is a function-local in `run()` (line 356) with no probe, so accumulation, `from_slot` filtering, and fan-out cardinality are all unobservable and unverified.

Worse, this test wires `NeverExpiringCalculator`, so `deleter.add()` returns `NoDeadline`, the `!= Scheduled` short-circuit at line 385 fires, and every event is `continue`d *before* line 392 — i.e. it exercises the non-accumulation path while being named for accumulation (the comment even admits it). The implemented deleter-removal path (line 375) and the `from_slot` filter (line 379) are likewise untested. Consider a test-only seam (an event-count probe channel, or a `#[cfg(test)]` accessor) so these can actually be asserted.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,547 @@
...
+                }
+
+                Some(e) = self.input_rx.recv() => {
+                    if e.duty.slot.inner() < self.from_slot {
+                        continue;
+                    }
+
+                    // Match Go's short-circuit: skip analyser entirely when
+                    // deleter returns non-Scheduled, avoiding a spurious timer.
+                    if self.deleter.add(e.duty.clone()).await != AddOutcome::Scheduled {
```
varex83agent, line 425: For every event the loop awaits two deadliner round-trips (`deleter.add().await` then `analyser.add().await`). Because there's one event per pubkey per step (the `send_pubkeys` fan-out), the same duty is re-added to both deadliners on the order of 10·N times over its lifetime. It's correct (the deadliner is idempotent), but each `add()` is an mpsc round-trip + oneshot ack, and while the loop is parked on them it drains neither `input_rx` nor the deadliner outputs.

Since entries are only removed by the deleter, a duty present in `events` means both deadliners already hold it — so you can gate on first sight:
```rust
if !events.contains_key(&e.duty) {
    if self.deleter.add(e.duty.clone()).await != AddOutcome::Scheduled { continue; }
    if self.analyser.add(e.duty.clone()).await != AddOutcome::Scheduled { continue; }
}
events.entry(e.duty.clone()).or_default().push(e);
```
Functionally equivalent, but cuts ~10·N round-trips to ~1 per duty and shrinks the window in which the loop isn't draining. (Charon calls `Add` every event too, but there it's a cheap in-process call, not a channel round-trip.)

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,547 @@
...
+                    return;
+                }
+
+                Some(duty) = self.analyser_rx.recv() => {
+                    // TODO: extract par sigs, analyse failed duty, report participation.
+                    tracing::debug!(duty = %duty, "Duty analysis triggered (not yet implemented)");
+                }
+
+                Some(duty) = self.deleter_rx.recv() => {
+                    events.remove(&duty);
```
varex83agent: `events.remove(&duty)` is the only place entries leave the map (the analyser branch is a stub this PR). The deleter deadliner emits expiries via a non-blocking `try_send` on a capacity-10 channel and, on `Full`, logs a warning and *still* removes the duty from its own tracking — so a dropped expiry notification means this `events` entry (holding `ParSignedData` for parsig steps) is never freed. With no compensating eviction, that's a permanent per-duty leak.

This mirrors Charon exactly (same try-send-then-delete in `core/deadline.go`) and lives in the unchanged deadliner, so it's not a blocker — flagging because in this skeleton the deleter is the *sole* memory bound. A TODO and/or a metric on the deadliner's drop would at least make the leak observable; revisit when the analyser is implemented.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,547 @@
...
+                // Cancellation and cleanup branches are checked first so that
+                // shutdown and HashMap shrinkage are never starved by a busy
+                // input channel.
+                biased;
+
+                _ = self.cancel.cancelled() => {
+                    return;
+                }
+
+                Some(duty) = self.analyser_rx.recv() => {
```
varex83agent: `Some(duty) = self.analyser_rx.recv()` (and the deleter equivalent on line 374) silently disable their select arm if the channel closes: `recv()` yields `None`, the `Some(..)` pattern fails, and `tokio::select!` permanently drops that branch. If a deadliner task dies abnormally (panic, or any early return not via the shared `cancel`), the tracker keeps running and accepting input but can no longer delete or analyse, so `events` grows unbounded. The module doc assumes both deadliners share the tracker's cancel token, but nothing enforces that at runtime. Consider treating a closed analyser/deleter receiver as fatal (log + shut the tracker down), mirroring Go's `quit` propagation.

crates/core/src/tracker/reason.rs
```diff
@@ -0,0 +1,228 @@
+/// A reason for a duty failing, matching Go's `tracker.reason`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Reason {
```
varex83agent, line 5: nit: The 30 constants are named `REASON_*` but live in module `reason`, so call sites read `reason::REASON_FETCH_BN_ERROR` — the prefix stutters (Go needs it because of flat package scope; Rust modules already namespace). Consider dropping the prefix (`reason::FETCH_BN_ERROR`) or, better, modeling them as associated consts on `Reason` (`Reason::FETCH_BN_ERROR`) so the set is enumerable/matchable — which the analyser will likely want.

crates/core/src/tracker/step.rs
varex83agent, line 35: nit: The comment marks `Sentinel` as load-bearing ("must always be last"; the backward step-scan in the analyser relies on it), but nothing enforces it — a variant added after it would silently break iteration. Cheap insurance: a static assertion `const _: () = assert!(Step::Sentinel as u8 == 12);` and/or a unit test locking the order (`Step::Fetcher < Step::Consensus < … < Step::Sentinel`).

varex83agent (dismissed):
This PR adds the tracker module skeleton (Step enum, Reason constants, the `Tracker` trait, and the `TrackerService` event loop), porting Charon's `core/tracker`. The port is faithful: the Step enum values/labels, all 30 reason strings, the deleter-before-analyser `||` short-circuit, and the `from_slot` lower-bound filter all match Charon v1.7.1 exactly, and analysis/participation reporting is appropriately deferred to a follow-up.

**Bugs (must fix before merge):**
1. The crate does **not compile** under the mandatory gate `cargo clippy --workspace --all-targets --all-features -- -D warnings` (AGENTS.md): the 30 `pub(crate)` reason constants and the `TrackerHandle.task` field are never read, triggering 31 `dead_code` errors. `#[allow(dead_code)]` was applied to `Event`/`peers` but missed `reason.rs` and `task`.
2. (same root cause, second site) the unused `task` field — whose doc comment about abort-on-drop is also factually wrong.

**Minor/nit themes:** the `Tracker` trait is not dyn-compatible (RPITIT) yet its doc claims `&[PubKey]` is "for object safety" — worth resolving before consumers are wired in, since Charon uses dynamic dispatch; the unit tests assert only no-panic/clean-shutdown and never their named behavior (filtering, fan-out, accumulation), leaving this PR's logic effectively untested; plus a few event-loop robustness/efficiency notes (redundant per-event deadliner round-trips, sole removal path, deadliner-death handling).

Requesting changes for the build break; the remaining items are minor/nit and left to author discretion.

mskrzypkows:
@varex83agent /review-pr

varex83agent (commented):
## Tracker module skeleton — multi-agent review

This PR ports the **skeleton** of Charon's `core/tracker` (`v1.7.1`): the `Tracker` trait, the background event loop, the `Step` enum, and the `reason` table. The actual failure analysis and participation reporting are deliberately deferred. **Overall the port is faithful and clean.** Verified against the Go reference: the `Step` enum values/ordering/`Display` labels match `stepLabels` exactly; all 28 `reason` strings match `reason.go` verbatim with correct `charon`→`pluto` substitutions (and faithfully preserved upstream typos); the method→step mapping, the `deleter`-before-`analyser` short-circuit (Go's `if !deleter.Add || !analyser.Add`), and the `from_slot` filter all match. Concurrency was checked: no data race (the input branch's awaits suspend the whole single-task loop, matching Go's synchronous `Run`) and no tracker↔deadliner deadlock (the deadliner's output `try_send` is non-blocking, breaking the cycle).

**Bugs (must-fix): none.**

**Findings are all minor / nit** (left for author discretion):
1. `Event.par_sig` doc comment is inaccurate — it omits `ParSigDBExternal` (which the code *does* set) and wrongly names `ValidatorAPI` (which has no callback). The code is correct; only the comment misleads. (mod.rs:151)
2. `events_accumulated_per_duty` tests the opposite of its name (with `NeverExpiringCalculator` every event is skipped before accumulation), and the suite as a whole only asserts "task didn't panic" — no assertions on map contents, fan-out cardinality, deletion, or `from_slot` filtering. (mod.rs:476, :500)
3. The `events` map's only eviction (deleter notification) is best-effort and can be dropped on a full output buffer, orphaning entries — but this is **faithful to Charon** and rooted in `deadline/mod.rs` (unchanged here), so it's flagged only for the follow-up analysis PR. (mod.rs:386)
4. Lifecycle/idiom notes: `JoinHandle` behind `Arc` is unreachable in prod and the doc overstates it (mod.rs:180); document the RPITIT `+Send` rationale (mod.rs:60); `biased` select trade-off vs Go (mod.rs:365); shutdown log noise (mod.rs:188); over-broad `#![allow(dead_code)]` and uninstantiable `pub Reason` (reason.rs); inert `#[repr(u8)]` discriminants (step.rs).

Verdict: **COMMENT** — no bugs or majors; the items above are quality/clarity improvements that don't block merge. Recommend at least fixing the `par_sig` doc comment (#1) and renaming/strengthening the misleading test (#2) before this lands, since both will mislead the follow-up implementer.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,554 @@
...
+/// Buffer capacity for the internal event channel.
+///
+/// Sized to absorb a full epoch's worth of events across all duty types and
+/// validators without back-pressuring producers while the loop is busy with a
+/// deadliner round-trip.
+const EVENT_BUFFER: usize = 1024;
+
+/// A single event emitted by a core workflow component.
+///
+/// `par_sig` is only set by `ValidatorAPI`, `ParSigDBInternal`, and
```
varex83agent: This doc comment is inaccurate in two ways, even though the **code is correct**:

1. It claims `ValidatorAPI` sets `par_sig`, but the `Tracker` trait has no `ValidatorAPI` callback at all (correctly — Go's `core.Tracker` has none; `validatorAPI` is a *derived* step). Nothing ever emits an event with `step == ValidatorAPI`, so `par_sig` is never set for it.
2. It omits `ParSigDBExternal`. In Charon `core/tracker/tracker.go` the callbacks that set `parSig:` are `ParSigDBStoredInternal`, `ParSigExBroadcasted`, **and** `ParSigDBStoredExternal`. The Rust impl faithfully mirrors this — `par_sig_db_stored_external` routes through `send_par_sig_set` and sets `par_sig: Some(..)`.

The comment appears to have been copied from Go's own stale `event` doc (which wrongly says `validatorAPI, parSigDBInternal and parSigExReceive`). Suggest describing the actual Rust behavior: *`par_sig` is set by `ParSigDBInternal`, `ParSigEx`, and `ParSigDBExternal` events.* This matters because the deferred `analyseParticipation` reads `par_sig` precisely on the internal+external par-sig steps, so an implementer trusting this comment could mishandle external events.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,554 @@
...
+
+        let raw = Arc::try_unwrap(handle).unwrap_or_else(|_| panic!("single Arc owner in test"));
+        tokio::time::timeout(Duration::from_secs(1), raw.task)
+            .await
+            .expect("task did not exit within timeout")
+            .expect("task panicked");
+    }
+
+    #[tokio::test]
+    async fn events_accumulated_per_duty() {
```
varex83agent: `events_accumulated_per_duty` does not test accumulation — it tests the opposite. With `NeverExpiringCalculator`, `deleter.add(..)` returns `NoDeadline` (not `Scheduled`), so every event hits the `!= AddOutcome::Scheduled` short-circuit at line 401 and is dropped **before** ever reaching `events.entry(..).or_default().push(e)` at line 408. The comment on line 502-504 admits this ("every event is skipped before accumulation ... verifies no panic"). So the test named for accumulation specifically exercises the path where accumulation never happens, and asserts nothing about the `events` map.

Either rename it (e.g. `never_expiring_events_are_skipped`) or rewrite with a future-deadline calculator plus a real assertion on map contents (which needs a test-only inspection hook, since `events` is private to the task).

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,554 @@
...
+
+        let raw = Arc::try_unwrap(handle).unwrap_or_else(|_| panic!("single Arc owner in test"));
+        tokio::time::timeout(Duration::from_secs(1), raw.task)
+            .await
+            .expect("task did not exit within timeout")
+            .expect("task panicked");
+    }
+
+    #[tokio::test]
+    async fn from_slot_filters_old_events() {
```
varex83agent, line 500: All four tests only assert that the spawned task exits within 1s without panicking after `cancel.cancel()`; none inspects observable behavior. `from_slot_filters_old_events` sends a below-threshold (slot 5) and above-threshold (slot 15) event but never asserts the old one was filtered or the new one scheduled. `fan_out_sends_one_event_per_pubkey` sends 3 pubkeys but never asserts 3 events were produced. There is no test for the deleter-removes-events path at all.

Because the loop is a fully-encapsulated task with no observability hook, these behaviors are currently unobservable from a test. Two consequences worth addressing before the analysis logic lands: (a) add a test-only way to read the `events` map (or an event count), then assert accumulation / fan-out cardinality / from_slot filtering / deleter removal; (b) the single `tokio::task::yield_now().await` (line 488) does not deterministically drain the input channel — the input branch performs two awaited deadliner round-trips across three tasks per event, so a single yield gives no ordering guarantee. The tests pass today only because they assert nothing about post-processing state; the moment a real state assertion is added, this becomes flaky-by-construction. Prefer awaiting observable state over `yield_now`.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,554 @@
...
+                        None => {
+                            tracing::error!("Analyser deadliner channel closed unexpectedly; stopping tracker");
+                            return;
+                        }
+                    }
+                }
+
+                duty = self.deleter_rx.recv() => {
+                    match duty {
+                        Some(duty) => { events.remove(&duty); }
```
varex83agent, line 410: Latent concern for when the analyser TODO is implemented (not a regression — flagging for the follow-up). The `events` map is evicted only here, on a duty arriving via `deleter_rx`. But that delivery is best-effort: the underlying `DeadlinerTask::handle_expired` emits expired duties with `try_send` on a capacity-10 channel and, on `Full`, only logs `"Deadliner output channel full"` and then **unconditionally** removes the duty from its own tracking set — so a dropped notification is never re-emitted. If a burst of same-slot deleter deadlines (>10) fires while this loop is busy in a `deleter.add(..).await`/`analyser.add(..).await` round-trip, surplus deletions are lost and the corresponding map entries (each a `Vec<Event>` holding cloned `ParSignedData`) are orphaned permanently → slow unbounded growth. A dropped *analyser* notification likewise silently skips that duty's failure analysis + participation report.

Note this is **faithful to Charon** (`core/deadline.go` does the identical `select { case chan <- duty: default: warn }` + unconditional `delete`), and the root cause lives in `deadline/mod.rs`, which this PR does not touch — so per the functional-equivalence rule it is arguably correct as-is. Raising it only so the follow-up analysis PR considers a reliability/health-signal mitigation (e.g. a max-slot watermark sweep) rather than relying solely on the lossy deleter notification.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,554 @@
...
+pub struct DeleterRx(pub mpsc::Receiver<Duty>);
+
+/// Public-facing handle returned by [`TrackerService::start`].
+///
+/// Holds the send-half of the event channel and implements the [`Tracker`]
+/// trait so core workflow components can submit events. The background loop
+/// that consumes those events lives in [`TrackerService`].
+pub struct TrackerHandle {
+    input_tx: mpsc::Sender<Event>,
+    /// Kept so callers can detect task completion or panics by awaiting it.
```
varex83agent, line 180: `TrackerService::start` returns `Arc<TrackerHandle>` and the `JoinHandle` is stored here behind that `Arc`. In production the `Arc` is cloned across producers, so `Arc::try_unwrap` will fail and the `JoinHandle` is permanently inaccessible — meaning this doc comment ("Kept so callers can detect task completion or panics by awaiting it" / "call `.abort()` to cancel it") describes something callers cannot actually do through the public API. Only the tests reach it, and only because they hold the sole reference. The task's real lifecycle is governed entirely by the shared `CancellationToken`; dropping the last `Arc` detaches (does not abort) the loop.

This is coherent (no leak as long as the token is eventually cancelled), but consider either (a) exposing a `shutdown()`/`join()` method, or (b) updating the comment to state the handle is detached and lifecycle is cancel-token-driven. Separately, the skeleton `#[allow(dead_code)]` markers here and on `Event`/`peers` would be better as `#[expect(dead_code, reason = "wired up in the analysis follow-up")]` so they re-fire (and prompt removal) once used.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,554 @@
...
+
+    async fn run(mut self) {
+        let mut events: HashMap<Duty, Vec<Event>> = HashMap::new();
+
+        loop {
+            tokio::select! {
+                // Cancellation and cleanup branches are checked first so that
+                // shutdown and HashMap shrinkage are never starved by a busy
+                // input channel.
+                biased;
```
varex83agent, line 389: nit: The `biased;` ordering (cancel → analyser → deleter → input) diverges from Charon's `Run`, whose `select` is fair/pseudo-random and cannot starve the input case. In practice the divergence is harmless and arguably safer here — the deadliner receivers are low-rate (one duty per expiry, cap-10 buffer) so sustained readiness that would starve `input_rx` is implausible, and prioritizing cleanup/shutdown avoids unbounded map growth. The existing comment explains the shutdown/shrinkage rationale well; consider adding one line noting this is an intentional fairness trade-off vs Go's non-biased select.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,554 @@
...
+    input_tx: mpsc::Sender<Event>,
+    /// Kept so callers can detect task completion or panics by awaiting it.
+    /// Dropping the handle detaches the task; call `.abort()` to cancel it.
+    #[allow(dead_code)]
+    pub(crate) task: tokio::task::JoinHandle<()>,
+}
+
+impl TrackerHandle {
+    async fn send_event(&self, event: Event) {
+        if let Err(e) = self.input_tx.send(event).await {
```
varex83agent, line 188: nit: On normal shutdown the loop returns (on cancel) and drops `input_rx`, after which every in-flight producer call funnels through here and emits a `warn!` per dropped event — potentially hundreds per fan-out duty-set. A closed tracker channel is expected at shutdown (mirroring how `deadline`'s `handle_expired` treats `Closed(_)` as a quiet return), so consider downgrading this to `debug!`, or rate-limiting, so it can't become its own log-flood under shutdown.

crates/core/src/tracker/mod.rs
```diff
@@ -0,0 +1,554 @@
...
+/// Tracker receives events from core workflow components for duty analysis and
+/// participation reporting, matching Go's `core.Tracker` interface.
+///
+/// Methods that only need validator pubkeys (fetcher, consensus, dutydb,
+/// sigagg, aggsigdb, bcast) accept `&[PubKey]`. Methods that also carry
+/// partial-signature data accept `&ParSignedDataSet`.
+///
+/// `err` is `Option<StepError>` (passed by value) so the caller's `Arc` can
+/// be cheaply cloned per event inside the implementation.
+pub trait Tracker: Send + Sync {
```
varex83agent, line 60: nit: Using `-> impl Future<Output = ()> + Send` in the trait while the impl uses plain `async fn` is the **correct** idiom here — bare `async fn` in a trait can't express the `+ Send` bound, which is load-bearing because `TrackerService::run` is `tokio::spawn`ed and the handle is shared across producers; `#[async_trait]` would add a `Box::pin` alloc on the hot fan-out path. The only gap is that nothing documents *why*, so a future contributor may "simplify" it to `async fn` and silently drop `Send`. Worth one line on the trait doc: methods return `impl Future + Send` so the returned futures are `Send` (required by the spawned consumer).

crates/core/src/tracker/reason.rs
```diff
@@ -0,0 +1,230 @@
+#![allow(dead_code)]
```
varex83agent, line 1: nit: A file-level `#![allow(dead_code)]` silences the lint for the entire module indefinitely, including future helpers that genuinely shouldn't be dead. Prefer `#![expect(dead_code, reason = "reasons consumed by the analyser in a follow-up PR")]` (stable) so the lint re-fires once the consts are wired up and the attribute gets removed — a blanket `allow` tends to become permanent.

crates/core/src/tracker/reason.rs
```diff
@@ -0,0 +1,230 @@
...
+
+/// A reason for a duty failing, matching Go's `tracker.reason`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Reason {
```
varex83agent, line 5: nit: `Reason` is `pub` with `pub` fields, but every `REASON_*` value is `pub(crate)`, so no external crate can ever obtain or construct a `Reason` — the `pub` on the type is currently dead public surface. Make the type `pub(crate)` to match the values (it's an internal tracker concept until participation reporting is exposed), or keep `pub` deliberately if a future public API will surface it. (Separately: the const-table-of-structs shape is a faithful Go port; an `enum Reason` with `code()/short()/long()` methods would be more idiomatic Rust and give exhaustiveness in the analyser match — optional, worth a comment if the table form is a deliberate parity choice.)

crates/core/src/tracker/step.rs
```diff
@@ -0,0 +1,57 @@
+use std::fmt::Display;
+
+/// Step in the core workflow, matching Go's `tracker.step`.
+///
+/// Variants are ordered by their position in the workflow; this ordering is
+/// used when scanning backwards to find the last reached step.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+#[repr(u8)]
```
varex83agent, line 8: nit: `#[repr(u8)]` plus the explicit `= 0..=12` discriminants are inert for the stated purpose. The backward-scan ordering comes entirely from `#[derive(PartialOrd, Ord)]`, which orders by *declaration order*, not the numeric value; the explicit discriminants/`repr` would only matter for `Step as u8` casts (forbidden by the style guide) or a stable wire representation, neither of which occurs. They do faithfully mirror Go's `iota`, which is a fine reason to keep them — just noting they're not required by the scan logic the doc cites.

mskrzypkows:
@varex83agent /review-pr

varex83agent:
⏳ @mskrzypkows — rate limit: please wait ~1 more minute(s) before triggering another review.

iamquang95 (approved):
lgtm

PR #447, opened 2026-05-27
fix(keystore): flush file before return

iamquang95:
https://docs.rs/tokio/latest/tokio/fs/struct.File.html

> A file will not be closed immediately when it goes out of scope if there are any IO operations that have not yet completed. To ensure that a file is closed immediately when it is dropped, you should call [flush](https://docs.rs/tokio/latest/tokio/io/trait.AsyncWriteExt.html#method.flush) before dropping it. Note that this does not ensure that the file has been fully written to disk; the operating system might keep the changes around in an in-memory buffer. See the [sync_all](https://docs.rs/tokio/latest/tokio/fs/struct.File.html#method.sync_all) method for telling the OS to write the data to disk.

### DIFF
--- a/crates/eth2util/src/keystore/store.rs
+++ b/crates/eth2util/src/keystore/store.rs
@@ -230,6 +230,7 @@ async fn write_file(path: impl AsRef<Path>, data: &[u8], mode: u32) -> Result<()
         .open(path.as_ref())
         .await?;
     file.write_all(data).await?;
+    file.flush().await?;
 
     Ok(())
 }


emlautarom1:
Are there any other similar scenarios (tests or production code) where we should flush? Let's extend the PR with those too if any.

iamquang95:
We need to flush for `tokio::fs::File::write_all` only, the `tokio::fs::File::write` in tokio wraps a blocking call `std:: fs::write` so it's safe, no need to manually flush. I checked the codebase, and there is only one place use the `tokio::fs::File::write_all` is in the `keystore`

emlautarom1 (approved):

emlautarom1:
Good opportunity to expand our agent skills and include this gotcha https://github.com/NethermindEth/pluto/blob/main/.claude/skills/rust-style/SKILL.md

varex83 (approved):

PR #448, opened 2026-05-28
feat(core): implement consensus qbft transport and sniffer

iamquang95:
Part of: https://github.com/NethermindEth/pluto/issues/157

This PR implements transport state and sniffer, part of the consensus qbft. This's currently on top of https://github.com/NethermindEth/pluto/pull/442 (hasn't merged yet)


iamquang95:
@claude review

iamquang95:
@varex83agent /review-pr

varex83agent (commented):
Solid, well-tested port of `transport.go` + `sniffer.go` with good module-level documentation in `msg.rs`. Sniffer is a clean 1:1 mapping. Transport faithfully bridges the tokio mpsc admission buffer with the crossbeam channel that the QBFT core consumes, and the comments at lines 81-83 and 193-194 capture the non-obvious design choices.

No bugs or hard blockers. Findings are predominantly **minor** style/idiom or contract-documentation gaps, plus a few **nit** cleanups:

- **Hash-binding contract** is implicit: `hash_proto_bytes` hashes raw `Any.value` bytes rather than re-marshalling deterministically like Go's `hashProto` does (transport.rs:131, msg.rs:281). This works today because prost emits deterministic bytes for the proto types that flow through QBFT (`BTreeMap` for map fields), but the invariant is undocumented and a future caller could easily reach for `hash_proto_bytes` on an inbound-validation path where the input bytes are NOT guaranteed deterministic. Worth tightening the docstring.
- A couple of intentional behavioral divergences from Go (`Ok(())` on outer-rx close vs Go's zero-value-loop, `Err(ReceiveBufferClosed)` on inner-send failure vs Go's panic) — both are improvements, but the docstring on `process_receives` should call them out.
- `Broadcaster` is typed `Box<dyn Fn(...) -> BoxFuture<'static, Result<()>>>` where `Result` is `transport::Result` — real implementations have no error variant to return wire/marshal failures through. Worth either adding `Error::Broadcast(...)` or relaxing the broadcaster signature before the runner wires this up.
- Self-send is detached as `tokio::spawn` (transport.rs:195) and relies on the caller to cancel `ct` at instance teardown; if the contract isn't held, the task can outlive the transport. The contract is implicit — worth a sentence in the doc.
- Several smaller style/clarity nits: shadowed `hash` in `get_value`, inconsistent `Arc::clone` style, field-name collision `Transport.values`/`ValueStore.values`, and one duplicate test.

Verdict: comment-only review — none of the findings block merge, but the contract-documentation items are worth addressing before the consumers of these helpers land.

crates/core/src/consensus/qbft/transport.rs
```diff
@@ -0,0 +1,762 @@
...
+                    .iter()
+                    .map(|(hash, value)| (*hash, value.clone())),
+            );
+    }
+
+    /// Returns a cached value by hash, after draining at most one local value.
+    pub(crate) fn get_value(&self, hash: [u8; 32]) -> Result<Any> {
+        let mut store = self.values.lock().unwrap_or_else(PoisonError::into_inner);
+        if let Ok(local) = store.value_rx.try_recv() {
+            let hash = msg::hash_proto_bytes(&local.value)?;
```
varex83agent, line 138: **Hash-determinism contract is implicit and undocumented.** Here we compute the consensus hash from the raw bytes of `Any.value` (whatever the producer happened to put in there). Go's `hashProto` (charon/core/consensus/qbft/msg.go:138-166) instead unmarshals to a typed proto and re-marshals with `proto.MarshalOptions{Deterministic: true}` before hashing — so Go is byte-canonical regardless of how the value was originally encoded.

This works today because:
- The local producer of `value_rx` uses `prost_types::Any::from_msg(&typed_msg)`, and prost's `encode_to_vec` is byte-deterministic in field-tag order.
- The proto messages that flow through consensus use `BTreeMap` for map fields (see `hash_proto_uses_btree_map_for_deterministic_encoding` in msg.rs:413) — without this, prost map-field encoding is non-deterministic.

The invariant `Any.value bytes ≡ canonical deterministic proto encoding of the inner message` is currently held by the build-proto configuration plus the producer's use of `Any::from_msg`. It's not enforced by the type system or documented at this call site. A future contributor implementing inbound wire-validation (analogous to Go's `valuesByHash`) could naturally reach for `hash_proto_bytes` and get silent consensus desync if any new proto message ships with a `HashMap` field.

Recommend a one- or two-line comment here that captures the contract, and consider linking it from the `hash_proto_bytes` docstring.

crates/core/src/consensus/qbft/msg.rs
```diff
@@ -274,9 +274,14 @@ where
     let mut encoded = Vec::with_capacity(msg.encoded_len());
     msg.encode(&mut encoded).map_err(Error::MarshalProto)?;
 
+    hash_proto_bytes(&encoded)
+}
+
+/// Returns the consensus hash for deterministic protobuf bytes.
+pub(crate) fn hash_proto_bytes(encoded: &[u8]) -> Result<[u8; 32]> {
```
varex83agent, line 279: `hash_proto` (line 266) explicitly rejects `Any` (test `hash_proto_rejects_any`). `hash_proto_bytes` has no such guard — it hashes whatever bytes it is handed. Doc currently reads only `Returns the consensus hash for deterministic protobuf bytes.`

Given the helper is `pub(crate)` and is the natural shortcut a future inbound-validation layer would reach for, the docstring should spell out the invariant explicitly: bytes MUST be the deterministic prost encoding of a non-`Any` concrete message, equivalent to `prost::Message::encode_to_vec` over a typed proto whose map fields are `BTreeMap`. Anything else risks silent hash mismatch against Go peers, which re-marshal deterministically before hashing.

crates/core/src/consensus/qbft/transport.rs
```diff
@@ -0,0 +1,762 @@
...
+    pub(crate) round: i64,
+    pub(crate) value_hash: [u8; 32],
+    pub(crate) prepared_round: i64,
+    pub(crate) prepared_value_hash: [u8; 32],
+    pub(crate) justification: Vec<qbft::Msg<ConsensusQbftTypes>>,
+}
+
+/// Errors returned by the QBFT consensus transport.
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum Error {
```
varex83agent, line 50: `Broadcaster` is `Box<dyn Fn(...) -> BoxFuture<'static, Result<()>>>` where `Result<()>` is the local `transport::Result<()>`. The `Error` enum (lines 49-75) only has variants for the transport's own failures (`UnknownValue`, `InvalidJustification*`, `InvalidDuty`, `ReceiveBufferClosed`, `Msg(msg::Error)`) — there is no variant for broadcaster-supplied failures (libp2p stream errors, peer unreachable, serialization issues from a real implementation).

Go's `broadcaster.Broadcast` returns a plain `error` (transport.go:22), so any implementation can return its own typed error. Here a real broadcaster will need either to invent a synthetic mapping or to stuff its errors into `Error::Msg(msg::Error)`, which is semantically wrong.

Consider adding `Error::Broadcast(#[source] Box<dyn std::error::Error + Send + Sync>)` and wrapping at the call site (line 206), or changing the broadcaster signature to return `Result<(), Box<dyn Error + Send + Sync>>` and wrapping into that new variant. Cheaper to fix now than after the runner is wired.

crates/core/src/consensus/qbft/transport.rs
```diff
@@ -0,0 +1,762 @@
...
+        let consensus_msg = msg.to_consensus_msg();
+        let inner_msg: qbft::Msg<ConsensusQbftTypes> = msg.clone();
+
+        let task_ct = ct.clone();
+        let recv_tx = self.recv_tx.clone();
+        let sniffer = sync::Arc::clone(&self.sniffer);
+        let sniffed_msg = consensus_msg.clone();
+        // Self-send is intentionally detached: the inner receive buffer can
+        // block, but network broadcast must still proceed.
+        tokio::spawn(async move {
```
varex83agent, line 207: The self-send is correctly detached (the inline comment captures why), but the spawn captures `task_ct`, `recv_tx.clone()`, `Arc<Sniffer>`, `Arc<msg::Msg>`, and a `QbftConsensusMsg`. It only completes when either `task_ct` is cancelled or `recv_tx.send` succeeds.

If the inner consumer stalls and the caller never cancels `ct` (e.g., `Transport` is dropped while its `ct` is still live elsewhere), each `broadcast` call piles up another task that lives until process exit. The `recv_tx` clone in particular keeps the channel's sender half alive even after `Transport` is dropped.

In Go this is mitigated by goroutines being cheap and the per-instance `ctx` being virtually guaranteed to cancel on instance teardown. In Rust, the contract `caller MUST cancel ct at instance teardown` is implicit. Either:
- document it on `Transport::new` / `broadcast`, or
- have `Transport` own an internal `CancellationToken` that's cancelled in `Drop`, so spawns can't outlive the transport even if the caller's token is leaked.

crates/core/src/consensus/qbft/transport.rs
```diff
@@ -0,0 +1,762 @@
...
+        (self.broadcaster)(ct, consensus_msg).await
+    }
+
+    /// Processes admitted outer messages until cancellation or channel close.
+    pub(crate) async fn process_receives(
+        &self,
+        ct: CancellationToken,
+        mut outer_rx: mpsc::Receiver<msg::Msg>,
+    ) -> Result<()> {
+        loop {
```
varex83agent, line 227: `process_receives` has two intentional behavioral divergences from Go (transport.go:148-164) that aren't called out in the docstring:

1. **Outer channel close → `Ok(())`** (lines 218-221). Go does `case msg := <-outerBuffer` without checking `ok`, so on channel close Go silently loops receiving zero-value `Msg{}` until ctx cancels. Rust's behavior is strictly better, but it's a divergence worth documenting.
2. **Inner-send failure → `Err(ReceiveBufferClosed)`** (line 231). Go's `recvBuffer <- msg` panics on closed channel; Rust returns an error variant, which is again better but diverges.

The docstring `Processes admitted outer messages until cancellation or channel close.` doesn't distinguish these two terminations. Recommend expanding to: returns `Ok(())` on cancellation or outer-channel close, and `Err(ReceiveBufferClosed)` if the inner buffer's receiver has been dropped — so future readers don't assume Go-style behavior.

crates/core/src/consensus/qbft/transport.rs
```diff
@@ -0,0 +1,762 @@
...
+            .values
+            .extend(
+                msg.values()
+                    .iter()
+                    .map(|(hash, value)| (*hash, value.clone())),
+            );
+    }
+
+    /// Returns a cached value by hash, after draining at most one local value.
+    pub(crate) fn get_value(&self, hash: [u8; 32]) -> Result<Any> {
```
varex83agent, line 132: `get_value` holds the `std::sync::Mutex` across `hash_proto_bytes(&local.value)` (line 131), which constructs an SSZ hasher, performs `put_bytes`, `merkleize`, and `hash_root`. For QBFT consensus values (unsigned data sets are KB-scale), this is non-trivial CPU work serialized against every concurrent `set_values` (called per inbound message in `process_receives`) and every other `get_value` (called per non-zero hash during `broadcast`).

Consider either:
- pulling the local value out under the lock, dropping the guard, hashing, then re-acquiring to insert; or
- splitting `value_rx` out of the `ValueStore` mutex (it's single-consumer in practice) so the hashmap and the channel are separately locked.

Not urgent given the message rate is low, but the previous commit `8515fd9 fix: single mutex for transport state` explicitly chose to consolidate — a brief comment on the rationale would help future maintainers decide whether to split it again.

crates/core/src/consensus/qbft/transport.rs
```diff
@@ -0,0 +1,762 @@
...
+            privkey: &self.privkey,
+        })?;
+        let msg = sync::Arc::new(msg);
+        let consensus_msg = msg.to_consensus_msg();
+        let inner_msg: qbft::Msg<ConsensusQbftTypes> = msg.clone();
+
+        let task_ct = ct.clone();
+        let recv_tx = self.recv_tx.clone();
+        let sniffer = sync::Arc::clone(&self.sniffer);
+        let sniffed_msg = consensus_msg.clone();
```
varex83agent, line 204: After commit `08b0edf fix: double clone consensus msg`, the broadcast path still performs two full clones of the consensus message:
1. `let consensus_msg = msg.to_consensus_msg();` builds a fresh `QbftConsensusMsg` (clones inner proto, justification protos, and every `Any` in `values`).
2. `let sniffed_msg = consensus_msg.clone();` clones it again so the spawned task can call `sniffer.add(sniffed_msg)`.

Then `consensus_msg` is moved into `(self.broadcaster)(ct, consensus_msg).await`.

The second clone exists only because the sniffer task needs ownership while the broadcaster also takes ownership. Either:
- wrap `consensus_msg` in `Arc<QbftConsensusMsg>` and adjust `Broadcaster` to take `Arc<QbftConsensusMsg>` (cheap clone shared by both paths), or
- have the spawned task call `msg.to_consensus_msg()` itself from the `Arc<msg::Msg>` it already holds.

The `values: Vec<Any>` cloned here can be kilobytes per message, so this is a real cost at consensus throughput, not a microbenchmark concern.

crates/core/src/consensus/qbft/transport.rs
varex83agent, line 138: nit: the function parameter `hash` is shadowed here — `let hash = msg::hash_proto_bytes(&local.value)?;` rebinds the name to the hash of the *just-drained local value*, which has nothing to do with the lookup key. The shadow correctly ends at the block's closing brace so line 135 still uses the parameter, but on first read it's confusing because a reader scanning the function may assume the lookup at line 135 uses the locally-computed hash.

Renaming the inner binding makes the intent obvious with zero behavioral change:
```rust
if let Ok(local) = store.value_rx.try_recv() {
    let local_hash = msg::hash_proto_bytes(&local.value)?;
    store.values.insert(local_hash, local);
}
```

crates/core/src/consensus/qbft/sniffer.rs
```diff
@@ -0,0 +1,105 @@
...
+            .lock()
+            .unwrap_or_else(PoisonError::into_inner)
+            .push(SniffedConsensusMsg {
+                timestamp: Some(Timestamp::from(SystemTime::now())),
+                msg: Some(msg),
+            });
+    }
+
+    /// Returns the buffered messages as a sniffed consensus instance.
+    pub(crate) fn instance(&self) -> SniffedConsensusInstance {
```
varex83agent, line 50: `Sniffer::instance` locks the mutex and clones the entire `Vec<SniffedConsensusMsg>` (each `SniffedConsensusMsg` carries a full `QbftConsensusMsg` with its `values: Vec<Any>`). Go's `Instance` (sniffer.go:46-57) aliases the slice (`Msgs: c.msgs`), which is racier but O(1).

If this is only called at instance end for the debug API, the deep clone is fine. The test helper `wait_for_sniffer_len` (transport.rs:657-665) busy-polls it, which is acceptable for tests but would be a perf cliff if anything in production polls similarly. Worth pinning the calling convention in the docstring (or, if frequent polling is expected, returning an `Arc<[...]>` or splitting `instance()` from `len()`).

crates/core/src/qbft/mod.rs
```diff
@@ -209,6 +209,11 @@ impl MessageType {
         Self(value)
     }
 
+    /// Returns the stable wire integer.
+    pub(crate) fn as_i64(self) -> i64 {
```
varex83agent: `as_i64` is borderline-idiomatic for a `Copy` newtype. Std convention reserves `as_*` for cheap reference-style views (`as_str`, `as_slice`, `as_bytes`). For extracting an inner `Copy` primitive, std uses either a `get(self)` method (`NonZeroI64::get`) or a `From` impl (`From<NonZeroU32> for u32`).

Since `MessageType(i64)` is already `Copy`, the more conventional options are:
```rust
// Option 1: From impl, symmetric to from_wire
impl From<MessageType> for i64 {
    fn from(value: MessageType) -> Self { value.0 }
}

// Option 2: getter named `get`
pub(crate) const fn get(self) -> i64 { self.0 }
```

Option 1 is the natural symmetric pair to `from_wire`, and it lets call sites at transport.rs:273 and 724 read as `i64::from(type_)` or `type_.into()`. Worth changing now while there's only a handful of call sites.

crates/core/src/consensus/qbft/transport.rs
```diff
@@ -0,0 +1,762 @@
+//! QBFT consensus transport adapter.
```
varex83agent, line 1: nit: the module doc is a one-liner while the companion `msg.rs:1-26` carries a detailed overview of the design (hash-vs-value separation, ValueMap reconciliation, justification wrapping, raw protobuf preservation).

`transport.rs` has comparable design subtleties worth surfacing at the module level rather than buried inline:
- the bridge between the tokio mpsc admission buffer and the crossbeam channel consumed by `core::qbft::run` (currently only the inline comment at lines 81-83),
- the rationale for the detached self-send + sniffer-add tokio task (currently the inline comment at lines 193-194),
- the contract that values referenced by hash MUST be cached before `broadcast` is called,
- how `process_receives` interacts with cancellation and channel close (see the divergence-doc comment below).

Would bring the module docs to parity with `msg.rs`.

crates/core/src/consensus/qbft/transport.rs
```diff
@@ -0,0 +1,762 @@
...
+
+/// Transport adapter for one QBFT consensus instance.
+pub(crate) struct Transport {
+    broadcaster: Broadcaster,
+    privkey: SecretKey,
+    // Async admission buffer for wrapped QBFT messages. Runner wiring bridges
+    // this tokio channel into the crossbeam receiver used by core::qbft::run.
+    recv_tx: mpsc::Sender<qbft::Msg<ConsensusQbftTypes>>,
+    sniffer: sync::Arc<Sniffer>,
+    values: Mutex<ValueStore>,
```
varex83agent, line 85: nit: field-name collision — `Transport.values: Mutex<ValueStore>` wraps a `ValueStore` whose own field is also called `values: ValueMap`. A reader of `self.values.lock()...values` has to remember that the outer one is a mutex around a struct that itself has a `values` field. Renaming one would help:
```rust
value_store: Mutex<ValueStore>,
// ...
struct ValueStore {
    value_rx: mpsc::Receiver<Any>,
    values: ValueMap,
}
```
This also matches the existing `value_rx` field's naming.

crates/core/src/consensus/qbft/transport.rs
```diff
@@ -0,0 +1,762 @@
...
+            value_hash,
+            prepared_round,
+            prepared_value_hash,
+            values,
+            justification: &justification,
+            privkey: &self.privkey,
+        })?;
+        let msg = sync::Arc::new(msg);
+        let consensus_msg = msg.to_consensus_msg();
+        let inner_msg: qbft::Msg<ConsensusQbftTypes> = msg.clone();
```
varex83agent, line 199: nit: this clones an `Arc` via method syntax (`msg.clone()` where `msg: Arc<msg::Msg>`), while the surrounding block (lines 191, and elsewhere in the file at 532/635/649) consistently uses the explicit `sync::Arc::clone(&...)` form. For consistency:
```rust
let inner_msg: qbft::Msg<ConsensusQbftTypes> = sync::Arc::clone(&msg);
```
Makes it obvious at the call site that this is a refcount bump, not a deep clone. Could also enable `clippy::clone_on_ref_ptr` if the project wants this enforced.

crates/core/src/consensus/qbft/transport.rs
```diff
@@ -0,0 +1,762 @@
...
+    /// Processes admitted outer messages until cancellation or channel close.
+    pub(crate) async fn process_receives(
+        &self,
+        ct: CancellationToken,
+        mut outer_rx: mpsc::Receiver<msg::Msg>,
+    ) -> Result<()> {
+        loop {
+            let msg = tokio::select! {
+                () = ct.cancelled() => return Ok(()),
+                msg = outer_rx.recv() => match msg {
```
varex83agent, line 230: nit: the `match` arm inside the `select!` is exactly what `let-else` was designed for, and the codebase already uses that idiom elsewhere:
```rust
let msg = tokio::select! {
    () = ct.cancelled() => return Ok(()),
    msg = outer_rx.recv() => msg,
};
let Some(msg) = msg else { return Ok(()); };
```
Subjective — feel free to ignore.

crates/core/src/consensus/qbft/transport.rs
```diff
@@ -0,0 +1,762 @@
...
+        let value_hash = value_hash(1);
+        let msg = wrapped_msg(qbft::MSG_PRE_PREPARE, value_hash, [0u8; 32], vec![]);
+
+        transport.set_values(&msg);
+
+        assert_eq!(transport.get_value(value_hash).unwrap(), any_timestamp(1));
+    }
+
+    #[test]
+    fn get_value_returns_cached_value() {
```
varex83agent, line 339: nit: `set_values_caches_message_values` (lines 316-324) and `get_value_returns_cached_value` (lines 327-334) are functionally identical — both build a `Msg`, call `set_values`, then assert `get_value` returns the cached `Any`. They differ only in name. One should either be removed or re-purposed to exercise a distinct path (e.g., overwriting an existing key, or `set_values` with multiple hashes that have to round-trip through `get_value`).

crates/core/src/consensus/qbft/transport.rs
```diff
@@ -0,0 +1,762 @@
...
+        assert_eq!(received.value(), value_hash);
+        assert_eq!(transport.get_value(value_hash).unwrap(), any_timestamp(1));
+        assert_eq!(transport.sniffer_instance().msgs.len(), 1);
+
+        ct.cancel();
+        runner.await.unwrap().unwrap();
+    }
+
+    #[tokio::test]
+    async fn process_receives_stops_on_cancel() {
```
varex83agent, line 564: nit: missing test coverage for two documented invariants:

1. **Detached self-send doesn't block external broadcast** (lines 193-194 promise this). No test fills `recv_tx` to capacity before calling `broadcast` and asserts the broadcaster still completes. A regression that accidentally awaited the self-send in-line wouldn't be caught.
2. **Cancel-during-inner-send branch of `process_receives`** (the inner `tokio::select!` at lines 228-234 has a `ct.cancelled()` arm). This test cancels BEFORE entering the loop, exercising only the outer select. A test that fills the inner buffer, sends one outer message, waits until `set_values` has recorded the value, then cancels, would pin this branch.

crates/core/src/qbft/mod.rs
mskrzypkows: This seems relevant

mskrzypkows (approved):
Looks good

varex83 (approved):
One comment regarding `From<...>` from agent good to have, other than that LGTM

PR #449, opened 2026-05-28
feat(core): implement validatorapi node_version and proposer_duties handlers

varex83:
## Summary

First two slices of the validatorapi port. Lands the router state plumbing and two handlers end-to-end so the next handler PRs can drop straight in.

- `AppState<H>` + `Arc<AppState<H>>` threaded through `new_router` via `with_state`.
- `node_version` and `proposer_duties` routes implemented against the `Handler` trait.
- Re-uses the auto-generated `pluto_eth2api` envelopes (`GetVersionResponseResponse`, `GetProposerDutiesResponseResponse`) as the on-the-wire shape — no hand-rolled parallel types. The `validatorapi/types.rs` module re-exports them under shorter names (`NodeVersionResponse`, `ProposerDutiesResponse`, `ProposerDuty`).
- New `testutils.rs` with `TestHandler` — implements all 19 `Handler` methods (2 real, 17 `unimplemented!()` stubs); future handler PRs override the relevant method to drive their unit test.

This sets the pattern for the remaining handler PRs: each method returns its endpoint's spec-precise auto-gen envelope rather than the generic `EthResponse<T>` wrapper, which will be retired per-PR as handlers land. Middleware (timeout, content-type validation, metrics) and the reverse-proxy fallback come in their own PRs after a few more handlers land.

## Test plan

- [x] `cargo check -p pluto-core --tests`
- [x] `cargo +nightly fmt --all -- --check`
- [x] `cargo test -p pluto-core validatorapi::router::tests` — `node_version_wraps_handler_value` and `proposer_duties_wraps_handler_value` pass


crates/core/src/validatorapi/body.rs
```diff
@@ -0,0 +1,17 @@
...
+    /// Version payload.
+    pub data: NodeVersionData,
+}
+
+/// `data` field of [`NodeVersionResponse`].
+#[derive(Debug, Clone, Serialize)]
+pub struct NodeVersionData {
+    /// Node version string.
+    pub version: String,
+}
```
emlautarom1: Already exists under `eth2api` as `GetNodeVersionRequest` and associated types.

emlautarom1 (changes requested):
Small stuff:

- The added DTOs already exist in `eth2api`: https://ethereum.github.io/beacon-APIs/#/Node/getNodeVersion
- The type parameters could get tricky down the road so we could use `Box<dyn ...>` instead.

crates/core/src/validatorapi/body.rs
varex83: I've realized that in the https://github.com/NethermindEth/pluto/pull/450

emlautarom1 (approved):

PR #451, opened 2026-05-28
feat(core): implement attester_duties, sync_committee_duties, attestation_data handlers

varex83:
## Summary

Implements three validator-API handlers, stacked over PR #449.

- **attester_duties** (`POST /eth/v1/validator/duties/attester/{epoch}`) — uses auto-gen `GetAttesterDuties*` envelope, accepts both numeric and string-encoded validator index arrays via a new `ValIndexes` dual-format deserializer, performs pubshare swap on the upstream response (unknown pubkey rejected).
- **sync_committee_duties** (`POST /eth/v1/validator/duties/sync/{epoch}`) — same shape as above, reuses `ValIndexes`.
- **attestation_data** (`GET /eth/v1/validator/attestation_data`) — query-string `slot`/`committee_index`, served from the local DutyDB (`MemDB::await_attestation`) rather than upstream. Component now holds an `Arc<MemDB>`.

All three return the auto-generated response envelope directly; `attestation_data` uses a small hand-rolled wrapper around `phase0::AttestationData` to avoid pointless typed-↔-string conversion.

## Test plan

- [x] `cargo test -p pluto-core validatorapi::` — 14/14 passing (5 new tests across router + component)


varex83:
@varex83agent /review-pr

crates/core/src/validatorapi/component.rs
```diff
@@ -144,23 +156,107 @@ impl Handler for Component {
...
     async fn attestation_data(
         &self,
-        _opts: AttestationDataOpts,
-    ) -> Result<EthResponse<AttestationData>, ApiError> {
-        unimplemented!("attestation_data not yet ported")
+        opts: AttestationDataOpts,
+    ) -> Result<AttestationDataResponse, ApiError> {
+        let data = self
+            .dutydb
+            .await_attestation(opts.slot, opts.committee_index)
```
varex83agent: **bug**: `await_attestation` blocks indefinitely with no timeout, no request-cancel propagation, and no eviction wakeup.

`MemDB::await_data` (`crates/core/src/dutydb/memory.rs:377-402`) selects only on a `Notified` future and the global `cancel` token (process shutdown). Three compounding issues:

1. **No timeout.** A `GET /eth/v1/validator/attestation_data?slot=<never-stored>&committee_index=...` future parks until process shutdown.
2. **No HTTP-client-disconnect propagation.** axum/hyper do not automatically cancel handler futures when the TCP peer closes — and even if they did, the handler doesn't poll any per-request cancel signal.
3. **`delete_duty` does not call `notify_waiters`.** When the deadliner evicts an expired attestation duty, waiters for that slot get no signal — they re-block forever after every subsequent `store` for any other slot wakes them, checks the map, and finds nothing.

Blast radius: each waiter holds a tokio task, a `Notified` future, and an `Arc<MemDB>`. Worse, every legitimate `store` calls `attestation_notify.notify_waiters()` which wakes *every* outstanding waiter — including any malicious ones — each of whom re-acquires `RwLock<State>` (read) and re-checks. N waiters × M legitimate stores = O(N·M) lock contention with the write side of `store`.

Charon (`charon/core/validatorapi/validatorapi.go:262-269`) wires the request `ctx` through `awaitAttFunc`, and `MemDB.AwaitAttestation` (`charon/core/dutydb/memory.go:178-205`) selects on `<-ctx.Done()`. Pluto loses this entirely.

Fix: wrap the await in `tokio::time::timeout(...)` bounded by the slot's wall-clock deadline, AND either (a) plumb a per-request `CancellationToken` derived from the HTTP request, or (b) make `delete_duty` call `notify_waiters` and `await_data` return `Err(Error::DutyExpired)` when the awaited key is evicted.

crates/core/src/validatorapi/component.rs
```diff
@@ -144,23 +156,107 @@ impl Handler for Component {
...
+                ApiError::new(StatusCode::BAD_GATEWAY, "upstream attester duties failed")
+                    .with_source(std::io::Error::other(err.to_string()))
+            })?;
+
+        let mut payload = match response {
+            GetAttesterDutiesResponse::Ok(payload) => payload,
+            other => {
+                return Err(ApiError::new(
+                    StatusCode::BAD_GATEWAY,
+                    format!("unexpected upstream attester duties response: {other:?}"),
```
varex83agent: **bug**: `format!("... response: {other:?}")` leaks upstream beacon-node response internals into the client-visible error body.

`other` is a `GetAttesterDutiesResponse` whose non-`Ok` variants wrap `BlindedBlock400Response`, defined (`crates/eth2api/src/types.rs`) as:
```rust
pub struct BlindedBlock400Response {
    pub code: f64,
    pub message: String,
    pub stacktraces: Option<Vec<String>>,
}
```
The `{other:?}` expansion includes the upstream's raw `message` and any `stacktraces` (which beacon nodes emit in debug mode, often with file paths and internal context). This string is assigned to `ApiError.message`, and `IntoResponse` serializes it directly to the HTTP body as `{ "code": ..., "message": "..." }`.

Result: a 503/500/400 from upstream is forwarded verbatim — including stacktraces — to the validator-API caller, fingerprinting the upstream beacon-node version and exposing internals to anyone who can reach Pluto's validator API.

Same pattern at line 233 (`sync_committee_duties`) and line 147 (`proposer_duties`, from the prior PR).

Fix: use a generic `message` (e.g. `"unexpected upstream attester duties response"`) and move the `{other:?}` expansion into `with_source(...)` — which `ApiError`'s response writer does not serialize.

crates/core/src/validatorapi/component.rs
```diff
@@ -77,9 +84,14 @@ impl Component {
 
     /// Builds a component that skips partial-signature verification on
     /// submit endpoints. Test use only.
-    pub fn new_insecure(eth2_cl: Arc<EthBeaconNodeApiClient>, share_idx: u64) -> Self {
+    pub fn new_insecure(
```
varex83agent, line 105: **major**: `Component::new_insecure` is plain `pub fn`, not `#[cfg(test)]`. It sets `insecure_test: true`, which (per the doc on lines 61-63) disables partial-signature verification on the submit endpoints landing in later PRs.

There is no compile-time barrier preventing a future caller from constructing an insecure `Component` in production code — by mistake during refactor, or because someone treats "insecure" as a debug-only flag.

Gate this behind `#[cfg(any(test, feature = "insecure-test"))]`, or move it into the `testutils` submodule. If it needs to remain `pub` for cross-crate tests, at minimum add a runtime guard (env var / feature) that refuses to enable `insecure_test` outside of a test build.

crates/core/src/validatorapi/router.rs
```diff
@@ -120,8 +132,20 @@ pub fn new_router(handler: Arc<dyn Handler>, builder_enabled: bool) -> Router {
         .with_state(state)
 }
 
-async fn attester_duties() {
-    todo!("vapi: attester_duties");
+async fn attester_duties(
```
varex83agent, line 146: **major**: `new_router` (defined above, line 49) installs no `DefaultBodyLimit`, no `TimeoutLayer`, and no concurrency cap on an attacker-reachable HTTP surface — and this handler (plus `sync_committee_duties`, `attestation_data`) is where that gap matters most.

- **Body limit**: axum 0.8 defaults to 2 MiB. A 2 MiB JSON array of numeric indices (`ValIndexes::Numbers`) parses into hundreds of thousands of `u64`s, each then reformatted as a `String` (types.rs:212). The `#[serde(untagged)]` retry on the `Strings` variant additionally buffers the parsed JSON into serde's `Content` cache, so peak per-request allocation is ~3× body size.
- **Timeout**: combined with the unbounded `await_attestation` (see other comment), a slow/malicious client can hold handler tasks indefinitely.
- **Concurrency**: nothing caps simultaneous in-flight requests.

Minimum: attach `.route_layer(DefaultBodyLimit::max(64 * 1024))` on the two `POST /duties/*` routes (64 KiB still allows ~10k indices, well beyond any realistic cluster) and a `TimeoutLayer` of ~30s on the validator routes (excluding the proxy fallback). Anchored here on the new attester_duties handler since `new_router` itself isn't in the diff.

crates/core/src/validatorapi/component.rs
```diff
@@ -144,23 +156,107 @@ impl Handler for Component {
...
+            .await
+            .map_err(|err| {
+                ApiError::new(StatusCode::BAD_GATEWAY, "upstream attester duties failed")
+                    .with_source(std::io::Error::other(err.to_string()))
+            })?;
+
+        let mut payload = match response {
+            GetAttesterDutiesResponse::Ok(payload) => payload,
+            other => {
+                return Err(ApiError::new(
```
varex83agent: **major**: `GetAttesterDutiesResponse` has discriminated variants for `BadRequest(400)`, `InternalServerError(500)`, `ServiceUnavailable(503)`, and `Unknown`. The match here collapses all of them under `StatusCode::BAD_GATEWAY` (502).

- A 400 from upstream means *our forwarded request* (epoch or indices) was invalid — 502 hides that from the VC.
- A 503 (`syncing`) deserves to propagate as 503 so the VC retries with backoff, not 502.
- An `Unknown` variant arguably justifies 502, but the others don't.

Same pattern repeats in `sync_committee_duties` (line 231) and `proposer_duties` (line 145, prior PR). Recommend matching on each variant and mapping to the faithful status — e.g. `BadRequest → 400`, `InternalServerError → 502`, `ServiceUnavailable → 503`, `Unknown → 502`.

crates/core/src/validatorapi/component.rs
```diff
@@ -334,17 +527,100 @@ mod tests {
...
+        let response = component
+            .attestation_data(AttestationDataOpts {
+                slot: SLOT,
+                committee_index: COMM_IDX,
+            })
+            .await
+            .unwrap();
+        assert_eq!(response.data.slot, SLOT);
+        assert_eq!(response.data.index, COMM_IDX);
+        assert_eq!(response.data.beacon_block_root, [0x11; 32]);
```
varex83agent, line 747: **major**: Test coverage gaps for the three new handlers. The added tests cover the happy path plus swap-helper edge cases (unknown pubkey, malformed pubkey). Not covered:

1. **Cancellation/timeout of `attestation_data`** — no test verifies that dropping the `await_attestation` future releases the dutydb waiter, and no test asserts a timeout bound.
2. **Upstream non-`Ok` variants** — `GetAttesterDutiesResponse::{BadRequest, InternalServerError, ServiceUnavailable, Unknown}` (and the sync-committee equivalents) are all collapsed to `BAD_GATEWAY` but no test exercises any of them.
3. **Malformed query params** for `attestation_data` — `?slot=foo`, missing `committee_index`, extra fields. Axum's error path is verified nowhere.
4. **Empty `indices` body** — no test for `POST /duties/attester/{epoch}` with `[]`, nor for mixed `[1,"2"]` (currently rejected by `ValIndexes`'s untagged enum, but no assertion).
5. **Negative `attestation_data` lookup** — store data at `(SLOT, COMM_IDX)` and confirm the handler does NOT resolve when asked for `(SLOT, COMM_IDX+1)` (with a short test-timeout, to confirm the keyed lookup behavior).

Addressing these would also lock in the fixes for the bug-level findings above.

crates/core/src/validatorapi/component.rs
```diff
@@ -144,23 +156,107 @@ impl Handler for Component {
...
-        unimplemented!("attestation_data not yet ported")
+        opts: AttestationDataOpts,
+    ) -> Result<AttestationDataResponse, ApiError> {
+        let data = self
+            .dutydb
+            .await_attestation(opts.slot, opts.committee_index)
+            .await
+            .map_err(|err| {
+                ApiError::new(
+                    StatusCode::INTERNAL_SERVER_ERROR,
```
varex83agent: **minor**: `await_attestation` failure is unconditionally mapped to `INTERNAL_SERVER_ERROR`. `MemDB::await_attestation` errors are limited to `Error::Shutdown` today (process shutdown), but once a per-request timeout / duty-expiry signal is added (see other comment), expired-duty and timeout should map to `408 Request Timeout` (or `504 Gateway Timeout`), and `Shutdown` should map to `503 Service Unavailable`. Charon's `writeError` (`charon/core/validatorapi/router.go:1674-1691`) explicitly maps `ctx.Err()` to 408.

crates/core/src/validatorapi/component.rs
```diff
@@ -272,6 +368,44 @@ fn swap_proposer_pubshares(
...
+/// pubkey indicates a misconfiguration and is rejected.
+fn swap_attester_pubshares(
+    duties: &mut [AttesterDuty],
+    pub_share_by_pubkey: &HashMap<BLSPubKey, BLSPubKey>,
+) -> Result<(), ApiError> {
+    for duty in duties {
+        let pubkey = parse_bls_pubkey(&duty.pubkey)?;
+        let share = pub_share_by_pubkey.get(&pubkey).ok_or_else(|| {
+            ApiError::new(
+                StatusCode::BAD_GATEWAY,
```
varex83agent: **minor**: `BAD_GATEWAY` (502) for "pubshare not found" misclassifies a Pluto/cluster configuration error as an upstream failure. The upstream returned a well-formed duty; the problem is that *this Pluto node doesn't have a share for the validator the upstream named*. That's either operator-supplied indices for non-cluster validators (`400 BAD_REQUEST`) or a misconfigured lock file (`500 INTERNAL_SERVER_ERROR`) — not an upstream-gateway failure.

Same applies at line 400 in `swap_sync_committee_pubshares`.

Note: a `parse_bls_pubkey` failure at lines 379/397 mapping to 502 is fine — that one is genuinely an upstream-format issue.

Charon doesn't set a specific status here (just `errors.New("pubshare not found")`, which maps to 500), so changing to 500 (or 400) would also be more parity-faithful.

crates/core/src/validatorapi/types.rs
```diff
@@ -179,3 +185,38 @@ pub struct BeaconCommitteeSelection {}
...
+            Numbers(Vec<u64>),
+            Strings(Vec<String>),
+        }
+
+        let value = Either::deserialize(deserializer)?;
+        let indices = match value {
+            Either::Numbers(ns) => ns.into_iter().map(|n| n.to_string()).collect(),
+            Either::Strings(strs) => {
+                for s in &strs {
+                    s.parse::<u64>().map_err(serde::de::Error::custom)?;
```
varex83agent: **minor**: The `Strings` arm fully materializes `Vec<String>` before validating any element. Combined with `#[serde(untagged)]` first speculatively parsing as `Vec<u64>` (which fails for string input but buffers a serde `Content` cache), peak per-request allocation is roughly 3× body size.

An attacker sending `["","",…]` or `["x","x",…]` forces N short-string allocations before validation rejects. Mitigations: (a) use a custom `Visitor` that streams elements via `SeqAccess::next_element`, validating each as it's read; (b) impose a hard cap on `indices.len()` (a real beacon-API client sends at most a few hundred); (c) attach a tight `DefaultBodyLimit` on the route (see router.rs:49 comment).

crates/core/src/validatorapi/component.rs
```diff
@@ -144,23 +156,107 @@ impl Handler for Component {
...
+                ApiError::new(StatusCode::BAD_REQUEST, "invalid attester duties request")
+                    .with_source(std::io::Error::new(
+                        std::io::ErrorKind::InvalidInput,
+                        err.to_string(),
+                    ))
+            })?;
+
+        let response = self
+            .eth2_cl
+            .get_attester_duties(request)
```
varex83agent: **minor**: Upstream `eth2_cl` calls (`get_attester_duties` here, `get_sync_committee_duties` at line 218, `get_proposer_duties` at line 134 from prior PR) have no `tokio::time::timeout` wrapper. If the upstream beacon node hangs (TCP accepted, no HTTP response), the Pluto handler hangs as long as reqwest's default timeout permits — typically unbounded unless configured on the shared `Client`. Combined with the absence of a router-level `TimeoutLayer`, an attacker who can influence the upstream gets a DoS lever.

Wrap each call in `tokio::time::timeout(...)`, or configure a request timeout on the shared `reqwest::Client` used by `EthBeaconNodeApiClient`.

crates/core/src/validatorapi/component.rs
```diff
@@ -144,23 +156,107 @@ impl Handler for Component {
...
-        unimplemented!("attester_duties not yet ported")
+        opts: AttesterDutiesOpts,
+    ) -> Result<AttesterDutiesResponse, ApiError> {
+        let request = GetAttesterDutiesRequest::builder()
+            .epoch(opts.epoch.to_string())
+            .body(opts.indices)
+            .build()
+            .map_err(|err| {
+                ApiError::new(StatusCode::BAD_REQUEST, "invalid attester duties request")
+                    .with_source(std::io::Error::new(
```
varex83agent: **minor**: `ApiError::with_source<E: std::error::Error + Send + Sync + 'static>` accepts any `Error`, so wrapping the builder error in `std::io::Error::new(std::io::ErrorKind::InvalidInput, err.to_string())` is ceremony — and it stringifies the cause, breaking `Error::source()` chaining and losing any wrapped context. Just pass `err` directly:
```rust
.with_source(err)
```
Same pattern repeats at lines 179, 211, 225, and 256.

crates/core/src/validatorapi/component.rs
```diff
@@ -334,17 +527,100 @@ mod tests {
         assert_eq!(err.status_code, StatusCode::BAD_GATEWAY);
     }
 
-    #[tokio::test]
-    async fn node_version_formats_pluto_string() {
-        // Use an unreachable upstream — node_version doesn't call it.
+    use chrono::{DateTime, Utc};
```
varex83agent: **minor**: The `tests` module already has `use super::*;` at line 431, then test functions, then a second block of imports here (`chrono`, `tokio::sync::mpsc`, `tokio_util::sync::CancellationToken`, then the `crate::...` block at lines 534-543). Pluto's style convention (per the project rust-style guide) is to put all `use` declarations at the top of a module before any items. Hoist these up next to `use super::*;`.

crates/core/src/validatorapi/router.rs
```diff
@@ -7,17 +7,29 @@ use std::sync::Arc;
...
+    types::{
+        AttestationDataOpts, AttestationDataResponse, AttesterDutiesOpts, AttesterDutiesResponse,
+        CommitteeIndex, NodeVersionResponse, ProposerDutiesOpts, ProposerDutiesResponse,
+        SyncCommitteeDutiesOpts, SyncCommitteeDutiesResponse, ValIndexes,
+    },
 };
 
+/// Query parameters for `GET /eth/v1/validator/attestation_data`.
+#[derive(Debug, Clone, Deserialize)]
+struct AttestationDataQuery {
```
varex83agent, line 39: **minor**: `Query<AttestationDataQuery>` produces axum's default `400 Bad Request` (plain text) for missing or malformed `slot` / `committee_index`. Every other error path in this router returns Pluto's `ApiError { code, message }` JSON envelope (and Charon's `uintQuery` similarly emits a structured 400). This inconsistency means VC error parsers will see two distinct error body shapes from the same endpoint.

Fix: extract the query manually (e.g. `Result<Query<AttestationDataQuery>, QueryRejection>` plus a manual map_err into `ApiError`) so all 4xx paths return the same shape.

crates/core/src/validatorapi/component.rs
```diff
@@ -272,6 +368,44 @@ fn swap_proposer_pubshares(
...
+fn swap_attester_pubshares(
+    duties: &mut [AttesterDuty],
+    pub_share_by_pubkey: &HashMap<BLSPubKey, BLSPubKey>,
+) -> Result<(), ApiError> {
+    for duty in duties {
+        let pubkey = parse_bls_pubkey(&duty.pubkey)?;
+        let share = pub_share_by_pubkey.get(&pubkey).ok_or_else(|| {
+            ApiError::new(
+                StatusCode::BAD_GATEWAY,
+                "pubshare not found for attester duty",
```
varex83agent, line 504: **minor**: `swap_attester_pubshares` (line 380-385) and `swap_sync_committee_pubshares` (line 398-403) act as validator-set-membership oracles. An attacker who can POST to `/eth/v1/validator/duties/attester/{epoch}` with chosen `indices` learns from the response code (200 vs 502) which validators belong to this cluster.

This matches Charon parity exactly (`charon/core/validatorapi/validatorapi.go:1167-1190` errors the same way), so it is not a regression — but worth flagging if the validator API is ever exposed beyond the local validator client. Hardening options (none implemented in either client): (a) filter `indices` against `pub_share_by_pubkey` before calling upstream; (b) return 200 with an empty `data` array instead of erroring.

crates/core/src/validatorapi/component.rs
```diff
@@ -334,17 +527,100 @@ mod tests {
...
+    }
+
+    /// Build a Component backed by a real (but never-expiring) DutyDB plus a
+    /// dummy upstream client. Useful for tests that only exercise endpoints
+    /// served from the DB.
+    fn make_test_component() -> (Component, Arc<MemDB>) {
+        let cancel = CancellationToken::new();
+        let (deadliner, _deadliner_rx) =
+            DeadlinerTask::start(cancel.clone(), "validatorapi-tests", FarFutureCalculator);
+        let (_unused_tx, evict_rx) = mpsc::channel(1);
```
varex83agent: nit: `_unused_tx` reads as leftover. Its purpose is to keep the eviction channel's sender alive so `evict_rx` doesn't observe a closed channel during the test — that intent is clearer as `_evict_tx` (mirroring `evict_rx` two lines down).

varex83agent (changes requested):
Implements three validator-API handlers (`attester_duties`, `sync_committee_duties`, `attestation_data`) with reasonable parity to Charon and unit tests for the swap helpers, the dual-format `ValIndexes` deserializer, and the DutyDB-backed attestation path. Code style is idiomatic. However, two bug-level issues need fixing before merge, plus several major hardening gaps on this attacker-reachable surface.

**Bugs (must fix):**
1. `attestation_data` awaits `MemDB` with no timeout, no request-cancel propagation, and no eviction signal — a request for a slot that never arrives becomes a permanent zombie task, contending on every legitimate `store` via `notify_waiters`. Charon plumbs the request `ctx` through `awaitAttFunc`; the Rust port loses this entirely. Wrap the await in a `tokio::time::timeout` (matching the slot deadline) and either wake on `delete_duty` or plumb a per-request `CancellationToken`.
2. `format!("unexpected upstream … response: {other:?}")` interpolates the upstream beacon node's Debug output — including `BlindedBlock400Response.message` and any `stacktraces` field — directly into the client-visible `ApiError.message`. This forwards upstream internals (and possibly file paths) to anyone reachable on the validator API. Move `{other:?}` into `with_source(...)` and use a generic `message`.

**Majors:**
- `Component::new_insecure` (which disables submit-side signature verification) is plain `pub fn`, not `#[cfg(test)]` — once submit endpoints land, this is a footgun.
- `new_router` installs no `DefaultBodyLimit`, no `TimeoutLayer`, and no concurrency cap. Combined with the unbounded await above, slowloris-style requests are trivially possible.
- All non-`Ok` upstream variants (`BadRequest` / `InternalServerError` / `ServiceUnavailable` / `Unknown`) are collapsed under 502. A syncing upstream (503) should propagate as 503 so the VC retries with backoff; a 400 from upstream means our forwarded request was bad.
- Test coverage is missing for: cancellation/timeout on `attestation_data`, the upstream non-Ok variants, malformed `slot`/`committee_index` query params, and empty `indices` body.

Minors and a nit inline. Verdict: **REQUEST_CHANGES** — please address the two bug-level findings before merge; majors can land in a follow-up or this PR at your discretion.

varex83:
@varex83agent /review-pr

crates/core/src/dutydb/memory.rs
```diff
@@ -190,6 +207,18 @@ struct State {
     contrib_duties: HashMap<ContribKey, altair::SyncCommitteeContribution>,
     contrib_keys_by_slot: HashMap<u64, Vec<ContribKey>>,
 
+    /// Slots whose attester duty has been evicted by the deadliner. Lets
+    /// `await_attestation` return `AwaitDutyExpired` immediately when the
+    /// awaited slot is gone, rather than spinning on every `store()` until
+    /// the request-level timeout fires.
+    evicted_attestation_slots: HashSet<u64>,
```
varex83agent: **Unbounded memory growth.** The four eviction sets (`evicted_attestation_slots`, `evicted_proposer_slots`, `evicted_aggregation_keys`, `evicted_contrib_keys`) are only ever `insert`/`extend`-ed in `delete_duty` (lines 666/679/686/694) and `contains`-checked in the `await_*` lookups — nothing ever removes or clears them (verified: no `.remove`/`.clear`/`.retain` references them). Since slots advance monotonically, every evicted slot/key is retained for the whole process lifetime: ~7200 attester-slot entries/day plus aggregation/contrib keys (each carrying a 32-byte root), forever. This is a real unbounded leak in a long-running node and a divergence from Charon v1.7.1, whose `MemDB` keeps no evicted set at all — it relies on the per-request `ctx` deadline to unblock parked queries.

The sets were added purely so `await_*` can return `AwaitDutyExpired` early instead of spinning until the request timeout; that optimization needs a bound. Options: track a single `min_live_slot` high-water-mark and treat `slot < min_live_slot && !present` as evicted (works directly for the slot-keyed sets; aggregation/contrib are keyed by root but `*_keys_by_slot` already gives the slot for pruning), or prune entries below a retention horizon whenever `delete_duty` runs.

Secondary point on the same mechanism: the `Proposer` (666) and `Attester` (679) arms `insert(slot)` unconditionally, even when no duty was ever stored for that slot, whereas the `Aggregator`/`SyncContribution` arms only `extend` when keys existed — so proposer/attester entries accumulate even faster than necessary.

crates/core/src/dutydb/memory.rs
```diff
@@ -333,8 +393,16 @@ impl MemDB {
             slot,
             committee_index,
         };
-        self.await_data(&self.attestation_notify, |s| s.attestation_duties.get(&key))
-            .await
+        self.await_data(&self.attestation_notify, |s| {
+            if let Some(v) = s.attestation_duties.get(&key) {
+                Lookup::Found(v.clone())
+            } else if s.evicted_attestation_slots.contains(&key.slot) {
```
varex83agent: **Missing `committee_index=0` dual-write — breaks post-Electra VCs.** This await keys strictly on `(slot, committee_index)`, which is correct given how data is stored — but `store_attestation` (memory.rs:535-543) writes the duty only under the duty's real committee index. Charon's `storeAttestationUnsafe` (core/dutydb/memory.go @ v1.7.1, lines 343-371) deliberately stores each attestation **twice**: once under the real index and once under a hardcoded `committee_index = 0`, with an explicit comment that post-Electra VCs request `attestation_data` with `committee_index=0` and need the duplicate key to resolve.

Before this PR `attestation_data` was `unimplemented!`; now that it serves live from the DutyDB, a conformant Electra VC requesting `?committee_index=0` against a duty stored under a non-zero index will never match and will hit the 24s timeout → 408, unable to attest. The root cause is in pre-existing store logic (outside this diff), but it's this PR that makes the gap user-visible. Please port the dual-write (or confirm the fetcher normalizes `committee_index` to 0 before storing). Note `attestation_data_does_not_resolve_for_wrong_committee_index` would still pass with the dual-write, so it doesn't cover this case.

crates/core/src/dutydb/memory.rs
```diff
@@ -302,24 +332,54 @@ impl MemDB {
...
+        // lock is harmless — woken tasks block on `state.read()` until we
+        // drop.
+        self.wake(duty.duty_type);
+
+        // Drain all expired duties that the deadliner has sent. Waiters
+        // whose duty just expired need to see `Lookup::Evicted` and exit,
+        // not re-park — so we wake the matching notify after each eviction.
         while let Ok(expired) = state.deadliner_rx.try_recv() {
+            let duty_type = expired.duty_type.clone();
             state.delete_duty(expired)?;
```
varex83agent, line 358: nit: `state.delete_duty(expired)?` propagates errors out of the drain loop. `delete_duty` returns `Err` for `BuilderProposer` (668) and the catch-all `UnknownDutyType` arm (697). If such a duty ever reaches the deadliner output, `store()` returns early — abandoning any further already-dequeued evictions while the just-stored duty's effects remain committed. In practice the deadliner only schedules duty types it can compute a deadline for, so this shouldn't trigger, but a single poisoned queue entry can wedge the whole drain. Prefer logging and `continue` over `?` here.

crates/core/src/validatorapi/router.rs
```diff
@@ -120,8 +139,20 @@ pub fn new_router(handler: Arc<dyn Handler>, builder_enabled: bool) -> Router {
         .with_state(state)
 }
 
-async fn attester_duties() {
-    todo!("vapi: attester_duties");
+async fn attester_duties(
+    State(state): State<Arc<AppState>>,
+    Path(epoch): Path<u64>,
```
varex83agent, line 148: An invalid `{epoch}` path segment (e.g. `/duties/attester/foo`) extracted via `Path(epoch): Path<u64>` produces axum's **default** `PathRejection` — a plain-text 400 body — not the `{ code, message }` `ApiError` envelope the rest of the router emits. The status (400) matches Charon's `uintParam` (router.go), but the body shape doesn't. Note `attestation_data` was deliberately given `Result<Query<...>, QueryRejection>` + `query_rejection_to_api_error` (lines 186-190, 217) to normalize exactly this; the three `{epoch}` handlers (attester/proposer/sync, lines 144/160/172) weren't. Consider `Result<Path<u64>, PathRejection>` + a mapper for envelope parity.

crates/core/src/validatorapi/types.rs
```diff
@@ -179,3 +190,100 @@ pub struct BeaconCommitteeSelection {}
 /// Sync-committee selection payload. Placeholder.
 #[derive(Debug, Clone)]
 pub struct SyncCommitteeSelection {}
+
+/// Validator-index request body for the `attester_duties` and
+/// `sync_committee_duties` endpoints.
+///
+/// Accepts both numeric (`[1, 2]`) and string-encoded (`["1", "2"]`) JSON
```
varex83agent, line 197: Parity note: Charon's `valIndexesJSON.UnmarshalJSON` is two-pass — it decodes the whole body as `[]uint64` and only on failure retries as `[]string`, so a mixed array like `[1, "2"]` fails **both** passes and is rejected. This `ValIndexes` visitor validates each element independently and accepts `[1, "2", 3]` (see `val_indexes_accepts_mixed_elements`). That's a strict *superset* of Charon's behavior — more permissive, so it can't break a conforming VC — but it is a documented divergence from the "functional equivalence" baseline. Either tighten to Charon's all-or-nothing format or note the intentional superset in the doc comment. (The per-element u64 validation, negative rejection, and empty-array acceptance all match Charon — only the mixed case differs.)

crates/core/src/validatorapi/component.rs
```diff
@@ -256,6 +394,81 @@ impl Handler for Component {
...
+        format!("upstream {endpoint} failed"),
+    )
+    .with_boxed_source(err)
+}
+
+/// Builds the `ApiError` returned when the upstream responds with a faithful
+/// HTTP status that we propagate (e.g. 400, 503). The upstream body is
+/// attached as a `source` for debug logging — never serialized into the
+/// client-visible message.
+fn upstream_status_error<B: std::fmt::Debug>(
```
varex83agent, line 425: `upstream_status_error` (424) and `upstream_unexpected` (442) — and `query_rejection_to_api_error` in router.rs:217 — wrap a formatted `{body:?}` string in `std::io::Error::other(...)` purely to satisfy `with_source`'s `E: std::error::Error` bound. This fabricates an `io::Error` with a misleading `ErrorKind::Other` for something that was never an I/O failure. The no-leak behavior is correct (verified: `into_response` serializes only `code`/`message`), but the carrier is a smell. Cleaner: add an `ApiError::with_debug_source(impl Into<String>)` that boxes a purpose-built `#[derive(thiserror::Error)] struct DebugString(String)`, or let `with_boxed_source` take the string directly. (The `with_source` vs `with_boxed_source` split itself is fine — the latter is justified for `anyhow::Error`, which isn't `Error`.)

crates/core/src/validatorapi/component.rs
```diff
@@ -256,6 +394,81 @@ impl Handler for Component {
...
+        "upstream {endpoint} variant: {response:?}"
+    )))
+}
+
+/// Maps a [`crate::dutydb::Error`] into the `ApiError` returned to the client
+/// when an `attestation_data` await fails. `Shutdown` propagates as 503 so the
+/// VC can retry; `AwaitDutyExpired` propagates as 408 — same as a timeout —
+/// since the duty is gone and the data will never arrive. Anything else is a
+/// programming error here and becomes 500.
+fn map_dutydb_error(err: DutyDbError) -> ApiError {
```
varex83agent, line 458: nit: This PR threads upstream bodies/variants into `ApiError::source` "for debug logging" across all the error helpers, but nothing in the request path ever logs it — `ApiError::into_response` (error.rs:102) drops `source` entirely, and there's no `tracing`/`Display` call on the error before it's converted to a response (no `TraceLayer` either). So the diagnostic payload is currently discarded and the stated debugging benefit doesn't materialize. Consider logging `err.source()` at `debug`/`warn` before returning (or adding a trace layer), otherwise the `source`-attaching machinery is dead weight.

crates/core/src/validatorapi/component.rs
```diff
@@ -4,30 +4,46 @@
...
+
+/// Hard deadline for upstream beacon-node calls. Bounds the worst-case
+/// handler latency when the upstream hangs or stalls. Roughly one slot.
+const UPSTREAM_REQUEST_TIMEOUT: Duration = Duration::from_secs(12);
+
+/// Hard deadline for the `attestation_data` await on the local DutyDB.
+/// Bounded so a request whose slot never produces consensus output cannot
+/// hold a handler task indefinitely. Sized at roughly two slots so a real
+/// attestation duty has time to flow through the pipeline.
+const ATTESTATION_DATA_TIMEOUT: Duration = Duration::from_secs(24);
```
varex83agent, line 47: nit: Charon's `AwaitAttestation` has no internal timeout — it blocks on `ctx.Done()`/shutdown and the deadline is entirely the VC's request context. The 24s `ATTESTATION_DATA_TIMEOUT` here fires independently of the caller. The client-observable status matches (Charon maps a cancelled ctx to 408 too), and 24s (~2 slots) is longer than any realistic attestation deadline, so this is a safe and justified deviation (it prevents handler-task leaks since the handler has no request-scoped cancellation token). Worth a one-line note in the doc comment that this is an intentional divergence from Charon rather than a ported behavior.

crates/core/src/validatorapi/types.rs
```diff
@@ -179,3 +190,100 @@ pub struct BeaconCommitteeSelection {}
...
+                    .map_err(|_| de::Error::custom("validator index must be non-negative"))
+            }
+
+            fn visit_str<E: de::Error>(self, v: &str) -> Result<Self::Value, E> {
+                v.parse::<u64>().map_err(de::Error::custom)?;
+                Ok(Element(v.to_owned()))
+            }
+        }
+
+        deserializer.deserialize_any(ElemVisitor)
```
varex83agent, line 287: nit: `Element::deserialize` ends in `deserializer.deserialize_any(ElemVisitor)`. Fine for JSON (self-describing) and the visitor correctly handles `u64`/`i64`/`str`, but `deserialize_any` is fragile for non-self-describing formats. Since this is only ever driven by axum's JSON extractor it's safe — a short `// JSON-only; deserialize_any is safe here` comment would document the constraint and prevent a future-reuse footgun.

crates/core/src/validatorapi/types.rs
```diff
@@ -4,15 +4,29 @@
...
+    GetSyncCommitteeDutiesResponseResponse as SyncCommitteeDutiesResponse,
+    GetSyncCommitteeDutiesResponseResponseDatum as SyncCommitteeDuty,
     GetVersionResponseResponse as NodeVersionResponse,
     GetVersionResponseResponseData as NodeVersionData,
-    spec::phase0::{Epoch, Root, Slot, ValidatorIndex},
+    spec::phase0::{self, Epoch, Root, Slot, ValidatorIndex},
 };
 
+/// Attestation data alias for the consensus-spec phase0 type.
+pub type AttestationData = phase0::AttestationData;
```
varex83agent, line 28: nit: `pub type AttestationData = phase0::AttestationData;` is consumed only by `AttestationDataResponse.data`, while tests and the dutydb construct `pluto_eth2api::spec::phase0::AttestationData` directly — two spellings for one type. Either drop the alias and use `phase0::AttestationData` inline, or apply the alias consistently. Cosmetic.

varex83agent (dismissed):
This PR wires up three validator-API handlers (`attester_duties`, `sync_committee_duties`, `attestation_data`) and is generally high quality — the hardening from the prior review round is solid (request timeouts, 64 KiB body cap, streaming `ValIndexes` visitor with an 8192 cap, no info leak into client-visible error messages, `new_insecure` gated to `#[cfg(test)]`), the `await_data` refactor is free of the classic lost-wakeup race (`notified()` is created before the lock-check), and the pubshare-swap parity with Charon (reject-unknown for attester/sync, pass-through for proposer) is correct. Test coverage is good.

Two must-fix issues before merge:

1. **Unbounded memory growth in the `evicted_*` HashSets** (`dutydb/memory.rs`). The four eviction-tracking sets introduced in the fix commit are only ever inserted into and never pruned, so every evicted slot/key is retained for the entire process lifetime — an unbounded leak in a long-running daemon and a divergence from Charon, which keeps no such set. This optimization needs a bound (e.g. a `min_live_slot` watermark, or pruning below a retention horizon).
2. **Missing `committee_index=0` dual-write breaks post-Electra VCs** (`dutydb/memory.rs`). Charon's `storeAttestationUnsafe` deliberately stores each attestation twice — under the real committee index *and* under index 0 — so post-Electra VCs that request `attestation_data?committee_index=0` resolve. Pluto's `store_attestation` writes only the real index; this PR makes the `attestation_data` endpoint live, so a conformant Electra VC asking for index 0 will now time out (408) and fail to attest. (Root cause is in pre-existing store logic, but it becomes reachable here.)

The remaining findings are minor/nit: an error-envelope inconsistency on the `{epoch}` path param, a `ValIndexes` parity superset (mixed numeric/string accepted where Charon is all-or-nothing), the `std::io::Error::other` debug-string carrier smell, and the fact that the `source` attached "for debug logging" is never actually logged. Requesting changes for the two majors; everything else is author discretion.

varex83agent:
## /loop-review-pr summary

Ran **3 review-and-fix iterations** (plus 2 verification passes) against this PR, each driving the same four-agent pipeline as `/review-pr` (functional-equivalence, security, rust-style, code-quality). No inline comments were posted during the loop — this is the single summary.

Terminated by: **completion (PR reached ideal state)** — internal review produced zero `bug`/`major` findings for two consecutive passes and all gates are green.

### Quality gates (final, full workspace, `--all-features`)
- `cargo +nightly fmt --all --check` — **pass**
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` — **pass**
- `cargo test --workspace --all-features` — **pass** (40/40 test groups, 0 failures)

### Resolved during the loop

**Bugs (0)**
- _none_

**Major (1)**
- Unbounded `MemDB` eviction state — the four `evicted_*` tombstone `HashSet`s were insert-only (one entry per evicted slot, forever), an unbounded memory leak on a long-running node. Replaced the three slot-keyed sets with O(1) `Option<u64>` high-water marks (attester/proposer/sync-contribution); dropped aggregation eviction tracking since it is awaited by root only and relies on the request timeout, matching Charon's Go dutydb (which keeps no eviction record at all). Added a regression test. — `crates/core/src/dutydb/memory.rs` — fix in `6a8a94c`

**Minor (1)**
- Malformed duties request body bypassed the `{code,message}` envelope and returned axum's default 422 for type errors. Now extracts `Result<Json<ValIndexes>, JsonRejection>` and maps via `json_rejection_to_api_error` to a uniform 400 (mirroring Charon's `unmarshal` and the router's own query-rejection path); the 413 body-size DoS defense is preserved explicitly. Added a router test. — `crates/core/src/validatorapi/router.rs` — fix in `b5eb0ca`

**Nits (1)**
- Regression-test arithmetic style (`SLOT - 1` → `SLOT.saturating_sub(1)`) for consistency — folded into `6a8a94c`

### Outstanding (minor/nit — do not block "ideal"; left for author judgement)
- **[minor]** A duties `POST` with a **missing** `Content-Type` header returns 400, whereas Charon's `wrap` treats an absent/empty Content-Type as `application/json` and parses the body (→ 200). Genuine but narrow Go-parity edge case; real validator clients always send the header. The fix is invasive (a content-type-injecting middleware, or swapping `Json` for `Bytes` + manual `serde_json`), so it was not applied — flagging for your call.
- **[minor]** `ValIndexes` accepts mixed numeric+string arrays (e.g. `[1, "2"]`) that Charon rejects. This appears **intentional** — the PR has a test (`val_indexes_accepts_mixed_elements`) asserting it — so it was left as-is.
- **[nit]** Accepted/intentional deviations: `VAL_INDEXES_MAX_LEN` (8192) + 64 KiB body cap (DoS hardening Charon lacks); `node_version` arch token uses Rust naming (`x86_64`) vs Go's `amd64` (cosmetic, no client impact); per-handler timeouts (12s/24s) vs Charon's uniform 10s (all resolve to 408).
- **[nit]** `wake()` takes `DutyType` by value (the clone is actually required — `DutyType` isn't `Copy` and `delete_duty` consumes the `Duty`); `parse_bls_pubkey(&str)` vs `impl AsRef<str>`; `json_rejection_to_api_error` could use an `@`-binding instead of re-constructing `PAYLOAD_TOO_LARGE`; `store()` holds the write lock across the async `deadliner.add()` (pre-existing, not worsened by this PR); the high-water-mark soundness invariant (deadliner expires in non-decreasing slot order) could carry a `debug_assert`.

### Verdict
**PR is ideal** — all `bug`/`major` findings resolved, the dutydb concurrency fix was independently verified sound (monotonic-slot invariant, no TOCTOU, no lost wakeup), and `fmt`/`clippy`/`test` are green across the full workspace. Remaining items are minor/nit and either intentional or low-value.

crates/core/src/validatorapi/component.rs
```diff
@@ -4,30 +4,46 @@
...
 };
-use crate::version;
+use crate::{
+    dutydb::{Error as DutyDbError, MemDB},
+    version,
+};
+
+/// Hard deadline for upstream beacon-node calls. Bounds the worst-case
+/// handler latency when the upstream hangs or stalls. Roughly one slot.
+const UPSTREAM_REQUEST_TIMEOUT: Duration = Duration::from_secs(12);
```
iamquang95: Why 12? Should be 10 seconds right?
`	defaultRequestTimeout                     = 10 * time.Second`

crates/core/src/validatorapi/router.rs
```diff
@@ -136,12 +171,76 @@ async fn proposer_duties(
...
+    ApiError::new(StatusCode::BAD_REQUEST, "invalid query parameters")
+        .with_source(std::io::Error::other(rejection.body_text()))
+}
+
+/// Renders an axum JSON body-extractor rejection as Pluto's standard
+/// [`ApiError`] body shape, so it shares the `{ "code", "message" }` schema
+/// instead of axum's default plain-text response.
+///
+/// Every genuine parse failure — malformed JSON (`400`), wrong element type
+/// (`422`), missing `content-type` (`415`) — is normalised to a uniform
```
iamquang95: missing content-type is accepted in charon

```
if contentHeader == "" || strings.Contains(contentHeader, string(contentTypeJSON)) {
			typ = contentTypeJSON
		} else if strings.Contains(contentHeader, string(contentTypeSSZ)) {
			typ = contentTypeSSZ
		} else {
			writeError(ctx, w, endpoint, apiError{
				StatusCode: http.StatusUnsupportedMediaType,
				Message:    "unsupported media type " + contentHeader,
			})

			return
		}
```

iamquang95 (approved):

crates/core/src/validatorapi/component.rs
varex83agent: Good catch — fixed in 12e2263 (timeout 12s → 10s, doc comment now points to `core/validatorapi/router.go:61`).

crates/core/src/validatorapi/router.rs
varex83agent: Fixed in 12e2263 — added an `enforce_json_content_type` middleware on the duties POST routes that mirrors Charon's policy (missing or JSON Content-Type passes; anything else → 415 with the offending value in the message). SSZ is not supported yet, so for now non-JSON is rejected; the middleware is the right seam to add SSZ later. Two new router tests cover the missing-CT and non-JSON cases.

PR #452, opened 2026-05-28
feat(core): implement `core/aggsigdb`

### LINKED ISSUE #169: Implement `core/aggsigdb`

emlautarom1:
Supersedes #430

---

Uses an Actor model as in Charon V1 which results in less locks and correct awaits on readers. Also, this implementation supports arbitrary cancellation through `CancellationToken`. This is not strictly needed, but adding it is trivial. Lastly, the latest changes on `deadliner` were merged which makes testing more straightforward.


emlautarom1:
@claude Review the code and take into consideration the original PR (430)

emlautarom1:
@claude Re-review the code and evaluate if the recent commits fix the issues detected previously.

crates/core/src/aggsigdb/memory.rs
```diff
@@ -0,0 +1,507 @@
...
+type Waiters =
+    HashMap<(types::Duty, types::PubKey), Vec<sync::oneshot::Sender<Box<dyn types::SignedData>>>>;
+
+struct Actor {
+    entries: HashMap<types::Duty, HashMap<types::PubKey, Box<dyn types::SignedData>>>,
+    waiters: Waiters,
+    deadliner: deadline::DeadlinerHandle,
+}
+
+impl Actor {
```
varex83: Can we use [actix](https://github.com/actix/actix)?

crates/core/src/aggsigdb/memory.rs
```diff
@@ -0,0 +1,507 @@
...
+    /// The request cannot be processed because the instance has been
+    /// terminated.
+    #[error("The instance has been terminated")]
+    Terminated,
+}
+
+type Waiters =
+    HashMap<(types::Duty, types::PubKey), Vec<sync::oneshot::Sender<Box<dyn types::SignedData>>>>;
+
+struct Actor {
```
mskrzypkows: I would name it differently to emphasize that it's db actor, e.g. `MemDBActor`.

crates/core/src/aggsigdb/memory.rs
```diff
@@ -0,0 +1,507 @@
...
+        response: sync::oneshot::Sender<Box<dyn types::SignedData>>,
+    },
+}
+
+/// An in-memory implementation of AggSigDB.
+///
+/// Share an instance by cloning. Cloning is cheap and creates a new reference
+/// to the same underlying data.
+#[derive(Clone)]
+pub struct Handle {
```
mskrzypkows: ditto

crates/core/src/aggsigdb/memory.rs
emlautarom1: I initially went that route and changed to this one for clarity but it seems like that's not the case.

Changed accordingly.

mskrzypkows (approved):

crates/core/src/aggsigdb/memory.rs
emlautarom1: Though I believe it's a pattern that accurately captures the way Charon models services so we can get a lot of benefit from it, we still don't have a lot of actors implemented in Pluto.

For now I would prefer to have them hand-rolled and if we see room for refactoring we can introduce some third-party library.

crates/core/src/aggsigdb/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+
+    /// The request cannot be processed because the instance has been
+    /// terminated.
+    #[error("The instance has been terminated")]
+    Terminated,
+}
+
+/// A persistent store for aggregated signed duty data.
+#[async_trait::async_trait]
+pub trait AggSigDB {
```
mskrzypkows: It's imported in memory.rs and below there is `pub mod memory;`, I'd advice to extract to other file, so there won't be cyclic dependencies.

emlautarom1: Could you elaborate on the suggested change? Right now there are no cyclic dependencies, and clients of this code look like this (example for the test suite):

https://github.com/NethermindEth/pluto/blob/f87b4b672df8fd15e8a90b403a9544d52e705c6c/crates/core/src/aggsigdb/memory.rs#L192-L201

mskrzypkows: I mean that `mod.rs` includes `memory.rs` and `memory.rs` has 
```
use crate::{
    aggsigdb::{AggSigDB, Error},
    deadline, types,
};
```
which is imported from `mod.rs`. 
So it goes like this mod.rs -> memory.rs -> mod.rs
I think it's better to use tree-like structure e.g.
mod.rs -> memory.rs -> aggsigdb.rs
mod.rs -> aggsigdb.rs

emlautarom1: I don't follow your suggestion, but we could make the structure look like:

```
aggsigdb/
  mod.rs
  traits.rs
  memory.rs
```

Where `mod.rs` declares both `traits.rs` and `memory.rs`.

Other modules like `parsigdb` and `dutydb` have everything in a single file/module.

mskrzypkows: Yes, my concern is about cyclic file dependency, so if we make as you described, mod.rs will include traits.rs and memory.rs, and we won't need to use mod.rs in memory.rs, we'll use trait.rs. In this case mod.rs is a parent and non of its children import code from it. I hope that this is clearer.

crates/core/src/aggsigdb/memory.rs
```diff
@@ -0,0 +1,491 @@
...
+            }
+
+            // After each message, trim waiters in case that the futures are dropped.
+            self.trim_readers();
+        }
+    }
+
+    async fn store(&mut self, duty: types::Duty, set: types::SignedDataSet) -> Result<(), Error> {
+        // TODO(charon): Distinguish between no deadline supported vs already expired.
+        let _ = self.deadliner.add(duty.clone()).await;
```
iamquang95, line 73: if set is empty, we should not add duty to deadliner

crates/core/src/aggsigdb/memory.rs
```diff
@@ -0,0 +1,491 @@
...
+        let for_duty = self.entries.entry(duty.clone()).or_default();
+        for (pub_key, signed_data) in set.into_iter() {
+            match for_duty.entry(pub_key) {
+                Entry::Vacant(slot) => {
+                    slot.insert(signed_data.clone());
+
+                    let k = (duty.clone(), pub_key);
+                    if let Some((_, waiters)) = self.waiters.remove_entry(&k) {
+                        for w in waiters {
+                            let _ = w.send(signed_data.clone());
```
iamquang95: `if !w.is_closed()` then send

iamquang95 (approved):

iamquang95 (approved):

crates/core/src/aggsigdb/memory.rs
emlautarom1: TOCTOU, just send and ignore the error (subscriber future dropped). Added only for optimization of the `.clone()`

crates/core/src/aggsigdb/memory.rs
emlautarom1: That matches the Charon behavior, fixed.

crates/core/src/aggsigdb/mod.rs
emlautarom1: Ok, that makes sense. Still, there is no issues with cyclic dependencies (Rust allows this), but I can make the change for clarity reasons.

PR #454, opened 2026-05-28
feat(core): implement `core/fetcher`

emlautarom1:
The fetcher is stage 2 of the duty pipeline (`Scheduler → Fetcher → Consensus → …`): it takes a `Duty` plus a per-validator `DutyDefinitionSet` and fetches the corresponding unsigned duty data from the beacon node, producing an `UnsignedDataSet` that it hands to its subscribers.

## Highlights

- **New `crates/core/src/fetcher/` module** with the full `Fetcher` and a `GraffitiBuilder`, covering all four duty paths: attester, aggregator, block proposer, and sync-committee contribution.
- **`DutyDefinition` reworked into a heterogeneous enum** (`Attester` / `Proposer` / `SyncCommittee`), mirroring Charon's `core.DutyDefinition` interface — replacing the previous generic wrapper.
- **`UnsignedDutyData` / `UnsignedDataSet` promoted into `core::types`** so the fetcher, DutyDB, and consensus share one canonical, `PubKey`-keyed type (matching Go's `map[core.PubKey]core.UnsignedData`).
- **Every Go test ported and passing**.
- **No new dependencies introduced.**

## Type-system changes (`crates/core/src/types.rs`)

The fetcher's input and output types did not previously exist in a usable form; both were reworked first.

### `DutyDefinition` → enum

Charon models duty definitions as a `DutyDefinition` interface with three implementations (`AttesterDefinition`, `ProposerDefinition`, `SyncCommitteeDefinition`). Pluto previously had a generic `DutyDefinition<T>` / `DutyDefinitionSet<T>` which cannot represent a heterogeneous, type-dispatched set the way the fetcher needs.

This was replaced with:

- `enum DutyDefinition { Attester(..), Proposer(..), SyncCommittee(..) }` with `as_attester` / `as_proposer` / `as_sync_committee` accessors (the Rust equivalent of Go's type assertions).
- `pub type DutyDefinitionSet = HashMap<PubKey, DutyDefinition>` (matches Go's `map[core.PubKey]core.DutyDefinition`).
- New `ProposerDuty` and `SyncCommitteeDuty` structs (faithful to eth2 `v1.ProposerDuty` / `v1.SyncCommitteeDuty`) and the three `*Definition` newtype wrappers. `AttesterDefinition` wraps the pre-existing `signeddata::AttesterDuty`.

### `UnsignedDataSet` promoted to `core::types`

`UnsignedDutyData` (enum over `Proposal` / `Attestation` / `AggAttestation` / `SyncContribution`) and `UnsignedDataSet = HashMap<PubKey, UnsignedDutyData>` were moved out of `dutydb/memory.rs` into `core::types`. The old generic type was removed.

### `graffiti.rs`

Port of `charon/core/fetcher/graffiti.go`: `GraffitiBuilder`, `client_graffiti_mappings`, and the build/default/token helpers. The default graffiti uses a `pluto/<version>-<commit>` prefix (vs Go's `charon/...`).

### `mod.rs`

Port of `charon/core/fetcher/fetcher.go`:

- `Fetcher::new`, `subscribe`, `register_agg_sig_db`, `register_await_att_data`, and `fetch` (dispatch on `DutyType`).
- The four data paths: `fetch_attester_data`, `fetch_aggregator_data`, `fetch_proposer_data`, `fetch_contribution_data`, plus `verify_fee_recipient` and the pubkeys logging tracker.
- Beacon-node calls map onto the generated `EthBeaconNodeApiClient`: `produce_attestation_data`, `get_aggregated_attestation_v2`, `produce_block_v3`, `produce_sync_committee_contribution`. Loosely-typed responses are decoded into the strongly-typed `signeddata` types via JSON round-trip; the `produce_block_v3` response is mapped into an unsigned `VersionedProposal` per fork/blinded variant.
- Aggregator selection reuses the existing `pluto_eth2util::eth2exp` (`is_att_aggregator`, `is_sync_comm_aggregator`).
- `AggSigDB` and `DutyDB` are injected as async callbacks (no hard dependency on those components); `Box<dyn SignedData>` values from `AggSigDB` are down-cast to concrete types via trait upcasting.
- Go error strings preserved, e.g. `aggregate attestation not found by root (retryable)`, and the `errors.Wrap(err, "fetch <type> data")` wrapping is reproduced via a `FetcherError::Fetch { context, source }` variant.

## Tests

All Go tests are ported and passing (test names kept where practical):

- **Graffiti** (5, internal): `fetch_beacon_node_token`, `build_graffiti`, `default_graffiti`, `get_graffiti`, `new_graffiti_builder`.
- **Attester**: `fetch_attester`.
- **Aggregator** (4): `different_committee`, `same_committee`, `no_aggregator`, `nil_aggregate`.
- **Sync contribution** (3): `aggregator`, `not_aggregator`, `data_error`.
- **Blocks**: `fetch_blocks`.
- **Fee recipient** (internal): `verify_fee_recipient`.

Per the chosen approach, beacon-node responses that depend on request parameters (aggregate attestation by data root, sync contribution by subcommittee/root, block proposal echoing randao + graffiti) are provided by **test-local `wiremock` responders** rather than by extending `BeaconMock`. The block-proposal test reuses the existing Electra `VersionedProposal` golden fixture as the `produce_block_v3` payload. Aggregator/sync tests use a custom spec carrying the selection fields required by `eth2exp`.

`verify_fee_recipient`'s comparison logic was factored into a pure `fee_recipient_mismatch` helper so it can be asserted directly without a tracing-log capture harness; the warning itself still lives in `verify_fee_recipient`.

## Files changed

- `crates/core/src/types.rs` — `DutyDefinition` enum, new duty/definition types, promoted `UnsignedDutyData`/`UnsignedDataSet`.
- `crates/core/src/dutydb/memory.rs`, `crates/core/src/dutydb/mod.rs` — import and re-export the promoted types instead of defining them.
- `crates/core/src/lib.rs` — register the `fetcher` module.
- `crates/core/src/fetcher/mod.rs`, `crates/core/src/fetcher/graffiti.rs` — new.

## Caveats and notes for review

- **Pre-existing deviation:** Pluto's `AttesterDuty` omits the `PubKey` field present in eth2's `v1.AttesterDuty`. The fetcher does not depend on it (the pubkey is always the map key), but it is worth keeping in mind when the scheduler begins populating these definitions.
- The fetcher's `DutyDefinitionSet` is produced by the scheduler, which is in-flight on `emlautarom1/core-scheduler`. The definition types added here may need reconciliation if that branch introduces overlapping types.

---

_This PR was automatically generated by [Claude](https://claude.com/claude-code)._


crates/core/src/types.rs
```diff
@@ -426,134 +431,137 @@ impl AsRef<[u8]> for PubKey {
...
-    }
-
-    /// Get an unsigned data by duty type.
-    pub fn get(&self, duty_type: &DutyType) -> Option<&UnsignedData<T>> {
-        self.0.get(duty_type)
-    }
+/// Per-validator duty definition. The Rust equivalent of Go's
+/// `core.DutyDefinition` interface: a closed set of concrete duty definitions.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum DutyDefinition {
```
emlautarom1, line 575: Change from the generic `DutyDefinition<T>` to an enum that mirrors Charon's `core.DutyDefinition` interface (`AttesterDefinition` / `ProposerDefinition` / `SyncCommitteeDefinition`). The set of types is known ahead of time so no need to replicate the `interface` with `trait` tricks.

This change is also required by `core/scheduler` so it's known to be correct.

crates/core/src/fetcher/mod.rs
```diff
@@ -0,0 +1,1627 @@
...
+};
+
+/// Boxed error returned by injected callbacks (subscribers, AggSigDB, DutyDB).
+type BoxError = Box<dyn std::error::Error + Send + Sync>;
+
+/// Future returned by an injected callback.
+type CallbackFuture<T> = Pin<Box<dyn Future<Output = std::result::Result<T, BoxError>> + Send>>;
+
+/// Subscriber callback invoked for each fetched duty data set.
+pub type Subscriber = Arc<dyn Fn(Duty, UnsignedDataSet) -> CallbackFuture<()> + Send + Sync>;
```
emlautarom1, line 40: Async-callback shape note: `Arc<dyn Fn(...) -> Pin<Box<dyn Future + Send>>>` is the same idiom `sigagg` already uses, so we're consistent across components.

`BoxError = Box<dyn std::error::Error + Send + Sync>` is intentionally loose so   callers (AggSigDB / DutyDB / subscribers) can return their own error types   without dragging concrete types into the fetcher's public surface. The cost is   that errors get stringified at the boundary.

crates/core/src/fetcher/mod.rs
```diff
@@ -0,0 +1,1627 @@
...
+    T: serde::de::DeserializeOwned,
+    S: serde::Serialize,
+{
+    let value = serde_json::to_value(value)?;
+    Ok(serde_json::from_value(value)?)
+}
+
+/// Converts a `produce_block_v3` response into an unsigned
+/// [`VersionedProposal`].
+fn versioned_proposal_from_response(
```
emlautarom1: This is the trickiest piece in the file. The generated `produce_block_v3` response carries a loosely-typed `data`; mapping it onto the strongly-typed `ProposalBlock` enum has to fan out per fork and per blinded/unblinded shape, because Deneb+ wrap the block in `{block, kzg_proofs, blobs}` while pre-Deneb is the block directly. The dispatch here matches `validatormock::propose.rs`'s `build_block_body` (Deneb/Electra/Fulu use `block_field(&data)`).

Risk surface: we're trusting the JSON shape the beacon node returns to match the fork keys exactly. The `fetch_blocks` test exercises the Electra-unblinded path (reusing the existing `VersionedProposal` golden); the other forks rely on the structural match. If a real beacon node ever returns an unexpected shape, the user sees `decode beacon node response: ...` from `Json(#[from] ...)` which is actionable.

Future improvement: once a versioned-decode helper lands in `eth2api`, we should swap this match out for it.

crates/core/src/fetcher/mod.rs
```diff
@@ -0,0 +1,1627 @@
...
+            actual = %actual,
+            "Proposal with unexpected fee recipient address"
+        );
+    }
+}
+
+/// Returns `Some((expected, actual))` when the proposal's fee recipient differs
+/// (case-insensitively) from `fee_recipient_address`. Returns `None` for forks
+/// without a fee recipient (pre-Bellatrix) or when the addresses match.
+fn fee_recipient_mismatch(
```
emlautarom1, line 600: The Go test for this is in `fetcher_internal_test.go` and asserts that a warning is emitted on mismatch / silent on match across every fork. Pluto has no log-capture utility yet, so the comparison logic was factored into `fee_recipient_mismatch` (pure, testable) and the `warn\!` stays in `verify_fee_recipient`. The ported test asserts on the pure helper.

Two notes:
* The extraction uses `execution_payload` / `execution_payload_header` JSON keys   uniformly across forks instead of typed field navigation. Cleaner than 5x   per-fork accessors and keeps the same behaviour, but it's silent if a future   fork renames either key. Worth a once-over if a new fork lands.
* `proposal.version()` is queried first to early-return on Phase0/Altair which   carry no fee recipient — matches Go's `default: return` in the switch.

emlautarom1 (commented):
Walking through the parts of this change that warrant the most discussion. Overall the structure mirrors Charon faithfully and the test coverage looks solid. Leaving notes on a few design points rather than blocking changes.

emlautarom1:
@claude Review this code

emlautarom1:
@claude Review this PR and take into consideration your previous comments.

emlautarom1:
The following issue is still present:

> Non-`Ok` BN response still collapses 404 and 5xx into the same retryable error** — `mod.rs:466`, `:496`, `:530`

We'll keep it simple for now and revisit this once we start doing the wiring. I don't think we should be spending much time on getting retrying working immediately without sufficient context.

emlautarom1:
@claude Review this PR and take into consideration your previous comments and my own.

varex83:
@varex83agent /review-pr

crates/core/src/fetcher/graffiti.rs
```diff
@@ -0,0 +1,347 @@
...
+pub enum GraffitiError {
+    /// More than one graffiti value was provided but the count did not match
+    /// the number of validators.
+    #[error("graffiti length must match the number of validators or be a single value")]
+    LengthMismatch,
+}
+
+/// Maps beacon node product tokens (the first `/`-separated component of the
+/// node version string) to their two-letter graffiti code.
+pub fn client_graffiti_mappings() -> HashMap<&'static str, &'static str> {
```
iamquang95: nit: it allocates a new hashmap on every call, better use a const?

crates/core/src/fetcher/graffiti.rs
```diff
@@ -0,0 +1,347 @@
...
+            graffiti: HashMap::from([(pubkeys[0], g0), (pubkeys[1], g1)]),
+        };
+
+        assert_eq!(builder.get_graffiti(&pubkeys[0]), g0);
+        assert_eq!(builder.get_graffiti(&pubkeys[1]), g1);
+        assert_eq!(builder.get_graffiti(&pubkeys[2]), super::default_graffiti());
+    }
+
+    #[tokio::test]
+    async fn new_graffiti_builder() {
```
iamquang95: nit: better split this test into multiple smaller ones

crates/core/src/fetcher/mod.rs
```diff
@@ -0,0 +1,1716 @@
...
+            };
+
+            let proposal = VersionedProposal::try_from(&response)?;
+
+            // Builders set the fee recipient to themselves, so it always differs
+            // from the validator's; only verify when the builder is disabled.
+            if !self.builder_enabled {
+                let fee_recipient = (self.fee_recipient)(pubkey);
+                verify_fee_recipient(&proposal, &fee_recipient);
+            }
```
iamquang95, line 378: https://github.com/ObolNetwork/charon/blob/v1.7.1/core/fetcher/fetcher.go#L294-L304

Charon check fee_recipient on raw proposal

iamquang95 (approved):
LGTM

crates/eth2api/src/spec/phase0.rs
```diff
@@ -372,6 +404,23 @@ pub struct Attestation {
...
 }
 
+impl TryFrom<&crate::GetBlockAttestationsV2ResponseResponseDataArray2> for Attestation {
+    type Error = ConversionError;
+
+    fn try_from(
+        value: &crate::GetBlockAttestationsV2ResponseResponseDataArray2,
+    ) -> Result<Self, Self::Error> {
+        Ok(Self {
+            aggregation_bits: BitList::from_ssz_bytes(decode_hex_var(
```
varex83agent, line 414: nit: `aggregation_bits` is decoded with the inherent `pluto_ssz::BitList::from_ssz_bytes(Vec<u8>) -> Self`, which is infallible by design: on empty input, or input whose final byte is `0` (missing sentinel bit), it returns `Self::default()` (an empty bitlist) rather than erroring. Every other field here errors on malformed input, and the sibling `committee_bits`/`BitVector` decode is error-checked via `<BitVector as ssz::Decode>::from_ssz_bytes(...).map_err(...)`, so the asymmetry stands out.

That said, this faithfully reproduces the serde `Deserialize` path it replaces (`BitList::Deserialize` also calls `from_ssz_bytes` infallibly — `crates/ssz/src/types.rs:428`), and the `attestation_try_from_matches_json_roundtrip` test confirms equivalence — so it is not a regression. Even routing through `ssz::Decode` would not help, since that impl just wraps the infallible version in `Ok`. Flagging for consistency/robustness only. The same pattern appears at `crates/eth2api/src/spec/electra.rs:86`.

crates/core/src/fetcher/mod.rs
```diff
@@ -0,0 +1,1716 @@
...
+        let ok = match self
+            .eth2_cl
+            .get_aggregated_attestation_v2(request)
+            .await
+            .map_err(EthBeaconNodeApiClientError::RequestError)?
+        {
+            GetAggregatedAttestationV2Response::Ok(ok) => ok,
+            // Some beacon nodes return nil if the root is not found; surface a
+            // retryable error.
+            _ => return Err(FetcherError::AggregateAttestationNotFound),
```
varex83agent, line 489: The `_ =>` arm maps *any* non-`Ok` response variant to `AggregateAttestationNotFound` ("...retryable"). The generated client emits a distinct variant per declared status (400/404/406/500/...), and transport failures are already handled separately via `map_err(RequestError)`, so a 400/406/500 here is reported as retryable-not-found. The same shape applies to `sync_committee_contribution` (line 529 -> `SyncContributionNotFound`) and `attestation_data` (line 468 -> `NilAttestationData`). In Charon the synthetic "not found (retryable)" string is only produced when the call *succeeds* but `Data == nil`; non-2xx surfaces the raw beacon-node error. Treating 404 as not-found is fine (the `fetch_aggregator_nil_aggregate` test relies on it), but mislabeling genuine 4xx/5xx as retryable can change retry behavior and operator diagnostics. Consider matching the 404/not-found variant specifically.

varex83agent (commented):
## Review: `feat(core): implement core/fetcher`

This is a faithful, well-structured port of Charon's `core/fetcher` (attester, aggregator, proposer, sync-contribution paths) plus the `GraffitiBuilder`, with the supporting `DutyDefinition` enum rework and a clean set of `TryFrom` conversions replacing JSON round-trips in `eth2api`. The Go test suite is ported comprehensively and the per-fork `verify_fee_recipient` table-driven test strengthens the original. Parity was verified across all four duty paths, the Electra committee-index-0 collapsing, fee-recipient verification, `attestation_payload` version dispatch, and `PubkeysTracker` Drop-vs-defer logging — all behave equivalently (the Rust `attestation_payload`/`DataVersion::from` totality is arguably safer than Go's).

**Bugs (must-fix): none.** No reachable panics in non-test code, no concurrency or state-machine defects, and the parsing paths propagate errors rather than crashing on malformed beacon-node input.

**Minor / nit findings** (author discretion):
1. `phase0`/`electra` `Attestation::try_from` decode `aggregation_bits` via the infallible `BitList::from_ssz_bytes`, which silently yields an empty bitlist on empty/zero-sentinel input — asymmetric with the error-checked `BitVector` sibling. This matches the serde round-trip it replaces (confirmed by the equivalence tests), so it is a robustness/consistency nit, not a regression.
2. Non-200 beacon responses for attestation-data / aggregate-attestation / sync-contribution collapse into the fixed `nil`/`not found (retryable)` errors, mislabeling 4xx/5xx as retryable-not-found.
3. Dead `FetcherError::UnsupportedProposalVersion` variant; `FetcherError::Signature(String)` is stringly-typed and drops the `SignedDataError` source chain.
4. `impl TryInto<AttesterDutyDefinition>` should be `impl TryFrom` for consistency with the sibling `ProposerDutyDefinition` and idiomatic guidance.
5. The documented `AttesterDuty`-omits-`PubKey` deviation is acceptable within the fetcher (pubkey is the map key), but worth tracking for downstream `AttestationData` serialization parity. Deneb+ `blobs`/`kzg_proofs` deserialize into unbounded `Vec`s (pre-existing types) — informational.

Verdict: **COMMENT** — only minor/nit items, none blocking. Left open for author discretion.

crates/core/src/fetcher/mod.rs
```diff
@@ -0,0 +1,1716 @@
...
+    #[error("convert beacon node response: {0}")]
+    Conversion(#[from] ConversionError),
+
+    /// Failed to decode a beacon node response into a signed-data type.
+    #[error("decode proposal: {0}")]
+    SignedData(#[from] SignedDataError),
+
+    /// A versioned proposal had an unsupported fork version.
+    #[error("unsupported proposal version: {0:?}")]
+    UnsupportedProposalVersion(ConsensusVersion),
```
varex83agent: nit: `FetcherError::UnsupportedProposalVersion(ConsensusVersion)` is declared but never constructed — `VersionedProposal::try_from` exhaustively matches all `ConsensusVersion` variants, so the unsupported-version path no longer exists. Public enum variants are exempt from `dead_code`, so this won't be flagged automatically. Remove it (or wire it in if a fallible path is intended).

crates/core/src/fetcher/mod.rs
```diff
@@ -0,0 +1,1716 @@
...
+
+    /// A signed data value could not produce a signature.
+    #[error("signature: {0}")]
+    Signature(String),
```
varex83agent: nit: `Signature(String)` is populated at line 352 via `.map_err(|e| FetcherError::Signature(e.to_string()))` where `e: SignedDataError`. Stringifying discards `Error::source()` chaining. Since the crate already has `SignedData(#[from] SignedDataError)`, prefer a typed `Signature(#[source] SignedDataError)` with `#[error("signature: {0}")]` to keep the distinct context while preserving the source.

crates/core/src/types.rs
```diff
@@ -445,18 +447,39 @@ impl TryInto<AttesterDutyDefinition>
     fn try_into(self) -> Result<AttesterDutyDefinition, Self::Error> {
```
varex83agent: nit: this impl block (reworked in this PR) implements `TryInto<AttesterDutyDefinition>` directly. Idiomatic guidance is to implement `TryFrom` and let the blanket impl provide `TryInto` for free, which also lets callers write `AttesterDutyDefinition::try_from(x)`. The sibling `ProposerDutyDefinition` in this same file already uses `TryFrom`; switching this one over would make them consistent.

crates/core/src/fetcher/mod.rs
```diff
@@ -0,0 +1,1716 @@
...
+    fee_recipient_address: &str,
+) -> Option<(String, String)> {
+    if matches!(
+        proposal.version(),
+        versioned::DataVersion::Phase0 | versioned::DataVersion::Altair
+    ) {
+        return None;
+    }
+
+    let value = serde_json::to_value(proposal_body(&proposal.block)).ok()?;
```
varex83agent: nit: `serde_json::to_value(...).ok()?` (and `proposal_body`'s `unwrap_or(Value::Null)`) silently swallow serialization errors. This is defensible — the whole path is best-effort, warn-only verification mirroring Go's `verifyFeeRecipient`, and the missing-address case is already guarded by `debug_assert!`/`warn!`. Consider a one-line comment noting the deliberate swallow, or routing a `to_value` failure into the same warn branch so a serialization regression isn't entirely invisible.

crates/core/src/fetcher/graffiti.rs
varex83: const / lazy would be good, agree on that

crates/core/src/fetcher/mod.rs
```diff
@@ -0,0 +1,1716 @@
...
+    }
+}
+
+/// Downcasts a `&dyn SignedData` to a concrete signed-data type.
+fn downcast<T: 'static>(data: &dyn SignedData) -> Option<&T> {
+    (data as &dyn std::any::Any).downcast_ref::<T>()
+}
+
+/// Formats bytes as a `0x`-prefixed lowercase hex string.
+fn hex_0x(bytes: &[u8]) -> String {
```
varex83: Do we need this function? I think we had this somewhere, maybe inline?

varex83 (approved):
LGTM!

crates/core/src/fetcher/mod.rs
emlautarom1: We have it defined a gazillion times. Agents love to add it. I'll inline it

crates/core/src/fetcher/graffiti.rs
emlautarom1: Changed to a `match` statement

crates/core/src/fetcher/graffiti.rs
emlautarom1: Splitted

crates/core/src/fetcher/mod.rs
emlautarom1: Removed case.

crates/core/src/fetcher/mod.rs
emlautarom1: Replaced with an inner `SignedDataError`

crates/core/src/fetcher/mod.rs
emlautarom1: Added a note regarding this potential issue.

crates/core/src/types.rs
emlautarom1: Changed to `TryFrom` following the codebase guidelines.

iamquang95 (approved):

crates/core/src/fetcher/mod.rs
emlautarom1: Could you elaborate a bit?

PR #455, opened 2026-05-29
feat(core): additional tracker modules

mskrzypkows:


mskrzypkows:
@claude review

crates/core/src/tracker/analysis.rs
```diff
@@ -0,0 +1,1456 @@
...
+/// Partial signatures grouped by message root, grouped by pubkey.
+///
+/// Equivalent to Go's `parsigsByMsg`.
+pub type ParSigsByMsg = HashMap<PubKey, HashMap<[u8; 32], Vec<crate::types::ParSignedData>>>;
+
+/// Returns true if every pubkey has at most one distinct message root.
+pub(crate) fn msg_roots_consistent(parsigs: &ParSigsByMsg) -> bool {
+    parsigs.values().all(|roots| roots.len() <= 1)
+}
+
```
mskrzypkows, line 43: `incl_supported()` acquires a `RwLock` read guard and allocates a `HashSet` on every invocation. `last_step()` calls it once; `duty_failed_step()` calls `last_step()` once. For proposer/aggregator/sync-contribution duties, `analyse_duty_failed` also recurses into prerequisite-duty analysers, each calling `duty_failed_step()` again — so 2–4 lock acquisitions per duty analysis.

Consider computing it once in `TrackerService::analyse()` and passing it as a parameter, or snapshot the feature flag at service construction time if it is not expected to change at runtime.

crates/core/src/tracker/analysis.rs
```diff
@@ -0,0 +1,1456 @@
...
+        Step::ChainInclusion => {
+            if err.is_none() {
+                err = Some(string_error("bug: missing chain inclusion error"));
+            } else {
+                reason = REASON_NOT_INCLUDED_ON_CHAIN;
+            }
+        }
+        Step::Zero => {
+            err = Some(string_error("no events for duty"));
+        }
```
mskrzypkows, line 250: Parity gap with Go's `analyseFetcherFailed` (tracker.go:299–305). Go has a three-tier reason assignment:
1. `eth2api.Error` → `reasonFetchBNError`
2. `context.Canceled` / `context.DeadlineExceeded` → stays default (treated as expected timeout, not a bug)
3. anything else → `reasonBugFetchError`

The Rust port only branches on eth2 vs non-eth2, so context-cancellation–equivalent errors are labelled `REASON_BUG_FETCH_ERROR` rather than being treated as expected. The emitted metric label is identical today, but this will diverge if distinct reason codes are ever added for the two sub-categories.

crates/core/src/tracker/analysis.rs
```diff
@@ -0,0 +1,1456 @@
...
+            REASON_SYNC_CONTRIBUTION_ZERO_PREPARES, REASON_UNKNOWN,
+            REASON_ZERO_AGGREGATOR_SELECTIONS, Reason,
+        },
+        step::Step,
+    },
+    types::{Duty, DutyType, PubKey},
+};
+
+/// Partial signatures grouped by message root, grouped by pubkey.
+///
```
mskrzypkows: nit: `ParSignedData` is used fully-qualified (`crate::types::ParSignedData`) inside the alias, but every other `crate::types` name in this file is imported via the `use` block. Add `ParSignedData` to the existing import and shorten to:
```rust
pub type ParSigsByMsg = HashMap<PubKey, HashMap<[u8; 32], Vec<ParSignedData>>>;
```

crates/core/src/tracker/analysis.rs
```diff
@@ -0,0 +1,1456 @@
...
+            Err(err) => {
+                tracing::warn!(error = %err, "Parsig message root");
+                continue;
+            }
+        };
+
+        resp.entry(e.pubkey)
+            .or_default()
+            .entry(root)
+            .or_default()
```
mskrzypkows, line 446: nit: `saturating_add(1)` on a `usize` counter of partial-signature events — this cannot realistically overflow. Plain `*slot += 1` is clearer and avoids implying overflow is a real concern.

crates/core/src/tracker/reporters.rs
```diff
@@ -0,0 +1,441 @@
...
+        metrics::TRACKER_METRICS,
+        reason::{
+            REASON_SYNC_CONTRIBUTION_ZERO_PREPARES, REASON_ZERO_AGGREGATOR_SELECTIONS, Reason,
+        },
+        step::Step,
+    },
+    types::{Duty, DutyType},
+};
+
+pub(crate) trait DutyFailureReporter: Send {
```
mskrzypkows: `DutyFailureReporter` is called for both failures **and** successes — `MetricsFailedDutyReporter` increments `success_duties_total` on the `!failed` path. The name implies it only fires on failure. Consider `DutyOutcomeReporter` or `DutyResultReporter` to match actual semantics.

crates/core/src/tracker/reporters.rs
```diff
@@ -0,0 +1,441 @@
...
+                );
+                TRACKER_METRICS.unexpected_events_total[&peer.name].inc_by(unexp as u64);
+            } else {
+                absent.push(peer.name.clone());
+                TRACKER_METRICS.participation[&labels].set(0);
+            }
+        }
+
+        // Only log when the absent set changes from the previous duty of this
+        // type, to avoid log spam every slot.
```
mskrzypkows, line 248: Unnecessary clone of the `prev_absent` Vec — it is only used for equality comparison, so cloning it allocates a new `Vec<String>` every analysis cycle. Compare via reference instead:
```rust
let same = self.prev_absent.get(&duty.duty_type).map_or(false, |p| p == &absent);
if !same { /* log */ }
```

crates/core/src/tracker/reporters.rs
```diff
@@ -0,0 +1,441 @@
...
+            failed,
+            participated,
+            unexpected,
+            expected_per_peer,
+        );
+    }
+}
+
+/// Reports inconsistent partial signature data across peers.
+pub fn report_par_sigs(duty: &Duty, parsigs: &ParSigsByMsg) {
```
mskrzypkows, line 284: The Rust port has **silently fixed a latent Go bug** here. Go's `reportParSigs` (tracker.go:851) writes `if len(parsigMsgs) <= 1 { continue }` where `parsigMsgs` is the *outer* map keyed by pubkey — so it checks the number of pubkeys in the whole duty, not the number of distinct message roots for the *current* pubkey. When only one pubkey has inconsistent roots, Go skips logging for it. The Rust `if by_root.len() <= 1 { continue; }` against the per-pubkey inner map is semantically correct. This intentional fix should have a comment so reviewers do not accidentally revert it to match Go.

mskrzypkows (commented):
This PR implements the tracker analysis pipeline — `analyse_duty_failed`, `analyse_participation`, `extract_par_sigs`, reporters, and metrics — a substantial and well-structured port of Go's Charon tracker. Test coverage for the analysis functions is thorough.

**Issues to address before merge:**
1. `incl_supported()` acquires a global `RwLock` and allocates a `HashSet` on every call — called 2–4× per duty analysis. Should be computed once and passed down.
2. The `events` HashMap in `run()` has no per-duty event cap — Vecs can grow without bound if a peer floods par-sig events before the deleter fires.
3. `analyse_fetcher_failed` is missing Go's `context.Canceled` special-case — parity gap in error classification.

Several minor style and naming issues noted inline.

mskrzypkows:
@claude review my changes, concentrate on major issues

crates/core/src/tracker/analysis.rs
```diff
@@ -0,0 +1,1480 @@
...
+}
+
+/// Set of duty types for which chain inclusion is supported.
+///
+/// The result is cached for the lifetime of the process. This assumes
+/// `GLOBAL_STATE` (and therefore `Feature::AttestationInclusion`) is
+/// configured once at startup and never mutated afterward — matching Go,
+/// which reads the flag on every call but relies on the same invariant.
+pub(crate) fn incl_supported() -> &'static HashSet<DutyType> {
+    static CACHE: OnceLock<HashSet<DutyType>> = OnceLock::new();
```
iamquang95, line 51: Not sure should we cache this or keep this same as charon, decision is up to you.

iamquang95 (approved):

crates/core/src/tracker/analysis.rs
mskrzypkows: If it's not modified later I would keep the cache, not to recreate a hashmap few times.

PR #456, opened 2026-05-29
refactor: split p2p relay

therustmonk:
Split `crates/p2p/src/relay.rs` into a `relay/` directory module with focused submodules:

  - `event` — public event/error types: `RelayManagerEvent`, `RelayDialError`, `RelayDialType`
  - `dial` — dial-campaign machinery: exponential backoff and the per-target `RelayDialState` stream
  - `manager` — the `RelayManager` behaviour, its `RelayConnectionState` lifecycle, and the `NetworkBehaviour` impl


varex83 (approved):

crates/p2p/src/relay/manager.rs
```diff
@@ -0,0 +1,702 @@
...
+            self.redial_relay(peer_id);
+            if was_reserved {
+                self.refresh_peer_dials();
+            }
+        } else if self.p2p_context.is_known_peer(&peer_id) {
+            self.reroute_peer(peer_id);
+        }
+    }
+
+    /// Reacts to a dial failure by logging and emitting a `DialFailed` event.
```
emlautarom1:
```suggestion
    /// Reacts to a dial failure by logging and emitting a [`RelayManagerEvent::DialFailed`] event.
```

emlautarom1: There are a couple like this that are no longer reference links.

emlautarom1 (approved):

crates/p2p/src/relay/dial.rs
```diff
@@ -0,0 +1,169 @@
...
+
+/// Returns true if both slices contain the same multiaddrs (order-independent).
+/// Used to decide whether a routing refresh actually expanded the available
+/// circuit paths to a peer — if it did, the dial state's backoff is reset.
+pub(super) fn addr_sets_equal(a: &[Multiaddr], b: &[Multiaddr]) -> bool {
+    if a.len() != b.len() {
+        return false;
+    }
+    let a_set: HashSet<&Multiaddr> = a.iter().collect();
+    b.iter().all(|x| a_set.contains(x))
```
iamquang95: This logic is not correct right? 

If b is `[A, A]`, and a is `[A, B]` this function return `true`

emlautarom1: That's correct, prefer to compare as sets instead (I would not even bother with the length check, it could be misleading in the case of duplicated entries):

```
let a_set: HashSet<&Multiaddr> = a.iter().collect();
let b_set: HashSet<&Multiaddr> = b.iter().collect();

a == b
```

emlautarom1: Note that this is a refactor, so the issue is preexisting: https://github.com/NethermindEth/pluto/blob/0f3cf70136e7282f00c6d882bdda436cbd6b705f/crates/p2p/src/relay.rs#L279-L285

PR #457, opened 2026-05-29
feat(core): implement validatorapi validators handler

varex83agent:
## Summary

- Implements `Component::validators` in `crates/core/src/validatorapi/component.rs`, replacing the previous `unimplemented!()` stub. Port of `Validators` in `core/validatorapi/validatorapi.go` (Charon v1.7.1, lines 1218–1296).
- Adds a `convert_validators` helper that mirrors `convertValidators` (`validatorapi.go` lines 1305–1332), rewriting each upstream validator's `inner.pubkey` from the cluster's root key to this node's public share. The `ignore_not_found` flag follows the Go contract — pass-through when the request filtered by indices only, error otherwise.
- Adds `invert_pub_share_map`, used once per request to translate the VC-supplied pubshares back into root pubkeys before forwarding to the upstream `POST /eth/v1/beacon/states/{state_id}/validators` endpoint.
- Aliases `Validator` to the auto-generated `GetStateValidatorsResponseResponseDatum` in `validatorapi/types.rs`, matching the established pattern used by the other duty types.
- Bounds the upstream call with `UPSTREAM_REQUEST_TIMEOUT` (504 on elapse), maps transport failures to 502, and propagates upstream 400 / 404 without leaking the upstream body into the client-visible message. Reuses the existing `upstream_timeout` / `upstream_call_failed` / `upstream_status_error` / `upstream_unexpected` helpers.

## Go reference

| Pluto (Rust) | Charon (Go) |
| --- | --- |
| `Component::validators` (`component.rs`) | `Component.Validators` (`validatorapi.go:1218–1296`) |
| `convert_validators` (`component.rs`) | `Component.convertValidators` (`validatorapi.go:1305–1332`) |
| `invert_pub_share_map` (`component.rs`) | `keysByShare` map built in `NewComponent` (Go derives once at construction; we build per request since it is small and only the `validators` handler needs it) |
| Input pubshare → root pubkey rejection (400 `unknown validator public key in request`) | `getPubKeyFunc` "unknown public key" branch in `validatorapi.go` |

Notable simplification vs Go: Pluto's `EthBeaconNodeApiClient` is the plain auto-generated client and has no `CompleteValidators` cache, so the Go cache-then-fetch fast path (lines 1234–1286) collapses to a single upstream call.

## Test plan

- [x] `cargo test -p pluto-core --all-features` — 379/379 passing (11 new tests covering the validators handler and its helpers: `convert_validators` happy-path / ignore-not-found pass-through / unknown-without-ignore rejection / malformed upstream pubkey; `invert_pub_share_map` round-trip; end-to-end happy path; pubkey-only filter pass-through of strangers; index-filter rejection of strangers; unknown input pubshare rejection (400); upstream timeout (504); malformed upstream pubkey (502); upstream 400 propagation without body leakage).


emlautarom1 (commented):
Can you merge main into this branch? It's currently stale and makes it hard to review

crates/core/src/validatorapi/component.rs
```diff
@@ -881,9 +882,91 @@ impl Handler for Component {
...
+                // Status filter is not exposed by Pluto's `ValidatorsOpts`; the
+                // Go reference also omits it from the upstream call.
+                statuses: None,
+            },
+        };
+
+        let response = tokio::time::timeout(
+            UPSTREAM_REQUEST_TIMEOUT,
+            self.eth2_cl.post_state_validators(request),
+        )
```
emlautarom1: Similar to #459, I don't think it's a good idea to have these timeouts locally on these handlers. Cancellation should be handled higher-up.

crates/core/src/validatorapi/component.rs
```diff
@@ -881,9 +882,91 @@ impl Handler for Component {
...
+                ));
+            }
+            other @ (PostStateValidatorsResponse::InternalServerError(_)
+            | PostStateValidatorsResponse::Unknown) => {
+                return Err(upstream_unexpected("validators", other));
+            }
+        };
+
+        // The `ignoreNotFound` flag mirrors the Go contract: when the caller
+        // filtered by indices only (no pubkeys), the upstream may return
```
emlautarom1: The comment is misleading, it is the other way around: when indices are provided return only those validators from the cluster's share map.

emlautarom1 (commented):
Similar comments to other validator API PRs where we have internal timeouts.

crates/core/src/validatorapi/component.rs
varex83agent: Done — dropped the local timeout around the upstream `post_state_validators` call (and removed the now-obsolete `validators_upstream_timeout_returns_504` test). The handler now awaits the upstream directly. The other upstream timeouts in `component.rs` came from earlier merged PRs, so I left those for a separate sweep rather than expanding this diff.

crates/core/src/validatorapi/component.rs
varex83agent: Fixed — the comment was inverted. Corrected it (and the matching `convert_validators` doc): when indices are provided, every returned validator must be in this cluster's share map so an unknown pubkey is rejected; when no indices are provided, unknown validators pass through with their root pubkey untouched.

varex83agent:
Merged `main` and addressed review (a04cb57):

- **Stale branch** — merged latest `main` (resolved conflicts against the #459 selections work; net diff vs `main` is just the validators handler).
- **Local timeout** — removed the `UPSTREAM_REQUEST_TIMEOUT` wrapper around the upstream `post_state_validators` call; the handler awaits directly now. Dropped the obsolete `validators_upstream_timeout_returns_504` test. (Scoped to this handler — the remaining upstream timeouts came from earlier merged PRs and are better swept separately.)
- **Misleading comment** — the `ignore_not_found` comment was inverted; corrected it and the matching `convert_validators` doc to: indices provided ⇒ unknown pubkey rejected; no indices ⇒ unknown validators pass through untouched.

`cargo +nightly fmt`, `clippy -D warnings`, and `cargo test -p pluto-core` (439 passing) all green.

emlautarom1 (approved):
LGTM

PR #458, opened 2026-05-29
feat(core): wire up validatorapi Component subscribe/await/verify hooks

varex83agent:
## Summary

- Adds the validatorapi Component plumbing every subsequent submit/await handler needs. No `unimplemented!()` arms are filled in — this PR only exposes the hooks. Mirrors Charon's `core/validatorapi/validatorapi.go:196-256` and `:1352`.
- New `Component` callback fields (six total — `await_proposal_fn`, `await_agg_attestation_fn`, `await_sync_contribution_fn`, `await_agg_sig_db_fn`, `duty_def_fn`, `pub_key_by_att_fn`) plus a `subs: Vec<SubscriberFn>` subscriber list. All stored as `Option<Arc<…>>` so registration happens before the `Component` is shared.
- `subscribe()` wraps the user closure with a set-clone step so each subscriber receives its own `ParSignedDataSet`, mirroring Go's clone-before-fanout at `validatorapi.go:249-256`.
- Six `register_*` methods overwrite on re-register, matching Go's single-function input semantics.
- `verify_partial_sig()` honours `insecure_test`, looks up this node's public share via the existing `pub_share_by_pubkey` map, then delegates to `pluto_eth2util::signing::verify`. The Rust hook takes `(domain_name, epoch, message_root, signature)` directly rather than projecting them through `core.Eth2SignedData` — the trait hasn't been ported yet, so submit handlers in PRs 3-6 will derive the triple from their concrete signed-data wrapper.

## Go reference

| Go (`charon/core/validatorapi/validatorapi.go`) | Rust (`crates/core/src/validatorapi/component.rs`) |
| --- | --- |
| `subs []func(...) error` field, `Subscribe` :249-259 | `subs: Vec<SubscriberFn>`, `Component::subscribe` |
| `awaitProposalFunc` field, `RegisterAwaitProposal` :208 | `await_proposal_fn`, `Component::register_await_proposal` |
| `awaitAggAttFunc` field, `RegisterAwaitAggAttestation` :238 | `await_agg_attestation_fn`, `Component::register_await_agg_attestation` |
| `awaitSyncContributionFunc` field, `RegisterAwaitSyncContribution` :220 | `await_sync_contribution_fn`, `Component::register_await_sync_contribution` |
| `awaitAggSigDBFunc` field, `RegisterAwaitAggSigDB` :243 | `await_agg_sig_db_fn`, `Component::register_await_agg_sig_db` |
| `dutyDefFunc` field, `RegisterGetDutyDefinition` :232 | `duty_def_fn`, `Component::register_get_duty_definition` |
| `pubKeyByAttFunc` field, `RegisterPubKeyByAttestation` :226 | `pub_key_by_att_fn`, `Component::register_pub_key_by_attestation` |
| `verifyPartialSig` :1352 | `Component::verify_partial_sig` |

## Out of scope

- The Eth2SignedData trait + `core.VerifyEth2SignedData` are not ported here. `verify_partial_sig` instead takes the BLS domain/epoch/message-root directly. Submit handlers (PRs 3-6) compute those from their concrete signed-data wrapper before calling the hook.
- `Handler` trait signatures are unchanged. No other crates touched.
- No cluster accessor added — the existing `pub_share_by_pubkey: HashMap<BLSPubKey, BLSPubKey>` on `Component` already provides the DV-root → public-share lookup that Go's `getVerifyShareFunc` exposes, so no `crates/cluster/` change was needed.

## Test plan

- [x] `cargo test -p pluto-core --all-features` — 374/374 passing (8 new tests: subscribe fanout-clones, register_await_proposal overwrite, the remaining five register hooks overwrite, unregistered hooks default to None, verify_partial_sig accept/reject valid+tampered, reject unknown pubkey, short-circuit in insecure_test)

### DIFF
--- a/crates/core/src/validatorapi/component.rs
+++ b/crates/core/src/validatorapi/component.rs
@@ -810,7 +810,12 @@ fn format_bls_pubkey(pubkey: &BLSPubKey) -> String {
 
 #[cfg(test)]
 mod tests {
+    use std::sync::Mutex;
+
     use chrono::{DateTime, Utc};
+    use pluto_crypto::{blst_impl::BlstImpl, tbls::Tbls};
+    use pluto_testutil::BeaconMock;
+    use serde_json::json;
     use tokio::sync::mpsc;
     use tokio_util::sync::CancellationToken;
 
@@ -820,9 +825,10 @@ mod tests {
         dutydb::{UnsignedDataSet, UnsignedDutyData},
         signeddata::{
             AttestationData as SignedAttestationData, AttesterDuty as SignedAttesterDuty,
+            SignedRandao, SyncContribution, VersionedAggregatedAttestation,
         },
         testutils::random_core_pub_key,
-        types::{Duty, DutyType, SlotNumber},
+        types::{Duty, DutyType, PubKey, SlotNumber},
         validatorapi::types::AttestationDataOpts,
     };
 
@@ -1255,18 +1261,6 @@ mod tests {
     // ====================================================================
     // Plumbing tests — Subscribe / Register* / verify_partial_sig
     // ====================================================================
-
-    use std::sync::Mutex;
-
-    use pluto_crypto::{blst_impl::BlstImpl, tbls::Tbls};
-    use pluto_eth2util::signing::DomainName;
-    use pluto_testutil::BeaconMock;
-    use serde_json::json;
-
-    use crate::{
-        signeddata::{SignedRandao, SyncContribution, VersionedAggregatedAttestation},
-        types::PubKey,
-    };
 
     fn dv_pubkey(byte: u8) -> BLSPubKey {
         [byte; 48]
--- a/crates/core/src/validatorapi/component.rs
+++ b/crates/core/src/validatorapi/component.rs
@@ -4,10 +4,11 @@
 //! and public-share mappings needed to translate between distributed-validator
 //! root keys and this node's threshold-BLS share.
 
-use std::{any::Any, collections::HashMap, future::Future, pin::Pin, sync::Arc, time::Duration};
+use std::{any::Any, collections::HashMap, future::Future, sync::Arc, time::Duration};
 
 use async_trait::async_trait;
 use axum::http::StatusCode;
+use futures::future::BoxFuture;
 use pluto_eth2api::{
     EthBeaconNodeApiClient, GetAttesterDutiesRequest, GetAttesterDutiesResponse,
     GetProposerDutiesRequest, GetProposerDutiesResponse, GetSyncCommitteeDutiesRequest,
@@ -46,16 +47,13 @@ use crate::{
 /// `error` return for `awaitX` / `dutyDef` / `pubKeyByAtt` callbacks.
 pub type CallbackError = Box<dyn std::error::Error + Send + Sync + 'static>;
 
-/// Convenience alias for the future returned by an async registered callback.
-type CallbackFuture<'a, T> = Pin<Box<dyn Future<Output = Result<T, CallbackError>> + Send + 'a>>;
-
 /// Subscriber callback for `Subscribe`. Receives the [`Duty`] and the
 /// [`ParSignedDataSet`] by reference; the registered wrapper clones the
 /// set exactly once before invoking the user closure so every subscriber
 /// observes an independent copy — mirroring Go's `Subscribe`
 /// clone-before-fanout behaviour at `validatorapi.go:249-256`.
 pub type SubscriberFn = Arc<
-    dyn for<'a> Fn(&'a Duty, &'a ParSignedDataSet) -> CallbackFuture<'a, ()>
+    dyn for<'a> Fn(&'a Duty, &'a ParSignedDataSet) -> BoxFuture<'a, Result<(), CallbackError>>
         + Send
         + Sync
         + 'static,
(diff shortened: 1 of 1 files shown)


varex83agent:
## /loop-review-pr summary

Ran 2 review-and-fix iteration(s) against this PR.
Terminated by: **completion_promise**.

### Pre-loop housekeeping
The branch was rebased onto `origin/main` via `git rebase --onto origin/main 5da9093 bohdan/validatorapi-plumbing`. Reason: PR #451 was squash-merged into main as `e75a114`, so the 5 ancestor commits this branch carried (5da9093, 26675fe, 11ad940, 6fe5785, 0767f40) were already in main and a plain `git rebase origin/main` produced spurious conflicts against the squashed version. After the `--onto` rebase the PR diff is a focused +627 / -4 in `crates/core/src/validatorapi/component.rs`.

### Quality gates (final)
- `cargo +nightly fmt --all --check` — pass
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` — pass
- `cargo test --workspace --all-features` — pass (`pluto-core` 413 unit tests, exit 0)

### Resolved during the loop

**Major (1)**
- subscribe wrapper double-clones `ParSignedDataSet` — `crates/core/src/validatorapi/component.rs:55` — fix in 8bfd2be. `SubscriberFn` now takes `&ParSignedDataSet`; the wrapper is the sole clone site, mirroring Go's `Subscribe` clone-before-fanout cost of one clone per subscriber.

**Minor (3)**
- Reinvented `Pin<Box<dyn Future>>` alias — `crates/core/src/validatorapi/component.rs:50` — fix in 9b3040c. Dropped the local `CallbackFuture` alias and adopted `futures::future::BoxFuture` across all 7 callback type aliases, matching the convention already used in `consensus/qbft/transport.rs`, DKG, and peerinfo.
- `AwaitAggSigDBFn` naming — `crates/core/src/validatorapi/component.rs:81` — fix in 0195b05. Renamed to `AwaitAggSigDbFn` per Rust acronym casing (RFC 430). The neighbouring `DutyDbError` already follows this convention; `BLSPubKey` is unrelated upstream and stays as-is.
- Mid-module `use` statements in test module — `crates/core/src/validatorapi/component.rs:1259` — fix in 8eedb3f. Hoisted the second `use` block at the plumbing-tests boundary up to the existing top-of-module block and dropped the redundant `DomainName` re-import (already in scope via `super::*`).

### Deliberately deferred (documented in source / state file)
- No subscriber-fanout helper yet — submit handlers don't exist in this PR; the helper lands with the first consumer PR alongside the loop's abort-on-error semantics.
- `DutyDefFn` returns `Box<dyn Any + Send + Sync>` — already documented in source as deliberate parity with Go's untyped `core.DutyDefinitionSet`; revisit when the first downcast lands.
- Callbacks lack a `CancellationToken` parameter — future submit handlers will wrap each callsite in `tokio::time::timeout`, matching the existing `attestation_data` pattern.
- `verify_partial_sig` takes `(domain_name, epoch, message_root, signature)` directly instead of projecting through an `Eth2SignedData` trait — documented in source; tightens when the trait lands.

### Outstanding
None — internal review pipeline (pluto-review / security / rust-style / code-quality, four agents in parallel) returned zero `bug` and zero `major` findings on iteration 2 after the iter-1 fixes landed.

### Verdict

crates/core/src/validatorapi/component.rs
```diff
@@ -32,12 +35,83 @@ use super::{
...
 };
 
+/// Boxed error returned by registered callbacks.
+pub type CallbackError = Box<dyn std::error::Error + Send + Sync + 'static>;
+
+/// Subscriber callback for `Subscribe`. Receives the [`Duty`] and the
+/// [`ParSignedDataSet`] by reference; the registered wrapper clones the
+/// set exactly once before invoking the user closure so every subscriber
+/// observes an independent copy.
+pub type SubscriberFn = Arc<
```
emlautarom1, line 53: nit: Callbacks don't need to be `Arc`, `Box` should suffice.

crates/core/src/validatorapi/component.rs
```diff
@@ -991,4 +1246,357 @@ mod tests {
...
+        let dutydb = Arc::new(MemDB::new(deadliner, evict_rx, &cancel));
+        let eth2_cl =
+            Arc::new(EthBeaconNodeApiClient::with_base_url("http://127.0.0.1:0").unwrap());
+        Component::new(eth2_cl, dutydb, 1, map, false)
+    }
+
+    /// `Subscribe` invokes every registered subscriber, each receiving its
+    /// own clone of the set. Mutating one clone does not affect the others.
+    #[tokio::test]
+    async fn subscribe_fanouts_clones_to_every_subscriber() {
```
emlautarom1, line 1281: I'm not sure if we need these kind of test: subscribers get a `ParSignedDataSet` (which we could make a `type` alias instead of a separate type), so it's impossible for subscribers to alter each other from the type system perspective.

emlautarom1 (approved):
Beyond `verify_partial_sig` everything is plumbing so it's hard to get the full picture (and thus deciding what makes sense/what not).

I particularly do not like the callback mechanism where we have non trivial type signatures and we store them into structs. Once we start with the wiring I think thinks will get clearer.

@therustmonk wdyt about this? Any pattern you suggest?

PR #459, opened 2026-05-29
feat(core): implement validatorapi beacon/sync committee selections handlers

varex83agent:
## Summary

- Ports `BeaconCommitteeSelections` (`core/validatorapi/validatorapi.go:798-864`) and `SyncCommitteeSelections` (`validatorapi.go:1072-1138`) into `crates/core/src/validatorapi/component.rs`, replacing both `unimplemented!()` arms.
- Adds a `register_active_validators` hook on `Component` so the handlers can translate per-selection `validator_index` values to DV root pubkeys, mirroring Go's `c.eth2Cl.ActiveValidators(ctx)` lookup.
- Replaces the `BeaconCommitteeSelection {}` / `SyncCommitteeSelection {}` placeholder structs in `validatorapi/types.rs` with type aliases for `pluto_eth2api::v1::BeaconCommitteeSelection` / `v1::SyncCommitteeSelection`, matching the established `AttestationData = phase0::AttestationData` pattern.
- Bounds each `await_agg_sig_db` lookup with a new `SELECTIONS_AGG_SIG_DB_TIMEOUT` (~two slots) so a stalled AggSigDB cannot pin a selections request forever; bounds the `active_validators` hook with `UPSTREAM_REQUEST_TIMEOUT`.
- No secrets/upstream-body leakage into client-visible messages: verification failures, unknown validator indices, subscriber failures, and AggSigDB lookup failures all use generic strings; underlying errors are attached on `ApiError::source` for debug logs only.

## Go reference

| Endpoint                          | Go (Charon v1.7.1)                                              | Rust                                                                            |
| --------------------------------- | --------------------------------------------------------------- | ------------------------------------------------------------------------------- |
| `beacon_committee_selections`     | `core/validatorapi/validatorapi.go:798-864`                     | `Component::beacon_committee_selections` in `validatorapi/component.rs`         |
| `sync_committee_selections`       | `core/validatorapi/validatorapi.go:1072-1138`                   | `Component::sync_committee_selections` in `validatorapi/component.rs`           |
| `ActiveValidators` lookup         | `c.eth2Cl.ActiveValidators(ctx)` (called at the top of both handlers) | new `register_active_validators` hook on `Component` returning `HashMap<ValidatorIndex, BLSPubKey>` |
| `verifyPartialSig`                | `validatorapi.go:1352-1368`                                     | `Component::verify_partial_sig` (PR-1 plumbing)                                  |
| `NewPartialSignedBeaconCommitteeSelection` | `core/signeddata.go:1241`                                | `signeddata::BeaconCommitteeSelection::new_partial`                              |
| `NewPartialSignedSyncCommitteeSelection`   | `core/signeddata.go:1303`                                | `signeddata::SyncCommitteeSelection::new_partial`                                |
| `NewPrepareAggregatorDuty(slot)`           | duty fanout key                                          | `Duty::new_prepare_aggregator_duty(SlotNumber::new(slot))`                       |
| `NewPrepareSyncContributionDuty(slot)`     | duty fanout key                                          | `Duty::new_prepare_sync_contribution_duty(SlotNumber::new(slot))`                |
| `awaitAggSigDBFunc(ctx, duty, pk)`         | `validatorapi.go:849` / `:1123`                          | `await_agg_sig_db_fn` hook (PR-1 plumbing) bounded by `SELECTIONS_AGG_SIG_DB_TIMEOUT` |

## Test plan

- [x] `cargo test -p pluto-core --all-features` - 384/384 passing (10 new tests covering: happy path / multi-selection fanout+stitching / unknown validator-index 400 / verification-failure short-circuit / await timeout - for each of the two handlers)


varex83agent:
## /loop-review-pr summary

Ran 3 review-and-fix iteration(s) against this PR. Terminated by: **completion_promise** (`PR_IDEAL`).

### Quality gates (final)
- `cargo fmt`    — pass
- `cargo clippy` — pass
- `cargo test`   — pass (1387 / 0 across the workspace)

### Resolved during the loop

**Major (8)**
- Router `beacon_committee_selections` / `sync_committee_selections` were `todo!()` stubs — endpoint reachable but panicked. Wired into `state.handler` with response wrappers matching Charon's `{data: [...]}` wire shape — `crates/core/src/validatorapi/router.rs` — fix in b1a8a1d
- Per-selection `SELECTIONS_AGG_SIG_DB_TIMEOUT` instead of per-request deadline — worst case was N*24s. Switched to a single shared `tokio::time::Instant` deadline anchored at the start of the fan-out phase, via `tokio::time::timeout_at` — `crates/core/src/validatorapi/component.rs` — fix in e7179d2
- Subscriber fan-out had no timeout — a stalled subscriber could pin the request indefinitely. Each `sub(&duty, set.clone())` call now shares the same per-request deadline — `crates/core/src/validatorapi/component.rs` — fix in e7179d2
- Body-size limit not applied to selection routes (would have inherited axum's 2 MiB default). Introduced `SELECTIONS_BODY_LIMIT` (64 KiB) — `crates/core/src/validatorapi/router.rs` — fix in b1a8a1d
- No structured logging on the new error paths. Every fan-out / AggSigDB-await error branch now emits a `tracing::warn!` (timeouts) or `tracing::error!` (wiring failures) with `slot` and source error — `crates/core/src/validatorapi/component.rs` — fix in e7179d2 + b5e7b44
- No test for empty `selections == []`. Added `{beacon,sync}_committee_selections_empty_input_returns_empty_data` — fix in e7179d2
- No test for `downcast_*_selection` 500 path. Added `{beacon,sync}_committee_selections_rejects_aggsigdb_type_mismatch` — fix in e7179d2
- Malformed JSON on POST routes returned axum's plain-text 400/422/415 instead of the router's `{code, message}` envelope (mirrors the `b5eb0ca` fix already on main that this branch predates). Introduced `json_rejection_to_api_error`; rewired all four POST handlers via `Result<Json<...>, JsonRejection>`. Added envelope tests for both selection routes and the attester duties route — `crates/core/src/validatorapi/router.rs` — fix in cef07e9 + 554ce69

**Minor (7)**
- `Instant::checked_add(...).expect(...)` hardcoded "24s" in the panic message and was duplicated across both handlers — extracted `selections_deadline()` helper — fix in b5e7b44
- `beacon_committee_selections` inlined `selection.slot.tree_hash_root().0` while the sync handler used `selection.message_root()` — switched to the helper for consistency; removed the now-unused `tree_hash::TreeHash` import — fix in b5e7b44
- Hook-missing `ok_or_else` branches returned 500 without logging — added `tracing::error!` to `fetch_active_validators` and both handlers' `await_agg_sig_db_fn` checks — fix in b5e7b44
- Subscriber `Err` and AggSigDB `Err` (non-timeout) were logged at `warn` — bumped to `error` (timeouts stay at `warn` as transient peer-side issues) — fix in b5e7b44
- `selections_post` / `duties_post` were mechanical duplicates — collapsed into a single generic `bounded_post(handler, body_limit)` helper — fix in cef07e9
- `const` declarations sat between two `use` blocks in router.rs — reordered so all imports come before constants — fix in cef07e9
- `component.rs` import grouping inverted vs the workspace `std` → external → `super`/`crate` convention — reordered — fix in 554ce69

**Nits (3)**
- `tracing::warn!(slot = *slot, ...)` deref noise — destructured `for (&slot, set) in &psigs_by_slot` so the eight log fields + four `SlotNumber::new(*slot)` calls collapse to plain `slot` — fix in b5e7b44
- `TestHandler` selection methods eager-cloned the Option before unwrapping — switched all six Option-based stubs to `match self.<field>.as_ref()` with `unimplemented!` on the `None` path — fix in b5e7b44 + 554ce69
- `SELECTIONS_BODY_LIMIT` doc-comment per-entry size was inflated (~300 B claimed vs ~210-250 B actual) — tightened — fix in cef07e9

### Outstanding (intentional deferrals)

These were flagged during the loop but deferred because they sit outside this PR's stated scope or trade off poorly against the size of the change required:

- **`epoch_from_slot` per-selection upstream-spec amplification** — Each selection currently triggers a fresh `GET /eth/v1/config/spec` because `EthBeaconNodeApiClient` has no client-side cache. With the body cap at 64 KiB this is bounded at ~260 spec fetches per request rather than unbounded, but a follow-up should either cache the slots-per-epoch / signing-domain on the client or hoist the lookup to once per request. Affects security/perf only; correctness is unchanged.
- **400 vs 500 status divergence with Charon** — Pluto surfaces `validator not found` and verification failures as 400 with a generic message + attached source; Charon returns 500 via `writeError`'s `apiError` fallback. Intentional Pluto policy choice (clearer client errors); see in-line comments at the call sites.
- **Empty `selections == []` returns `{"data":[]}`** — Charon returns `{"data":null}` for the same case. The Pluto shape is more beacon-API-spec-compliant; documented in the empty-input tests.
- **Style refactors deferred** — Two selection handlers share ~90% of their code; the two `downcast_*_selection` helpers are mechanical duplicates; test fixture builders `make_selections_component_{insecure,secure}` are ~95% duplicates; `endpoint: &'static str` discriminator could be a `SelectionEndpoint` enum. None of these affect correctness.
- **Defensive `MAX_SELECTIONS_PER_REQUEST` cap** — The 64 KiB body cap already bounds the maximum element count to ~260; an explicit element-count cap would mirror `VAL_INDEXES_MAX_LEN` but adds little extra defence-in-depth.
- **`ApiJson<T>` extractor wrapper** — Would remove the repeated `let Json(x) = x.map_err(json_rejection_to_api_error)?;` lines at the top of each POST handler. Deferred for parity with the existing `query_rejection_to_api_error` pattern.
- **Multi-N shared-deadline coverage test** — All existing `*_timeout` tests use N=1, so a refactor that accidentally restored per-await timeouts would not be caught. The shared-deadline contract is encoded in `selections_deadline()` and exercised indirectly; an explicit N=3 test would lock the contract.

### Verdict

State log: `.claude/loop-review-state/pr-459.md`

crates/core/src/validatorapi/types.rs
```diff
@@ -22,8 +22,21 @@ pub use pluto_eth2api::{
...
 };
 
+/// Beacon-committee selection payload. Aliases the consensus-spec
+/// `v1::BeaconCommitteeSelection` so the handler can operate on the
+/// validator-index / slot / selection-proof tuple directly.
+pub type BeaconCommitteeSelection = V1BeaconCommitteeSelection;
+
+/// Sync-committee selection payload. Aliases the consensus-spec
+/// `v1::SyncCommitteeSelection`.
+pub type SyncCommitteeSelection = V1SyncCommitteeSelection;
```
emlautarom1: Use the underlying types directly, no reason to have these aliases. If anything, prefix them with `v1::` instead of directly importing them.

crates/core/src/validatorapi/component.rs
```diff
@@ -865,17 +948,253 @@ impl Handler for Component {
...
+                selection.clone(),
+                self.share_idx,
+            );
+
+            // The selection-proof signs the slot under
+            // `DOMAIN_SELECTION_PROOF`. `BeaconCommitteeSelection::message_root()`
+            // centralises this SSZ root computation (see
+            // `crates/eth2api/src/v1.rs`) — call it through the helper so the
+            // beacon and sync handlers share the same single source of truth
+            // for the per-variant signing root.
```
emlautarom1: nit: AI comment is not useful here

crates/core/src/validatorapi/component.rs
```diff
@@ -865,17 +948,253 @@ impl Handler for Component {
...
+        // Pull every aggregated selection back out of the AggSigDB.
+        let await_fn = self.await_agg_sig_db_fn.as_ref().ok_or_else(|| {
+            tracing::error!(
+                "beacon_committee_selections: await_agg_sig_db hook not registered — Component is half-wired"
+            );
+            ApiError::new(
+                StatusCode::INTERNAL_SERVER_ERROR,
+                "await_agg_sig_db hook not registered",
+            )
+        })?;
```
emlautarom1: question: Can the component be half-wired under normal usage? If not it seems to be bug and we should fail fast. If so, prefer `.expect()` instead.

emlautarom1 (commented):
The only thing I would look into is cancellation: right now we have an inner timeout on these new operations (`beacon_committee_selections` and `sync_committee_selections`) that only cover part of the statements inside them (flagged by agents), resulting in a complex cancellation configuration.

If you follow the `ctx` used in Charon for cancellation it traces back to `main`, and considering that we are not performing any kind of graceful shutdown my suggestion would be to drop entirely everything related to cancellation in this module.

crates/core/src/validatorapi/types.rs
varex83agent: Done — dropped the `pub type` aliases and now re-export `pluto_eth2api::v1::{BeaconCommitteeSelection, SyncCommitteeSelection}` directly under the same names (consistent with how the rest of this module re-exports eth2api types).

crates/core/src/validatorapi/component.rs
varex83agent: Removed.

crates/core/src/validatorapi/component.rs
varex83agent: Can't happen under normal usage — a missing hook is a wiring bug, so switched to `.expect("await_agg_sig_db hook must be registered before serving requests")` to fail fast instead of returning a 500.

varex83agent:
Pushed fixes for the review (8d3a00f):

- **Selection type aliases** — dropped; re-export the `v1` types directly.
- **AI comment** — removed.
- **`await_agg_sig_db` half-wired** — switched to `.expect()` to fail fast (a missing hook is a wiring bug, not a runtime condition).
- **Cancellation** — removed the selections-specific timeout machinery (`SELECTIONS_PHASE_TIMEOUT` / `selections_deadline` / `timeout_at` around the subscriber fan-out and AggSigDB awaits) plus the now-obsolete timeout tests. The two handlers now just `.await` straight through, matching Charon's ctx-only model. I scoped this to the selections code added in this PR — the remaining upstream/proposal/attestation timeouts in `component.rs` came from earlier merged PRs, so dropping those is better handled as a separate follow-up rather than ballooning this diff.

`cargo +nightly fmt`, `clippy -D warnings`, and `cargo test -p pluto-core` (414 passing) all green.

emlautarom1 (approved):
varex83 should probably check it again to be sure but LGTM.

PR #460, opened 2026-05-29
feat(core): implement validatorapi voluntary_exit + validator_registrations handlers

varex83agent:
## Summary

- `submit_voluntary_exit`: implements `validatorapi.go:752-795` — looks up
  the DV root pubkey via a new `active_validators_fn` hook, derives the
  duty slot as `slots_per_epoch * epoch`, builds the `ParSignedData`
  through `signeddata::SignedVoluntaryExit::new_partial`, verifies the
  partial signature against the `DOMAIN_VOLUNTARY_EXIT` domain at the
  exit's epoch, and fans the set out to every registered subscriber.
- `submit_validator_registrations`: implements `validatorapi.go:674-749`
  — early-returns on empty input, swallows on disabled builder mode with
  a `tracing::warn!`, iterates entries through `submit_one_registration`
  which silently skips non-DV pubkeys (mirrors Go's `swallowRegFilter`),
  derives the slot from the registration timestamp + genesis time +
  slot duration, and verifies under `DOMAIN_APPLICATION_BUILDER` at
  epoch 0 (matches Go's `VersionedSignedValidatorRegistration.Epoch`).
- Replaces the placeholder `SignedVoluntaryExit` and
  `SignedValidatorRegistration` unit structs in `validatorapi/types.rs`
  with real wrappers around `phase0::SignedVoluntaryExit` and
  `versioned::VersionedSignedValidatorRegistration`. The handler trait
  signatures are unchanged.
- Adds one new `ActiveValidatorsFn` Component hook +
  `register_active_validators` method (same PR-1 pattern: `Arc<…>`,
  overwrite-on-re-register). Drops the now-stale `#[allow(dead_code)]`
  attributes on `share_idx`, `builder_enabled`, `subs`, and
  `verify_partial_sig` now that they're all consumed.

## Go reference

| Pluto                                              | Charon Go                                            |
| -------------------------------------------------- | ---------------------------------------------------- |
| `Component::submit_voluntary_exit`                 | `Component.SubmitVoluntaryExit` (`validatorapi.go:752-795`) |
| `Component::submit_validator_registrations`        | `Component.SubmitValidatorRegistrations` (`validatorapi.go:731-749`) |
| `Component::submit_one_registration` (private)     | `Component.submitRegistration` (`validatorapi.go:674-728`) |
| `slot_from_timestamp` (private helper)             | `SlotFromTimestamp` (`validatorapi.go:41-70`)         |
| `ActiveValidatorsFn` hook                          | `c.eth2Cl.ActiveValidators(ctx)`                      |
| `signeddata::SignedVoluntaryExit::new_partial`     | `core.NewPartialSignedVoluntaryExit`                  |
| `signeddata::VersionedSignedValidatorRegistration::new_partial` | `core.NewPartialVersionedSignedValidatorRegistration` |

## Test plan

- [x] `cargo test -p pluto-core --all-features` — 383/383 passing (9 new):
  - `submit_voluntary_exit_resolves_validator_and_fanouts` (happy path),
  - `submit_voluntary_exit_rejects_unknown_validator` (400),
  - `submit_voluntary_exit_returns_500_when_hook_unregistered`,
  - `submit_voluntary_exit_rejects_bad_signature` (real BLS, 400),
  - `submit_validator_registrations_swallows_when_builder_disabled`,
  - `submit_validator_registrations_no_op_on_empty_input`,
  - `submit_validator_registrations_swallows_non_dv_pubkey`,
  - `submit_validator_registrations_happy_path_fanouts`,
  - `submit_validator_registrations_rejects_bad_signature` (real BLS, 400).


varex83agent:
## /loop-review-pr summary

Ran 3 review-and-fix iteration(s) against this PR.
Terminated by: **completion_promise (PR_IDEAL)**.

### Quality gates (final)
- `cargo +nightly fmt --all --check` — pass
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` — pass
- `cargo test --workspace --all-features` — pass (1375 passed, 0 failed)

### Resolved during the loop

**Bugs (0)**
- _(none)_

**Major (1)**
- `verify_partial_sig` per-entry domain-resolution amplification (2N upstream calls on `/spec` + `/genesis` for an N-entry batch) — `crates/core/src/validatorapi/component.rs:376` — fix in `afea16b`. Iteration 1 hoisted `fetch_slots_config` + `fetch_genesis_time` out of the loop, but `verify_partial_sig → signing::verify → get_domain` was still issuing two HTTP calls per entry. Iteration 2 added `signing::verify_with_domain` (synchronous, takes a pre-resolved `phase0::Domain`) plus `Component::verify_partial_sig_with_domain`, and `submit_validator_registrations` now resolves the `ApplicationBuilder` domain once before the loop. Total upstream cost is now bounded at 4 HTTP calls per batch regardless of N.

**Minor (4)**
- `submit_validator_registrations` per-entry upstream amplification on `fetch_slots_config` / `fetch_genesis_time` — `crates/core/src/validatorapi/component.rs:350` — fix in `2ac73b9`. Hoisted both calls out of the per-registration loop; `submit_one_registration` now takes `(slot_duration, genesis_time)` as args.
- Subscriber fan-out is not transactional (Charon parity) — `crates/core/src/validatorapi/component.rs:711` — fix in `2ac73b9`. Added a doc comment on `submit_validator_registrations` documenting the per-entry, non-transactional semantics so subscriber authors know to be idempotent.
- Inconsistent `|_: Elapsed|` closure annotation in `attestation_data` — `crates/core/src/validatorapi/component.rs:640` — fix in `2ac73b9`. Dropped the annotation and the now-unused `tokio::time::error::Elapsed` import to match the `|_|` pattern used by every other timeout site in the file.
- Builder-disabled swallow logged at `warn!` — `crates/core/src/validatorapi/component.rs:725` — fix in `2ac73b9`. Downgraded to `debug!`. Vouch-style VCs send registrations every slot, so `warn!` would be noisy in non-builder configurations.

**Nits (1)**
- `sig_to_array` no-op identity helper — `crates/core/src/validatorapi/component.rs:988` — fix in `2ac73b9`. Both `phase0::BLSSignature` and `crate::types::Signature` resolve to `[u8; 96]`; the two call sites now pass `&v1.signature` / `&exit.0.signature` directly.

### Tests added
- `verify_with_domain_accepts_valid_signature` in `crates/eth2util/src/signing.rs` — locks in that the new pre-resolved-domain path produces a bit-identical signing root to the original `verify`.
- `verify_with_domain_rejects_zero_signature` in `crates/eth2util/src/signing.rs` — mirrors the zero-signature rejection on the new entry-point.

### Outstanding (deferred / out-of-scope, not addressed by this loop)
- `dutydb/memory.rs` unbounded eviction `HashSet`s. Inherited from an earlier stacked commit (`5da9093` "fix(core): address PR #451 review feedback"); a high-water-marks bounding fix already exists on `main` as commit `6a8a94c` "fix(core): bound dutydb eviction state with high-water marks". Will resolve on rebase, so not a fix for this PR.
- `crates/core/src/validatorapi/router.rs:246` still routes `submit_validator_registrations` and `submit_exit` to `todo!()`. HTTP wiring is deliberately deferred to a follow-up PR per the PR description.
- HTTP `400` vs Charon's default `500` for several client-error paths (validator-not-found, missing V1 payload, signing-failure). Documented in the code as a deliberate Pluto improvement (non-retriable rejection without leaking upstream details).
- `slot_from_timestamp` falls back to slot 0 when timestamp is before genesis, where Charon falls back to `time.Now()`. Documented as a deliberate determinism improvement (cluster-consistent rather than per-node).
- `active_validators_fn` unregistered → 500. Pluto-specific safety feature; Charon assumes the hook is always present and would panic.

### Charon parity notes
- Iteration 2's commit message claims "Total upstream cost for any registration batch is now bounded at 3 calls." The actual count is 4 HTTP calls (`fetch_slots_config` and `fetch_domain_type` both internally invoke `fetch_spec_data`; `fetch_genesis_time` and `fetch_genesis_domain` both internally invoke `fetch_genesis_data`). The security property — O(1) regardless of N — is unaffected, which is what the fix was about. Charon's `eth2wrap` caches both endpoints; once the Pluto eth2 client gains caching, the explicit hoist can be removed without behaviour change.
- Iteration 1's commit message attributes the `warn!→debug!` downgrade to "matching Charon's `log.Debug`". Charon actually has **no log** at the builder-disabled gate (`validatorapi.go:737-739`); the `log.Debug` is on the per-entry non-DV swallow filter. Pluto's `tracing::debug!` is an extra log, not a parity restoration. Harmless at debug level.

### Verdict

crates/core/src/validatorapi/component.rs
```diff
@@ -540,6 +536,112 @@ impl Component {
...
+    /// Verifies and fans out a single builder-registration. Factored out so
+    /// [`Self::submit_validator_registrations`] can iterate over its input.
+    /// The `slot_duration`, `genesis_time`, and `builder_domain` arguments are
+    /// hoisted out of the loop so a batched request issues at most one
+    /// `fetch_slots_config`, one `fetch_genesis_time`, and one builder-domain
+    /// resolution upstream call, regardless of input size.
+    async fn submit_one_registration(
+        &self,
+        registration: SignedValidatorRegistration,
+        slot_duration: std::time::Duration,
```
iamquang95: `slot_duration: Duration`

iamquang95 (approved):
lgtm

crates/core/src/validatorapi/component.rs
```diff
@@ -967,17 +1069,118 @@ impl Handler for Component {
...
+    async fn submit_voluntary_exit(&self, exit: SignedVoluntaryExit) -> Result<(), ApiError> {
+        // Resolve the DV root pubkey for the validator index carried by the
+        // exit. The lookup runs through the per-epoch validator cache.
+        let active = self.fetch_active_validators().await?;
+
+        let validator_index = exit.0.message.validator_index;
+        let root_pubkey = active.get(&validator_index).copied().ok_or_else(|| {
+            // Bubble up as 400 so a misbehaving VC sees a non-retriable
+            // rejection without leaking upstream details.
+            ApiError::new(StatusCode::BAD_REQUEST, "validator not found")
```
iamquang95, line 1139: nit: Go use `500` instead of `400`, but i think we can keep 400 here

PR #463, opened 2026-06-01
feat(consensus): implement consensus/qbft

### LINKED ISSUE #157: Implement `core/consensus/qbft`

iamquang95:
Fix: https://github.com/NethermindEth/pluto/issues/157

This PR ports Charon’s `core/consensus/qbft` wrapper into Rust. The existing `pluto_core::qbft` algorithm remains the generic QBFT engine; this PR adds the production consensus layer around it: admission, value handling, runner lifecycle, transport/self-send, p2p wiring, decision dispatch, and parity tests.

Because Charon’s `qbft.go` is large, the Rust port is split by responsibility:

- `component.rs`: `Consensus`, constructor, `subscribe`, `start`, inbound admission, peer/pubkey lookup, instance IO map.
- `runner.rs`: `propose`, `propose_priority`, `participate`, internal propose flow, `run_instance`, async-to-core bridge, cancellation/teardown.
- `definition.rs`: core `qbft::Definition`, leader selection, decide callback, compare callback, round-change logging, timeout reason logic.
- `transport.rs`: adapter between core QBFT broadcasts and `QbftConsensusMsg`; value cache, self-send, sniffer integration.
- `p2p.rs`: QBFT libp2p protocol, handler, behaviour, framing, send/receive loop.
- `msg.rs`: Charon `msg.go` equivalent: hashing, signing, verification, wrapping/unwrapping consensus messages.
- `sniffer.rs`: Charon `sniffer.go` equivalent.

Inbound flow:
`p2p.rs` receives `QbftConsensusMsg` → calls `Consensus::handle` in `component.rs` → admission checks duty/peer/signature/hash/justification → `msg.rs` wraps protobuf into core `qbft::Msg` → message enters that duty’s receive buffer → `runner.rs` bridges it into `pluto_core::qbft`.

Outbound flow:
`pluto_core::qbft` calls broadcast callback from `definition.rs` → `transport.rs` converts core message to `QbftConsensusMsg`, resolves values, self-sends locally, records sniffer data → broadcaster sends via `p2p.rs`.

Runner flow:
`runner.rs` implements `propose`, `propose_priority`, `participate`, and `run_instance`. It gets `InstanceIo` from `component.rs`, wires local value/hash/verify channels, builds `definition.rs` + `transport.rs`, then runs the core QBFT engine.

Decision flow:
Core QBFT decides → `definition.rs` decide callback decodes the value → `component.rs` subscribers receive the decided duty/value.

Test coverage includes:

- message hashing/signing/wrapping,
- admission and validation,
- runner lifecycle,
- transport self-send and value cache,
- libp2p framing and loopback,
- full in-memory QBFT consensus scenarios,
- strategy simulator parity coverage.


iamquang95:
@claude review this PR

iamquang95:
@varex83agent  /review-pr

crates/consensus/src/qbft/runner.rs
```diff
@@ -0,0 +1,779 @@
...
+
+    let core_transport: qbft::Transport<ConsensusQbftTypes> = qbft::Transport {
+        broadcast: Box::new({
+            let transport = Arc::clone(&transport);
+            let runtime = runtime.clone();
+            let instance_ct = instance_ct.clone();
+            let transport_error = Arc::clone(&transport_error);
+            move |request| {
+                let justification = request.justification.cloned().unwrap_or_default();
+                let result = runtime.block_on(transport.broadcast(transport::BroadcastRequest {
```
varex83agent, line 336: The core `broadcast` closure runs on the `spawn_blocking` thread and calls `runtime.block_on(transport.broadcast(...))`. This is sound on a multi-threaded Tokio runtime (the call is not on a runtime worker), but it parks one blocking-pool thread for the whole broadcast — which itself awaits the external `Broadcaster` and `tokio::spawn`s the detached self-send task. On a `current_thread` runtime those spawned tasks would never make progress while this thread is parked, and a saturated blocking pool would serialize instances. Worth documenting the multi-thread-runtime requirement on `Consensus`/`run_instance` (or driving the broadcast from the async side instead of `block_on`).

crates/consensus/src/qbft/runner.rs
```diff
@@ -0,0 +1,779 @@
...
+                .map_err(|err| transport::Error::Broadcast(err.to_string()))
+        })
+    })
+}
+
+/// Converts a runner result into the channel payload shared with joiners.
+fn to_runner_result(result: &Result<()>) -> RunnerResult {
+    match result {
+        Ok(()) => Ok(()),
+        Err(err) => Err(Box::new(RunnerResultError(err.to_string())) as RunnerError),
```
varex83agent: `to_runner_result` stringifies the typed `runner::Error` into `RunnerResultError(err.to_string())`, and `wait_instance_result` (runner.rs:393) stringifies it again into `Error::RunnerResult(err.to_string())`. The structured error and its `#[source]` chain (e.g. `Core(QbftError)`, `Transport(..)`) are flattened to a bare `String` before any joiner sees it, so `source()` and `matches!` on the joined error are impossible. `RunnerError = Box<dyn StdError + Send + Sync>` can already carry the typed error; consider preserving one level of structure (e.g. `RunnerResult(Box<dyn StdError ...>)`) instead of `String`.

crates/consensus/src/qbft/qbft_run_test.rs
```diff
@@ -0,0 +1,174 @@
...
+        let consensus = Arc::new(
+            Consensus::new(Config {
+                peers: peers.clone(),
+                local_peer_idx: i64::try_from(index).expect("test peer index fits i64"),
+                privkey: component::tests::secret_key(
+                    u8::try_from(index.checked_add(1).expect("test peer index increments"))
+                        .expect("test peer index fits u8"),
+                ),
+                broadcaster,
+                compare_attestations: false,
```
varex83agent: The multi-node integration test always uses `compare_attestations: false`, so the attester-compare path — `definition.rs::compare_attester`, the `verify_rx` → `core_verify_rx` bridge, and the cross-thread `local_compare_value` cancellation-token wait — is only exercised by `definition.rs` unit tests that drive `compare` directly. The end-to-end handshake during a real decided round (a blocking callback inside the consensus loop) is never validated together. Consider an integration variant with `compare_attestations: true` and Attester duties carrying real SSZ attestation data to confirm no deadlock.

crates/consensus/src/instance.rs
```diff
@@ -171,6 +169,11 @@ impl<T> InstanceIo<T> {
             .is_ok()
     }
 
+    /// Returns true once this instance's runner has been started.
+    pub fn has_started(&self) -> bool {
+        self.running.load(Ordering::Relaxed)
```
varex83agent, line 167: nit: `has_started()` loads `running` with `Relaxed` and is consulted (component.rs / runner.rs) *after* observing the recv/value channel is closed. Because the flag and the channel are independent and Relaxed, observing channel-closed does not establish happens-before with the `running=true` store, so this may legally read a stale `false`. It is sound — worst case is a benign fallback (`InputChannelFull`/`TimeoutEnqueuingReceiveBuffer` instead of swallowing a late message) — but the doc comment states the flags "do not publish or synchronize," while these read sites do lean on the flag to interpret a channel event. A one-line note (or `Acquire`/`Release` if you want the stronger guarantee) would clarify.

crates/consensus/src/qbft/p2p.rs
```diff
@@ -0,0 +1,1408 @@
...
+            Ok(()) => Ok(()),
+            Err(error) if is_ignorable_close_error(&error) => Ok(()),
+            Err(error) => Err(error.to_string()),
+        }
+    })
+    .await;
+
+    match result {
+        Ok(Ok(())) => Ok(()),
+        Ok(Err(error)) => Err(error.to_string()),
```
varex83agent: nit: the inner future returns `Result<(), String>`, so in this `Ok(Err(error))` arm `error` is already a `String`; `Err(error.to_string())` performs a needless clone. Use `Err(error)` directly.

varex83agent (commented):
## QBFT consensus port — review

This PR ports Charon's `core/consensus/qbft` component to a new `pluto-consensus` crate (moving the existing `consensus` module out of `pluto-core`), plus the supporting libp2p adapter, transport, sniffer, round timers, and the `dutydb` unsigned-data-set SSZ/JSON decoding. Overall this is a high-quality, faithful port with strong test coverage and golden-vector parity tests. Three independent review passes (functional-equivalence vs Charon v1.7.1, security, and concurrency/state-machine) found no correctness bugs.

**Verified parity (v1.7.1):** leader election `(slot+dutyType+round)%nodes`, quorum `(2n+2)/3`, `verify_msg`/`handle` validation order and error strings, justification duty-differs check, `compare_attester` field comparison, `hash_proto` (SSZ PutBytes over deterministic proto + BTreeMap ordering), increasing/eager-double-linear/proposal timers, sniffer fields & protocol id, dutydb SSZ offsets (8/96), and `participate()` duty-type/feature gating all match the pinned Go reference.

**Security:** signature verification runs before any expensive work or instance creation; the instances map is gated by the duty gater + deadliner and reaped by the expired-cleanup loop; inbound size/timeout bounds are enforced; SSZ offset/length validation in `decode_attestation_data_ssz` is sound; no key material reaches logs/Debug; justification recursion is depth-1. No reachable issues found.

**Bugs (must-fix):** none.

**Findings:** 3 minor (a `block_on`-on-blocking-thread robustness/documentation note, error double-stringification that loses the typed source chain, and a test-coverage gap for the attester-compare end-to-end path) and 2 nits, all inline. None block merge — leaving open for author discretion.

Note (out of scope, FYI only): `crates/consensus/src/timer.rs` `linear_subsequent_round_timeout` uses `from_millis` with a comment claiming charon#4537 landed in the reference, but v1.7.1's `linearRoundTimer` still uses the bare `time.Duration(200*(round-1)+200)` (nanoseconds). This is pre-existing code relocated by the rename, not introduced here, so it is not flagged inline — worth a follow-up to confirm the deliberate deviation.

crates/consensus/src/qbft/p2p.rs
```diff
@@ -0,0 +1,1410 @@
...
+}
+
+impl Handle {
+    /// Enqueues a QBFT message for async broadcast to every non-self peer.
+    ///
+    /// The token is accepted for the shared broadcaster shape. After enqueue,
+    /// network fanout is best-effort and is not cancelled by this token.
+    pub async fn broadcast(
+        &self,
+        _ct: CancellationToken,
```
varex83: can we drop it?

emlautarom1 (commented):
Could you provide some guidance on this PR? At over 6k LOC it's hard to figure out what its is scope.

iamquang95:
@emlautarom1 I updated the PR description on the structure of the PR. Also I have to move consensus out from core to avoid the cyclic dependency

crates/core/src/dutydb/memory.rs
```diff
@@ -152,6 +161,167 @@ pub enum UnsignedDutyData {
 /// `core.UnsignedDataSet`.
 pub type UnsignedDataSet = HashMap<PubKey, UnsignedDutyData>;
 
+/// Converts an unsigned-data-set protobuf into domain unsigned duty data.
+/// Currently decodes attester data; other duty types return unsupported.
+pub fn unsigned_data_set_from_proto(
```
emlautarom1: We have a bit of a mess with (de)serialization location:
- Some of it exists on `crates/core/src/types.rs`
- Some in `crates/core/src/parsigex_codec.rs`
- Now here

We need to localize this code and use existing traits when appropriate. For example, the `AttesterDutyJson` definition has nothing to do with an in-memory implementation of `DutyDB`

crates/core/src/dutydb/mod.rs
```diff
@@ -2,4 +2,4 @@
 
 pub mod memory;
 
-pub use memory::{Error, MemDB, UnsignedDataSet, UnsignedDutyData};
+pub use memory::{Error, MemDB, UnsignedDataSet, UnsignedDutyData, unsigned_data_set_from_proto};
```
emlautarom1: As noted, `UnsignedDataSet`, `UnsignedDutyData`, `unsigned_data_set_from_proto` should not come from an implementation of `DutyDB`

crates/consensus/src/qbft/qbft_run_test.rs
```diff
@@ -0,0 +1,825 @@
...
+    ct.cancel();
+    start_ct.cancel();
+    drop(expired_txs);
+    for task in start_tasks {
+        task.await.unwrap();
+    }
+}
+
+#[tokio::test]
+async fn qbft_consensus_participate_then_late_propose() {
```
emlautarom1, line 210: nit: some things can be simplified by using `JoinSet` and `.join_all()` rather than looping and waiting.

crates/consensus/src/qbft/mod.rs
```diff
@@ -0,0 +1,28 @@
...
+pub mod msg;
+
+/// Concrete libp2p adapter for QBFT consensus messages.
+pub mod p2p;
+
+pub(crate) mod sniffer;
+pub(crate) mod transport;
+
+#[cfg(test)]
+pub(crate) static FEATURESET_TEST_LOCK: tokio::sync::Mutex<()> = tokio::sync::Mutex::const_new(());
```
emlautarom1, line 26: The shared lock is a big contortion to make the global `featureset` work. I suggest adding a comment on why this is needed.

Other PRs already make use of this so let's not change this in this PR; still we should revisit this and prefer to thread the `FeatureSet` as a dependency instead.

crates/consensus/examples/qbft.rs
```diff
@@ -0,0 +1,827 @@
...
+
+    let (decision_tx, mut decision_rx) = tokio::sync::mpsc::unbounded_channel();
+    let consensus = build_consensus(&fixture, timeout, cancel.child_token(), decision_tx)?;
+    let (qbft_behaviour, handle) = qbft::p2p::Behaviour::new(qbft::p2p::Config {
+        consensus: Arc::clone(&consensus.component),
+        p2p_context: p2p_context.clone(),
+        peers: fixture.peer_ids.clone(),
+        local_peer_id: fixture.peer_ids[fixture.local_index],
+        cancellation: cancel.child_token(),
+    })?;
```
emlautarom1: This is some tricky wiring:
- We build a `ConsensusRuntime`
- Then, we pass the `consensus.component` to the `qbft` behavior builder
- We use the resulting `Handle` to update the `ConsensusRuntime`

Can we find a better alternative?

crates/consensus/examples/qbft.rs
```diff
@@ -0,0 +1,827 @@
...
+            builder.with_gater(conn_gater).with_inner(ExampleBehaviour {
+                relay: relay_client,
+                relay_manager: RelayManager::new(relays.clone(), p2p_context),
+                mdns,
+                qbft: qbft_behaviour,
+            })
+        },
+    )?;
+
+    println!(
```
emlautarom1: For consistency, logs in example apps should use the `pluto_tracing` setup.

crates/consensus/examples/qbft.rs
```diff
@@ -0,0 +1,827 @@
...
+                duty_run.try_start(
+                    &consensus.component,
+                    &fixture,
+                    connected_cluster_peers.len(),
+                    start_after,
+                    cancel.child_token(),
+                );
+            }
+            _ = tokio::time::sleep(timeout), if !duty_run.started && !duty_run.is_complete() => {
+                bail!("timeout waiting for enough peers to start QBFT");
```
emlautarom1: This sleep is reset on every loop. If the goal is to timeout when nothing happens (by default for 60 secs) then it's OK. If the goal is to gate the entire process in a timeout then it's not working as intended.

crates/consensus/examples/qbft.rs
```diff
@@ -0,0 +1,827 @@
...
+        consensus_peers,
+        lock_hash: lock.lock_hash,
+    })
+}
+
+struct ConsensusRuntime {
+    component: Arc<qbft::Consensus>,
+    handle_slot: Arc<OnceLock<qbft::p2p::Handle>>,
+    lifecycle_task: tokio::task::JoinHandle<()>,
+}
```
emlautarom1: question: Is this `ConsensusRuntime` pattern the way we're going to interact with the consensus code or is it just for testing purposes? Does it have an equivalent Charon code?

crates/consensus/examples/qbft.rs
```diff
@@ -0,0 +1,827 @@
...
+    #[arg(long)]
+    start_after_peers: Option<usize>,
+
+    /// Maximum time to wait for connections and decision.
+    #[arg(long, default_value_t = 60)]
+    timeout_secs: u64,
+
+    /// Print discovery, relay, send, receive, and connection-error details.
+    #[arg(long, default_value_t = false)]
+    verbose_p2p: bool,
```
emlautarom1: With the `tracing` approach these logs would use `debug!`

crates/consensus/src/qbft/component.rs
```diff
@@ -0,0 +1,1304 @@
...
+
+    #[cfg(test)]
+    pub(crate) fn pubkeys(&self) -> &HashMap<i64, PublicKey> {
+        &self.pubkeys
+    }
+
+    #[cfg(test)]
+    pub(crate) fn peer_labels(&self) -> &[String] {
+        &self.peer_labels
+    }
```
emlautarom1: This is only used in tests in a trivial fashion. Remove it and if you need to check the peers expose the inner `peers` field.

crates/consensus/src/qbft/component.rs
```diff
@@ -0,0 +1,1304 @@
...
+            }
+        }
+    }
+}
+
+/// QBFT consensus component.
+pub struct Consensus {
+    peers: Vec<Peer>,
+    #[cfg(test)]
+    peer_labels: Vec<String>,
```
emlautarom1: This is a big smell. We should not need to add test-only fields to structs.

emlautarom1 (changes requested):
I did not finish reviewing the entire PR but I would like to share some partial feedback to not block.

crates/consensus/src/qbft/component.rs
```diff
@@ -0,0 +1,1304 @@
...
+    where
+        F: Fn(Duty, pbpriority::PriorityResult) -> SubscriberResult + Send + Sync + 'static,
+    {
+        self.subscribers.push(Subscriber::Priority(Box::new(fn_)));
+    }
+
+    /// Validates, wraps, and queues an inbound QBFT consensus message.
+    pub async fn handle(
+        &self,
+        ct: &CancellationToken,
```
emlautarom1: nit: for consistency, the `CancellationToken` should be last (makes reading signatures/invocations easier)

crates/consensus/src/qbft/component.rs
iamquang95: This is mostly for parity with charon, and this field is used in logs/metrics later. I added the log here in this PR so we can remove the `#[cfg(test)]`

crates/consensus/examples/qbft.rs
iamquang95: Mostly for the test, the app run later will use the Controller. I removed this struct to simplify the flow.

crates/consensus/examples/qbft.rs
iamquang95: Simplified, removed the `ConsensusRuntime`  and set handler.

crates/consensus/src/qbft/definition.rs
```diff
@@ -0,0 +1,1161 @@
...
+
+/// Returns the QBFT quorum threshold for `nodes`.
+fn quorum(nodes: i64) -> usize {
+    let quorum = nodes
+        .checked_mul(2)
+        .and_then(|nodes| nodes.checked_add(2))
+        .and_then(|nodes| nodes.checked_div(3))
+        .expect("node count permits quorum calculation");
+    usize::try_from(quorum).expect("quorum fits usize")
+}
```
emlautarom1: Duplicated from https://github.com/NethermindEth/pluto/blob/30302e718c292c8f0c8de64c45b0934a86bba5f9/crates/core/src/qbft/mod.rs#L153-L161

emlautarom1 (approved):
I have to say the code is quite convoluted in both Pluto and Charon. It seems to be the result of:

- Use of many callbacks which accept references/arc/box
- Methods on structs with different lifetimes, specially in closures
- Two type of channels (crossbeam and tokio)

I don't see any specific issue beyond some minor duplication. P2P requires an extra pair of eyes though before approving.

iamquang95:
@varex83 can you help to take a look at p2p layer

varex83:
sure, let me take a look

> @varex83 can you help to take a look at p2p layer

crates/consensus/src/qbft/p2p.rs
```diff
@@ -0,0 +1,1556 @@
...
+    },
+}
+
+#[doc(hidden)]
+#[derive(Debug)]
+pub enum FromHandler {
+    Received,
+    InboundError(String),
+    Sent { request_id: u64 },
+    SendError { request_id: u64, error: String },
```
varex83: nit: I think it would be better to make a structured errors here

crates/consensus/src/qbft/p2p.rs
```diff
@@ -0,0 +1,1556 @@
...
+where
+    S: AsyncWrite + Unpin,
+{
+    if let Err(error) = stream.close().await {
+        debug!(%error, "failed to close qbft p2p stream");
+    }
+}
+
+/// Formats libp2p timeout errors consistently.
+fn timeout_error(_error: Elapsed) -> String {
```
varex83: nit: can we use `Elapsed` / don't pass it at all?

crates/consensus/src/qbft/p2p.rs
```diff
@@ -0,0 +1,1556 @@
...
+    fn drain_commands(&mut self, cx: &mut Context<'_>) {
+        while let Poll::Ready(Some(command)) = self.cmd_rx.poll_recv(cx) {
+            self.handle_broadcast(command);
+        }
+    }
+
+    /// Fans a broadcast command out to every non-self peer.
+    fn handle_broadcast(&mut self, command: BroadcastCommand) {
+        let mut target_count = 0usize;
+        for peer_idx in 0..self.config.peers.len() {
```
varex83: I think we can switch `self.config.peer` to use `known_peers` from `P2PContext`, it will follow single source of truth patter better

varex83:
```suggestion
        for peer in self.config.peers {
```

crates/consensus/src/qbft/p2p.rs
```diff
@@ -0,0 +1,1556 @@
...
+                target_count,
+            }));
+    }
+
+    /// Sends immediately to connected peers or queues a dial first.
+    fn enqueue_send(&mut self, peer_id: PeerId, pending: PendingSend) {
+        if self.is_connected(&peer_id) {
+            self.pending_events.push_back(ToSwarm::NotifyHandler {
+                peer_id,
+                handler: NotifyHandler::Any,
```
varex83, line 551: One possible concern here - if we have multiple connection handlers (in case of multiple connection protocols), we can get to the point of sending multiple `qbft` messages

crates/consensus/src/qbft/p2p.rs
```diff
@@ -0,0 +1,1556 @@
...
+
+    /// Advertises the single QBFT stream protocol.
+    fn listen_protocol(&self) -> SubstreamProtocol<Self::InboundProtocol> {
+        SubstreamProtocol::new(ReadyUpgrade::new(protocol_id()), ())
+    }
+
+    /// Queues a behaviour send request until libp2p opens a stream.
+    fn on_behaviour_event(&mut self, event: Self::FromBehaviour) {
+        match event {
+            ToHandler::Send { request_id, msg } => self.pending_open.push_back((request_id, msg)),
```
varex83, line 315: Could be wrong here, but I think we should make a check for a duplicate send events in the connection handler, it should be created globally per peer id in the behaviour and passed to conn handlers.

crates/consensus/src/qbft/p2p.rs
```diff
@@ -0,0 +1,1556 @@
...
+    /// Fans a broadcast command out to every non-self peer.
+    fn handle_broadcast(&mut self, command: BroadcastCommand) {
+        let mut target_count = 0usize;
+        for peer_idx in 0..self.config.peers.len() {
+            let peer_id = self.config.peers[peer_idx];
+            if peer_id == self.config.local_peer_id {
+                continue;
+            }
+
+            target_count = target_count.saturating_add(1);
```
varex83: Isn't `target_count` essentially `self.config.peers.len() - 1`?

crates/consensus/src/qbft/p2p.rs
iamquang95: `Any` is actually exactly one (not all connections), so we will only send one message

crates/consensus/src/qbft/p2p.rs
iamquang95: I don't think a duplicate can happen here. Broadcast will emit only one `Send` event per peer

crates/consensus/src/qbft/p2p.rs
iamquang95: Changed implementation

PR #464, opened 2026-06-01
test: add p2p end-to-end tests

therustmonk:
The networking layer had almost no automated coverage. The existing `frostp2p_integ_test` uses a different path (`Node::new_server` + custom behaviour for FROST); the normal client path and the relay circuit hop were untested.

Two real-socket integration tests for the p2p layer:

- **`direct_tcp_connection.rs`** — two real Pluto nodes connect directly over TCP and run the full production `PlutoBehaviour`: connect both ways, exchange `identify` (`Received`), and `ping`.
- **`relay_circuit.rs`** — two clients connect *only* through an in-process relay (reservation + relayed dial), never directly.

Both log swarm errors at `warn` instead of swallowing them, so failures show the real cause, not a bare timeout.

### DIFF
--- a/crates/p2p/tests/relay_circuit.rs
+++ b/crates/p2p/tests/relay_circuit.rs
@@ -16,7 +16,6 @@
 
 use std::{fmt::Debug, time::Duration};
 
-use anyhow::{Context as _, ensure};
 use futures::StreamExt as _;
 use libp2p::{Multiaddr, PeerId, multiaddr::Protocol, relay, swarm::SwarmEvent};
 use pluto_p2p::{
@@ -73,16 +72,16 @@ fn note_swarm_event<E: Debug>(label: &str, event: &SwarmEvent<E>) {
 }
 
 #[tokio::test]
-async fn two_nodes_connect_through_relay_circuit() -> anyhow::Result<()> {
+async fn two_nodes_connect_through_relay_circuit() {
     init_logs();
 
     let relay_key = generate_insecure_k1_key(1);
     let listener_key = generate_insecure_k1_key(2);
     let dialer_key = generate_insecure_k1_key(3);
 
-    let relay_peer = peer_id_from_key(relay_key.public_key()).context("relay peer id")?;
-    let listener_peer = peer_id_from_key(listener_key.public_key()).context("listener peer id")?;
-    let dialer_peer = peer_id_from_key(dialer_key.public_key()).context("dialer peer id")?;
+    let relay_peer = peer_id_from_key(relay_key.public_key()).expect("relay peer id");
+    let listener_peer = peer_id_from_key(listener_key.public_key()).expect("listener peer id");
+    let dialer_peer = peer_id_from_key(dialer_key.public_key()).expect("dialer peer id");
 
     // --- Relay server node. ---
     let relay_config = relay::Config {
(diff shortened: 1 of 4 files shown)


therustmonk:
@claude review

varex83 (approved):

mskrzypkows (approved):

iamquang95 (approved):

crates/p2p/tests/relay_circuit.rs
```diff
@@ -0,0 +1,235 @@
...
+use tokio::time::timeout;
+use tracing_subscriber::EnvFilter;
+
+const TEST_TIMEOUT: Duration = Duration::from_secs(30);
+
+/// Returns true if `addr` is a relayed (`/p2p-circuit`) address.
+fn is_circuit_addr(addr: &Multiaddr) -> bool {
+    addr.iter()
+        .any(|proto| matches!(proto, Protocol::P2pCircuit))
+}
```
emlautarom1: nit: duplicate of https://github.com/NethermindEth/pluto/blob/ca5e9599326b1890dfa2101a4ef2ee11319d7f82/crates/p2p/src/utils.rs#L159-L162

crates/p2p/tests/relay_circuit.rs
```diff
@@ -0,0 +1,235 @@
...
+/// Installs a test tracing subscriber. Defaults to `warn` so reservation and
+/// circuit failures surface even without `RUST_LOG`; set e.g.
+/// `RUST_LOG=libp2p=debug,pluto_p2p=debug` for full transport tracing.
+fn init_logs() {
+    let filter = EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("warn"));
+    let _ = tracing_subscriber::fmt()
+        .with_env_filter(filter)
+        .with_test_writer()
+        .try_init();
+}
```
emlautarom1: I'm not sure we want to enable tracing in tests. It will generate a lot of noise that provides no value 99% of the time. If there is a regression then the test should fail loudly with an error so we can review it.

crates/p2p/tests/relay_circuit.rs
```diff
@@ -0,0 +1,235 @@
...
+
+/// Logs swarm events instead of silently dropping them: failures at `warn`, the
+/// rest at `trace`. A rejected reservation or failed circuit dial then shows up
+/// in the logs rather than as a bare timeout.
+fn note_swarm_event<E: Debug>(label: &str, event: &SwarmEvent<E>) {
```
emlautarom1: In this case, are any of these errors expected in any way? If they're not (they are all actual errors) then it should panic so we can get a better trace when debugging tests.

crates/p2p/tests/relay_circuit.rs
```diff
@@ -0,0 +1,235 @@
...
+//! 3. a *dialer* client dials the listener through that circuit address and the
+//!    two establish a relayed connection.
+//!
+//! This exercises the production [`Node`] plumbing for both the relay server
+//! and relay client paths over real sockets — the relay reservation and circuit
+//! hop, not just a direct dial.
+
+use std::{fmt::Debug, time::Duration};
+
+use anyhow::{Context as _, ensure};
```
emlautarom1: nit: I think we can drop anyhow here and use `expect`/`assert!` instead for all scenarios

emlautarom1 (commented):
I would remove the tracing setup and logging: in tests we should avoid all logs to not clutter the output. If a test fails better to do so with a clear panic/assert that will actually fail the test suite. It's hard to know what to do with a warning (should it fail? does it need attention? is this expected?)

PR #466, opened 2026-06-01
feat(core): implement `core/scheduler`

### LINKED ISSUE #176: Implement `core/scheduler`
### Description
Port the module `core/scheduler` from the reference Go implementation.

### Acceptance Criteria

emlautarom1:
Closes #176

---

Uses an Actor model to implement the `core/scheduler` module, solving some concurrency hacks that the original implementation had.

Metrics have been ported using the existing Vise patterns but are untested.


emlautarom1:
@claude review

crates/core/Cargo.toml
```diff
@@ -37,9 +38,9 @@ pluto-featureset.workspace = true
 pluto-ssz.workspace = true
 ssz.workspace = true
 tree_hash.workspace = true
+anyhow.workspace = true
```
varex83: I think we can move it back to dev dependecies

crates/core/src/scheduler.rs
```diff
@@ -0,0 +1,1523 @@
...
+            slot_broadcast: sync::broadcast::channel(100).0,
+            duty_broadcast: sync::broadcast::channel(100).0,
+            reorg_rx: sync::mpsc::channel(100).1, // A channel that never receives
+        }
+    }
+
+    /// Subscribes a callback function for triggered slots.
+    pub fn subscribe_slot(
+        &mut self,
+        f: impl Fn(&types::Slot) -> Result<()> + Send + 'static,
```
varex83: Should we support async functions here as well?

crates/core/src/scheduler.rs
```diff
@@ -0,0 +1,1523 @@
...
+        label: impl AsRef<str> + Send + 'static,
+    ) {
+        let mut rx = self.slot_broadcast.subscribe();
+
+        // TODO: We might want to return a handle so clients can `.abort()` them to drop
+        // the subscription
+        tokio::spawn(async move {
```
varex83, line 123: I would pass cancellation token here / store the handles in the struct and then on join them on drop

emlautarom1: Not needed actually: when the actor gets dropped the `rx.recv()` call returns `Err(Closed)` so it exits automatically.

I'm thinking that maybe this approach of passing callbacks is not very good, and instead we should return the `.subscribe()` result received channel and let the caller decide what to do (ex. run it as `tokio::spawn`/`tokio::spawn_blocking`, handle cancellation, etc.)

crates/core/src/scheduler.rs
emlautarom1: Changed to support async callbacks. There is a tradeoff here to make:

- If we await on each callback we can block the subscriber, meaning that it could lag behind the producer (current behavior), but ensures that callbacks are proceed in order with as single running task.
- If we run each callback on separate tasks we don't lag behind, but we have an unbounded number of running tasks and we cannot ensure ordering (ex. the callback for slot 1 could not complete before we start running the callback for 2)

Charon uses the second approach but I'm not sure it's the right one.

Note that if we just returned the `Receiver` channel to the caller then it would be a per-caller decision (more flexibility, we push the decision for later).

emlautarom1:
@claude Re-review this PR. Check for your previous review and recent comments.

crates/core/src/scheduler.rs
```diff
@@ -0,0 +1,1581 @@
...
+            .await
+            .ok_or(SchedulerError::Terminated)??;
+
+        let slot_rx = new_slot_ticker(&client, ct.clone()).await?;
+
+        let actor = SchedulerActor {
+            client: client.clone(),
+            // TODO: Figure out what to pass as `pub_keys`.
+            // In Charon, these are not used (dead code)
+            valcache: valcache::ValidatorCache::new(client.clone(), Vec::new()),
```
iamquang95: Should the `build` takes `BeaconNodeClient` instead of `client: client::EthBeaconNodeApiClient`, the `BeaconNodeClient`is `eth2wrap.Client` which have the api and valcache already.

crates/core/src/types.rs
```diff
@@ -423,74 +426,133 @@ impl AsRef<[u8]> for PubKey {
...
+    ///Index of validator in validator registry.
+    pub v_idx: u64,
+    /// The slot at which the validator must propose a block.
+    pub slot: SlotNumber,
+}
 
-impl<T> DutyDefinitionSet<T>
-where
-    T: Clone + Serialize + StdDebug,
+impl TryInto<ProposerDutyDefinition>
```
iamquang95: nit: better to implement `TryFrom` instead of `TryInto`

crates/core/src/scheduler.rs
```diff
@@ -0,0 +1,1581 @@
...
+                    .map_err(|_| pluto_eth2api::EthBeaconNodeApiClientError::UnexpectedResponse)
+            })
+            .collect::<std::result::Result<Vec<_>, _>>(),
+        _ => Err(pluto_eth2api::EthBeaconNodeApiClientError::UnexpectedResponse),
+    }?;
+
+    let mut remaining = validators
+        .iter()
+        .map(|v| (v.v_idx, true))
+        .collect::<std::collections::HashMap<_, _>>();
```
iamquang95: nit: `HashSet` should be better

crates/core/src/scheduler.rs
```diff
@@ -0,0 +1,1581 @@
...
+
+            slot_broadcast: self.slot_broadcast,
+            duty_broadcast: self.duty_broadcast,
+
+            resolved_epoch: u64::MAX,
+            duties: HashMap::new(),
+            duties_by_epoch: HashMap::new(),
+        };
+
+        let (msg_tx, msg_rx) = sync::mpsc::channel(100);
```
iamquang95: nit: `100` is repeated twice, add a const for this

crates/core/src/scheduler.rs
```diff
@@ -0,0 +1,1581 @@
...
+///
+/// The production of slots is cancelled when the provided [`CancellationToken`]
+/// is cancelled.
+async fn new_slot_ticker(
+    client: &client::EthBeaconNodeApiClient,
+    ct: CancellationToken,
+) -> Result<sync::mpsc::Receiver<types::Slot>> {
+    let genesis_time = client.fetch_genesis_time().await?;
+    let (slot_duration, slots_per_epoch) = client.fetch_slots_config().await?;
+    let slot_duration = chrono::Duration::from_std(slot_duration).expect("withing range");
```
iamquang95: within

crates/core/src/scheduler.rs
emlautarom1: Refactored to use the new `BeaconNodeClient` wrapper.

varex83 (approved):

iamquang95 (approved):

PR #468, opened 2026-06-03
test: Add ParSigEx end-to-end signature exchange test

therustmonk:
Adds the first automated end-to-end test for partial signature exchange: four nodes connect over real loopback-TCP libp2p, each signs a shared message with its own threshold-BLS share, and broadcasts the partial through the `ParSigEx` protocol.

One node then aggregates the threshold of partials it received over the network and verifies the result against the group public key, proving the exchange-and-aggregation path works end to end.

The test is self-contained (no relay, no external network) and reuses the existing `ParSigEx` behaviour as-is. Also adds `pluto-crypto`, `pluto-testutil`, and `k256` as `dev-dependencies`.


crates/parsigex/tests/parsigex_e2e.rs
```diff
@@ -0,0 +1,495 @@
...
+    }
+}
+
+/// Awaits `future`, mapping an elapsed [`TEST_TIMEOUT`] to `msg`.
+async fn with_timeout<F, T>(future: F, msg: &'static str) -> Result<T>
+where
+    F: Future<Output = Result<T>>,
+{
+    timeout(TEST_TIMEOUT, future).await.context(msg)?
+}
```
emlautarom1: nit: I don't think this helper provides much value.

crates/parsigex/tests/parsigex_e2e.rs
```diff
@@ -0,0 +1,495 @@
...
+
+/// Broadcasts one partial signature set through ParSigEx and waits for the
+/// terminal outcome.
+async fn broadcast_partial(
+    handle: Handle,
+    duty: Duty,
+    data_set: ParSignedDataSet,
+) -> parsigex::Result<u64> {
+    handle.broadcast_and_wait(duty, data_set).await
+}
```
emlautarom1: Single usage, prefer to inline it

crates/parsigex/tests/parsigex_e2e.rs
```diff
@@ -0,0 +1,495 @@
...
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Each node signs [`MSG`] with its share and broadcasts it via ParSigEx.
+    fn broadcast_all(&self) -> Result<Vec<JoinHandle<parsigex::Result<u64>>>> {
+        let duty = Duty::new(SlotNumber::new(SLOT), DutyType::Randao);
+        let mut tasks = Vec::with_capacity(self.running.len());
```
emlautarom1: This should be a `JoinSet`

crates/parsigex/tests/parsigex_e2e.rs
```diff
@@ -0,0 +1,495 @@
...
+    behaviours::pluto::PlutoBehaviourEvent,
+    config::P2PConfig,
+    p2p::{Node, NodeType},
+    p2p_context::P2PContext,
+    peer::peer_id_from_key,
+};
+use pluto_parsigex::{self as parsigex, DutyGater, Event, Handle, Verifier};
+use pluto_testutil::random::{generate_insecure_k1_key, generate_test_bls_key};
+use tokio::{
+    spawn,
```
emlautarom1: Prefer to not import standalone functions, instead use them with their module (ex. `tokio::spawn` instead of importing `spawn`)

crates/parsigex/tests/parsigex_e2e.rs
```diff
@@ -0,0 +1,495 @@
...
+#[derive(Clone)]
+struct EventSinks {
+    conn_tx: mpsc::UnboundedSender<(usize, PeerId)>,
+    listen_tx: mpsc::UnboundedSender<(usize, Multiaddr)>,
+    recv_tx: mpsc::UnboundedSender<ReceivedPartial>,
+}
+
+/// Spawns each node's swarm loop, forwarding listen/connection/receive events.
+fn spawn_nodes(bundles: Vec<NodeBundle>, sinks: &EventSinks) -> Vec<RunningNode> {
+    let mut running = Vec::with_capacity(bundles.len());
```
emlautarom1, line 393: Use `JoinSet`

crates/parsigex/tests/parsigex_e2e.rs
```diff
@@ -0,0 +1,495 @@
...
+    }
+
+    /// Collects partials addressed to `receiver` until the threshold is met.
+    async fn collect_partials(&mut self, receiver: usize) -> Result<HashMap<u64, Signature>> {
+        with_timeout(
+            self.recv_threshold_partials(receiver),
+            "timed out collecting partial signatures",
+        )
+        .await
+    }
```
emlautarom1: Inline this with the comment, it's a single call used only once

crates/parsigex/tests/parsigex_e2e.rs
```diff
@@ -0,0 +1,495 @@
...
+    let mut harness = Harness::start()?;
+    harness.connect().await?;
+    let broadcasts = harness.broadcast_all()?;
+
+    // Node 0 holds share 1; it must collect the threshold of *other* nodes'
+    // partials purely from the network before it can aggregate.
+    let received = harness.collect_partials(0).await?;
+    ensure!(
+        received.len() == THRESHOLD,
+        "node 0 should receive exactly {THRESHOLD} partials, got {}",
```
emlautarom1: As it is, this `ensure!` is not useful:
- We receive one partial signature at the time, until it receives `THRESHOLD`
- Early return happens on panic, internal error or timeout (.await?)

Instead, the timeout should be handled inside `recv_threshold_partials` so that on timeout we can actually print the number of received shares (on error/panic we propagate which is fine already)

crates/parsigex/tests/parsigex_e2e.rs
```diff
@@ -0,0 +1,495 @@
...
+        received.len()
+    );
+    ensure!(
+        !received.contains_key(&1),
+        "node 0 must not receive its own share (index 1)"
+    );
+
+    harness.aggregate_and_verify(&received)?;
+    await_broadcasts(broadcasts).await?;
```
emlautarom1, line 116: This is handled better by `JoinSet`

crates/parsigex/tests/parsigex_e2e.rs
```diff
@@ -0,0 +1,495 @@
...
+}
+
+/// Drives one node's swarm until stopped, dialing on request and forwarding
+/// listen, connection and received-partial events to `sinks`.
+async fn run_swarm(
+    mut node: Node<parsigex::Behaviour>,
+    index: usize,
+    sinks: EventSinks,
+    mut dial_rx: mpsc::UnboundedReceiver<Vec<Multiaddr>>,
+    mut stop_rx: oneshot::Receiver<()>,
```
emlautarom1: You can use a `CancellationToken` here for consistency, and then await the `join` handle at the end.

Note that this is not needed, everything will be dropped at the end of the scope

crates/parsigex/tests/parsigex_e2e.rs
emlautarom1: Also, many operations use this timeout which is quite large (30s), meaning that the full test could run for multiple minutes. Prefer a per-case timeout

crates/parsigex/tests/parsigex_e2e.rs
```diff
@@ -0,0 +1,495 @@
...
+                        let _ = sinks.conn_tx.send((index, peer_id));
+                    }
+                    SwarmEvent::Behaviour(PlutoBehaviourEvent::Inner(Event::Received {
+                        data_set,
+                        ..
+                    })) => {
+                        for data in data_set.inner().values() {
+                            if let Ok(signature) = data.signed_data.signature() {
+                                let _ = sinks.recv_tx.send((index, data.share_idx, signature));
+                            }
```
emlautarom1: We're ignoring potential errors on `.signature()`. We should probably propagate them at the caller, or panic.

emlautarom1 (changes requested):
Few things could be improved. Main issue right now is that the test swallows certain errors, and the timeout is extremely generous.

emlautarom1 (approved):
LGTM

Issue #398, opened 2026-05-12
Implement `core/consensus` controller + wrapper + debugger


varex83:
## Scope

Port the three top-level files of `charon/core/consensus/`:

- `controller.go` (~123 LOC) — instance lifecycle controller
- `wrapper.go` (~69 LOC) — protocol wrapper
- `debugger.go` (~116 LOC) — debug/inspection helpers

These are tightly coupled; one PR is acceptable. The protocol-specific submodules (qbft, instance, timer, metrics, mocks) are tracked separately under #157, #48, #158, #62, #39.

## References

- Go: https://github.com/ObolNetwork/charon/blob/main/core/consensus/
- Tests: `controller_test.go`, `wrapper_internal_test.go`, `debugger_internal_test.go`

## Acceptance

- Public API matches the Go reference shape
- All three internal tests ported and green
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean

PR #471, opened 2026-06-05
feat(consensus): implement consensus debugger + wrapper + controller

iamquang95:
Part of https://github.com/NethermindEth/pluto/issues/398

Stack on consensus/qbft

### DIFF
--- a/crates/consensus/src/debugger.rs
+++ b/crates/consensus/src/debugger.rs
@@ -81,11 +81,8 @@ pub enum Error {
     #[error("marshal proto: {0}")]
     MarshalProto(#[from] prost::EncodeError),
     /// Gzip writer failed.
-    #[error("zip proto: {0}")]
-    ZipProto(std::io::Error),
-    /// Gzip writer close failed.
-    #[error("close gzip writer: {0}")]
-    CloseGzipWriter(std::io::Error),
+    #[error("gzip: {0}")]
+    Gzip(std::io::Error),
 }
 
 impl Debugger {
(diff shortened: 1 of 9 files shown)


iamquang95:
@claude review this PR

crates/consensus/src/debugger.rs
```diff
@@ -0,0 +1,382 @@
...
+pub enum Error {
+    /// Protobuf encoding failed.
+    #[error("marshal proto: {0}")]
+    MarshalProto(#[from] prost::EncodeError),
+    /// Gzip writer failed.
+    #[error("zip proto: {0}")]
+    ZipProto(std::io::Error),
+    /// Gzip writer close failed.
+    #[error("close gzip writer: {0}")]
+    CloseGzipWriter(std::io::Error),
```
emlautarom1: nit: merge these two into a single `GzipError`

crates/consensus/src/qbft/component.rs
```diff
@@ -417,17 +421,33 @@ impl Consensus {
...
+        let expired_rx = self
+            .expired_rx
+            .lock()
+            .unwrap_or_else(PoisonError::into_inner)
+            .take();
+        let instances = Arc::clone(&self.instances);
+
         tokio::spawn(async move {
+            let Some(mut expired_rx) = expired_rx else {
+                return;
```
emlautarom1: Multiple starts usually are signs of bad wiring and should fail immediately, so I'd use an `.expect` instead of `let ... else return`

A redesign that separates building from starting could solve this issue but given our current constraints I'd leave it as is and document it as a future improvement.

emlautarom1 (approved):
LGTM, some wiring could be better but for now it does the job good enough

PR #476, opened 2026-06-08
refactor(eth2wrap): move eth2wrap to a separate crate 

varex83agent:
## Summary

- New `crates/core/src/validatorapi/validator_cache.rs` defines the `CachedValidatorsProvider` async trait. Mirrors Charon's `app/eth2wrap.CachedValidatorsProvider` — the validator-API `Component` looks up the cluster's `validator_index -> DV root pubkey` map through this trait.
- The trait lives in `pluto-core` (not in a new crate, not pulled from `pluto-app`) so the validator-API module has zero new cross-crate coupling. `pluto-app` already depends on `pluto-core`, so the existing `crates/app/src/eth2wrap/valcache.rs::ValidatorCache` can `impl` this trait at the wiring layer in a follow-up — no edits to `crates/app/` here.
- `Component` gains a required `Arc<dyn CachedValidatorsProvider>` constructor arg + a private `fetch_active_validators` helper bounded by `UPSTREAM_REQUEST_TIMEOUT` (mapping cache errors to `502 BAD_GATEWAY`). Both are `#[allow(dead_code)]` until the follow-up PRs (#459, #460, #462) consume them.
- `TestValidatorCache` test double inside `mod tests` lets unit tests supply the map directly without spinning up a beacon-node mock.

## Why

PRs #459, #460, and #462 each independently introduced an `ActiveValidatorsFn` closure hook on `Component` to map validator indices into DV root pubkeys. Charon does the same lookup through `eth2Cl.ActiveValidators(ctx)`, which is backed by `app/eth2wrap/valcache.go`. This PR is the foundation that lets all three submit-handler PRs drop their bespoke hook and call a single helper rooted in the same trait.

## Go reference

| Pluto                                              | Charon Go                                            |
| -------------------------------------------------- | ---------------------------------------------------- |
| `validator_cache::CachedValidatorsProvider` (trait)| `app/eth2wrap.CachedValidatorsProvider`              |
| `Component::fetch_active_validators` (helper)      | `c.eth2Cl.ActiveValidators(ctx)` call sites          |

## Test plan

- [x] `cargo test -p pluto-core --all-features` — 430 passing (2 new tests on `fetch_active_validators`: happy path and provider-error → 502 mapping)


crates/core/src/validatorapi/validator_cache.rs
```diff
@@ -0,0 +1,35 @@
...
+///
+/// Implementations may populate the underlying cache on demand — callers
+/// must not assume the call is non-blocking.
+#[async_trait]
+pub trait CachedValidatorsProvider: Send + Sync {
+    /// Returns the `validator_index -> DV root BLS public key` map.
+    async fn active_validators(
+        &self,
+    ) -> Result<HashMap<ValidatorIndex, BLSPubKey>, CachedValidatorsError>;
+}
```
emlautarom1: We already have this trait defined: https://github.com/NethermindEth/pluto/blob/6a965c8a8a33c0c6ead2270dd4d4124b97f051e7/crates/app/src/eth2wrap/valcache.rs#L57-L65

for the error, we also have

https://github.com/NethermindEth/pluto/blob/6a965c8a8a33c0c6ead2270dd4d4124b97f051e7/crates/app/src/eth2wrap/valcache.rs#L13-L19

I'd suggest:

- Extend the trait to make it `async` by default
- Adjust the implementers (due to `async` changes)
- Reuse the existing error type, and adjust it as needed.
- Move the modules around if needed.

emlautarom1 (changes requested):
The agent missed some important duplication on the `CacheValidatorsProvider` trait.

nit: we could also deviate from Charon and name the trait `ValidatorsProvider` since caching is an implementation detail.

varex83:
Closing as duplicate of #477

PR #486, opened 2026-06-16
feat(core): implement gater

### LINKED ISSUE #399: Implement `core/gater`
## Scope

Port `charon/core/gater.go` — the **core-layer** connection/peer gater. Distinct from the p2p-layer gater (`crates/p2p/src/gater.rs`), which is already implemented.

## References

- Go: https://github.com/ObolNetwork/charon/blob/main/core/gater.go
- Tests: `core/gater_test.go`

## Acceptance

- Public API matches the Go reference
- Tests ported and green
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean

iamquang95:
Fix https://github.com/NethermindEth/pluto/issues/399

Ports the core/gater component: a `DutyGater` that rejects duties arriving from peers whose type is invalid or whose epoch is too far in the future.

  - crates/core/src/gater.rs — `DutyGater::new(client)` fetches genesis time + slot config from the beacon node and builds the gater (default budget: 2 future epochs); `allows(&duty) -> bool` is the predicate, and `into_fn()` yields the shared `DutyGaterFn (Arc<dyn Fn(&Duty) -> bool + Send + Sync>)` that the wire components consume.
  - Canonical type — `pluto_core::gater::DutyGaterFn` is now the single source of truth; parsigex, consensus, and dkg use it directly instead of each redefining an identical alias.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1512,9 +1512,9 @@ dependencies = [
 
 [[package]]
 name = "built"
-version = "0.8.0"
+version = "0.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f4ad8f11f288f48ca24471bbd51ac257aaeaaa07adae295591266b792902ae64"
+checksum = "5c0e531d93d39c34eef561e929e8a7f86d77a5af08aac4f6d6e39976c51858e9"
 dependencies = [
  "cargo-lock",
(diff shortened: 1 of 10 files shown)


iamquang95:
@claude review this PR

varex83 (approved):

PR #489, opened 2026-06-17
feat(core): implement priority

### LINKED ISSUE #177: Implement `core/priority`
### Description
Port the module `core/priority` from the reference Go implementation.

### Acceptance Criteria

iamquang95:
Fix https://github.com/NethermindEth/pluto/issues/177

## Summary

  Ports Charon's `core/priority` protocol to Rust as new crate `crates/priority`, plus a small vendored fork of libp2p `multistream-select` so Pluto can negotiate Charon's slash-less protocol id `charon/priority/2.0.0` and interop with unmodified Charon nodes.

  The protocol resolves a **deterministic, cluster-wide priority result per duty**: peers exchange signed `PriorityMsg`s, every node computes the same result, and it's committed via QBFT. Output must be **bit-for-bit identical to Charon Go** — that drives the review risk below. ~7k insertions (~3.6k crate, ~3.4k vendored fork).

  ## Breakdown

  | Area | Go counterpart | Role |
  |---|---|---|
  | `src/calculate.rs` | `calculate.go` | **Consensus-critical** result calculation + msg validation |
  | `src/component.rs` | `component.go` | Domain types, secp256k1 sign/verify, Any/protobuf conversions |
  | `src/prioritiser.rs` | `prioritiser.go` | Per-duty exchange state machine + consensus orchestration |
  | `src/consensus.rs`, `src/error.rs` | iface / scattered errors | QBFT trait seam; typed error enum |
  | `src/p2p/*` | `p2p/sender.go`, `receive.go` | Hand-written libp2p request/response transport |
  | `tests/*` | `prioritiser_test.go` | 3-host e2e; real-TCP slash-less negotiation guard |
  | `third_party/multistream-select/` | (none) | Fork via `[patch.crates-io]`; only delta = relax leading-`/` (3 `// PLUTO PATCH:` edits in `protocol.rs`) |

  Supporting (mechanical): `qbft/msg.rs` + `p2p/utils.rs` made some fns `pub`; `deadline/calculator.rs` adds `impl ... for Arc<T>`; `Cargo.toml` member + patch.

  ## Suggested review order

  1. `PATCHES.md` + `Cargo.toml` patch + the 3 `protocol.rs` edits — the interop hack, riskiest non-obvious piece (rest of fork is upstream-identical).
  2. `calculate.rs` vs `calculate.go` — read closely, must match Go exactly.
  3. `component.rs` (signing/Any) → `prioritiser.rs` (engine) → `consensus.rs`/`error.rs` → `p2p/*` → tests.

  ## Intentional deviations (not bugs)

  `RECEIVE_TIMEOUT` 5s in `handle_request` (Go has none); `DuplicateInstance` guard (Go tolerates concurrent per-duty); `peer_id` derived from privkey; typed errors; in-process `MemoryTransport` in e2e test, slash-less negotiation proven separately over TCP.

  _Minor: `PATCHES.md` "no other files differ" misses a behaviour-neutral test removal in `length_delimited.rs:486`._


iamquang95:
@claude review this pr

varex83 (approved):
LGTM!

third_party/multistream-select/PATCHES.md
```diff
@@ -0,0 +1,54 @@
...
+slash-less single-protocol round-trip, empty-name rejection, and the empty
+`Protocols([])` round-trip (the case the property test would have caught). The
+upstream `tests/` integration tests (slash-prefixed negotiation) are unchanged.
+
+## Maintenance
+
+On a libp2p bump, re-vendor the matching multistream-select version and
+re-apply these three edits. The `[patch]` requires the version to stay
+semver-compatible with what libp2p depends on (`^0.13.0`), else cargo silently
+ignores the patch.
```
emlautarom1, line 54: This is an important observation. We should hard-code the `multistream-select` version so that the patch always gets applied.

crates/priority/src/component.rs
```diff
@@ -0,0 +1,661 @@
...
+    }
+}
+
+/// Returns the proto form of a topic proposal.
+pub(crate) fn topic_proposal_to_proto(p: &TopicProposal) -> PriorityTopicProposal {
+    PriorityTopicProposal {
+        topic: Some(string_to_any(&p.topic)),
+        priorities: p.priorities.iter().map(|s| string_to_any(s)).collect(),
+    }
+}
```
emlautarom1: nit: could use `impl From` instead of standalone function

crates/priority/src/component.rs
```diff
@@ -0,0 +1,661 @@
...
+/// Extracts the string from a structpb [`Value`].
+///
+/// Returns [`Error::TopicValueNotString`] when the value is not a structpb
+/// string.
+fn value_string(value: Value) -> Result<String> {
+    match value.kind {
+        Some(Kind::StringValue(s)) => Ok(s),
+        _ => Err(Error::TopicValueNotString),
+    }
+}
```
emlautarom1, line 216: question: these are very protobuf-specific functions. Do we have a module with helpers for it? If not, is there anything that we can use from the protobuf library to replace this code?

crates/priority/src/component.rs
```diff
@@ -0,0 +1,661 @@
...
+/// signing, hiding the underlying protobuf types.
+pub struct Component {
+    peer_id: PeerId,
+    privkey: SecretKey,
+    prioritiser: Prioritiser,
+    calculator: Arc<dyn DeadlineCalculator>,
+    /// Drives the deadliner's expired-duty channel; consumed once by [`start`].
+    ///
+    /// [`start`]: Component::start
+    expired: std::sync::Mutex<Option<tokio::sync::mpsc::Receiver<Duty>>>,
```
emlautarom1: No need to use the `Mutex<Option<...>>` pattern; prefer to pass the `expired` channel as argument to `start` or even `new_component`.

Also, consider making the `Component::start` method take `self` (own) as argument so the compiler can enforce the "Must be called exactly once" constraint.

crates/priority/src/component.rs
```diff
@@ -0,0 +1,661 @@
...
+/// behaviours use), so the priority behaviour gates against the cluster's known
+/// peers and resolves outbound dial addresses from the identify-populated peer
+/// store. Its known-peer set must cover every peer in `peers`; this is enforced
+/// here — a `peers` entry the context does not recognise returns
+/// [`Error::PeerNotInContext`]. (Without this check such a peer would be gated
+/// to a no-op handler, its exchange silently skipped, and the instance could
+/// reach consensus on a partial message set after the exchange timeout.)
+#[allow(clippy::too_many_arguments)]
+pub fn new_component(
+    ctx: CancellationToken,
```
emlautarom1: nit: `ctx` can be confused with `P2PContext`, prefer to use `ct`

crates/priority/src/error.rs
```diff
@@ -0,0 +1,167 @@
...
+    Deadline(#[source] DeadlineError),
+
+    /// The deadliner could not compute the duty's deadline (computation error
+    /// or shutdown), as distinct from the duty being expired.
+    #[error("deadline computation failed")]
+    DeadlineComputeFailed,
+
+    /// The prioritise instance's context was cancelled.
+    #[error("context canceled")]
+    Cancelled,
```
emlautarom1: There is no "context" in Pluto, replace it's references with `CancellationToken` if appropriate

crates/priority/src/calculate.rs
```diff
@@ -0,0 +1,424 @@
...
+///   - mismatching duties,
+///   - duplicate topics within a peer,
+///   - more than 1000 priorities within a topic,
+///   - duplicate priorities within a topic.
+fn validate_msgs(msgs: &[PriorityMsg]) -> Result<()> {
+    if msgs.is_empty() {
+        return Err(Error::MessagesEmpty);
+    }
+
+    let mut duty: Option<Duty> = None;
```
emlautarom1: nit: can be simplified by taking the first msg's `.duty`. Avoids mut/Option

crates/priority/src/calculate.rs
```diff
@@ -0,0 +1,424 @@
...
+#[allow(clippy::cast_possible_truncation, clippy::cast_possible_wrap)]
+const COUNT_WEIGHT: i64 = MAX_PRIORITIES as i64;
+
+/// Returns the SSZ hash root of an `Any` envelope's deterministic protobuf
+/// encoding.
+///
+/// The priority protocol treats topics and priorities as opaque `Any` values
+/// and binds equality to the encoded envelope (`type_url` + `value`), so the
+/// envelope bytes are hashed directly rather than the inner concrete message.
+fn hash_any(any: &Any) -> Result<[u8; 32]> {
```
emlautarom1: nit: we could use some type alias for the SSZ hash instead of the raw `[u8; 32]` for clarity

crates/priority/src/prioritiser.rs
```diff
@@ -0,0 +1,935 @@
...
+    /// Per-instance exchange timeout before falling back to consensus.
+    exchange_timeout: Duration,
+    /// Cluster peers participating in the protocol.
+    peers: Vec<PeerId>,
+    /// Cluster consensus over priority results.
+    consensus: Arc<dyn Consensus>,
+    /// Validates received messages (peer membership + signature).
+    msg_validator: Arc<MsgVerifier>,
+    /// Cancelled when the engine shuts down, signalling instances to stop.
+    quit: CancellationToken,
```
emlautarom1: Why do we need an inner `CancellationToken`? Is it not enough with the one we receive as parameter?

crates/priority/src/prioritiser.rs
```diff
@@ -0,0 +1,935 @@
...
+
+/// Resolves cluster-wide priorities for duties.
+#[derive(Clone)]
+pub struct Prioritiser {
+    inner: Arc<Inner>,
+    /// Output subscribers; appended via [`Prioritiser::subscribe`] before
+    /// [`Prioritiser::start`]. Wrapped so the consensus subscription can read
+    /// the current set without sharing ownership at construction time.
+    subs: Arc<Mutex<Vec<Arc<Subscriber>>>>,
+}
```
emlautarom1, line 229: I'm under the impression that this could get easier (as in less Arc/Mutex/OnceLock) if we split the work in Builder/Actor/Handler following the handler pattern. I don't see anything particularly wrong, but there is quite a lot of bookeping.

emlautarom1 (commented):
Generally LGTM, no immediate issues (p2p code requires careful attention though).

I would only ask to explore:
- Alternatives to vendoring the library. Can we point to a fork as we have done with `vise`? Any other more lightweight approach?
- There are places were there are implicit assumptions about wiring (ex. "run only once"), and other have a lot of workarounds for lifetimes and concurrency (ex. locks, arcs, etc). Maybe a different design could end up with simpler code.

If nothing provides any benefit then proceed as is.

third_party/multistream-select/PATCHES.md
iamquang95: We should not pin the `multistream-select` since later libp2p can require higher version. I added a check on CI to make sure the patched version of `multistream-select` is used.

crates/priority/src/prioritiser.rs
iamquang95: It's same with charon. `quit` is the shutdown signal for the inbound handle_request

crates/priority/src/prioritiser.rs
iamquang95: Refactored to drop OnceLock and inner Arc. I would not split into `Builder/Actor/Handler` now, it will be a bigger refactor and harder to match with Charon

crates/priority/src/component.rs
iamquang95: I can't find any helpers or libraries for that.

crates/priority/src/component.rs
iamquang95: I moved the `expired` as return on `new_component `, then later we need to put this into `start`. Which make the start can only called once.

`start` can't take `self` since we still need to call `prioritise/subscribe` later

PR #498, opened 2026-06-24
feat(app): implement health

iamquang95:
## Port Charon `app/health` to Rust

  Ports Charon's `app/health` package to `crates/app/src/health` — a background
  service that periodically scrapes process metrics, evaluates a fixed set of
  health checks over a rolling window, and publishes per-check pass/fail as the
  `app_health_checks{severity,name}` gauge (plus `app_health_metrics_high_cardinality`).

  ### What's included
  - **Engine (1:1 with Go):** metric model, label selectors, series reducers, the    9 checks, and the query function — `model.rs`, `select.rs`, `reducers.rs`,    `checks.rs`, `checker.rs`. Error strings match Charon's.
  - **vise bridge (`gatherer.rs`):** vise/`prometheus-client` exposes no structured    gather, so the gatherer encodes the registry to text in the exact format the    exporter serves (`OpenMetricsForPrometheus`) and parses it back. Queried    counter/gauge families are name-validated so stray lines can't be folded in.
  - **Metric-name parity:** reshaped `crates/tracing` log metrics to    `app_log_error_total{topic}` / `app_log_warn_total{topic}` so they match    Charon and are picked up by Charon's Grafana dashboards.

  ### Deviations from Charon 
  - `high_warning_log_rate` queries `app_log_warn_total` (the emitted name) rather    than Charon's `app_log_warning_total` — Charon's check name doesn't match its    own producer (this will be removed on later version of Charon)


iamquang95:
@claude review this PR

crates/app/src/health/model.rs
```diff
@@ -0,0 +1,64 @@
...
+
+/// A single metric sample (one time series at one scrape).
+#[derive(Debug, Clone)]
+pub struct Metric {
+    /// Labels on this series.
+    pub labels: Vec<LabelPair>,
+    /// Counter value, if this is a counter sample.
+    pub counter: Option<f64>,
+    /// Gauge value, if this is a gauge sample.
+    pub gauge: Option<f64>,
```
varex83: nit: we can either use enum here with two variants and implement `.value()` for it, or, based on usage, we can store those values in one field.

varex83 (approved):
LGTM! Solid work! 

The only part I'm a little bit unsure about it `gatherer.rs` - I think it's good as it is, but curious if it's worth using some library for parsing?

iamquang95:
The `gatherer` is self-cooked based on vise output. Since we are using vise for all metrics-related things, I think the `gatherer.rs` is good enough solution for this. There are some parser crate but they are no longer maintained. Let's keep the gatherer.rs now and revisit later.

PR #499, opened 2026-06-25
refactor: remove featureset::GLOBAL_STATE

iamquang95:
Feature state was a process-global `RwLock` (a port of charon's mutex-guarded global), which forced every feature-gated test to serialize on a lock with save/restore guards. This replaces it with an  injected `Arc<FeatureSet>` resolved once at startup. Charon never mutates the state at runtime (only `Init` + the gnosis hotfix during startup), so an immutable `Arc` is behaviorally equivalent and keeps  the read path lock-free.

  **Changes**
  - `featureset`: removed `GLOBAL_STATE`; `FeatureSet` now derives `Debug + Clone`.
  - `consensus`: `Consensus` and the round timers hold `Arc<FeatureSet>` (ConsensusParticipate, ProposalTimeout, Linear/EagerDoubleLinear).
  - `core` tracker: `TrackerService` and `InclusionCore` hold it; `incl_supported` is now a plain function (dropped the `OnceLock` cache).
  - `infosync`: `Component::new` takes `&FeatureSet`.

  Deleted all the test serialization scaffolding (`FEATURESET_TEST_LOCK`, `FeatureSetGuard`, `with_featureset`, `OnceLock`, `insert_submission`) — feature-gated tests now build their own set and run in parallel.

### DIFF
--- a/crates/core/src/tracker/analysis.rs
+++ b/crates/core/src/tracker/analysis.rs
@@ -1,13 +1,10 @@
 //! Pure analysis functions for tracker duty failure detection and peer
 //! participation accounting.
 
-use std::{
-    collections::{HashMap, HashSet},
-    sync::OnceLock,
-};
+use std::collections::{HashMap, HashSet};
 
 use pluto_eth2api::EthBeaconNodeApiClientError;
-use pluto_featureset::{Feature, GLOBAL_STATE};
+use pluto_featureset::{Feature, FeatureSet};
 use pluto_ssz::HashRoot;
 
 use crate::{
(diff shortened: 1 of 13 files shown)


iamquang95:
@claude review this PR

varex83 (approved):

crates/core/src/tracker/analysis.rs
```diff
@@ -587,15 +588,23 @@ mod tests {
         PubKey::from([byte; 48])
     }
 
+    /// Default feature set for analysis tests, matching the previous global
+    /// default (`AttestationInclusion` disabled ⇒ only proposers track chain
+    /// inclusion).
+    fn fs() -> FeatureSet {
+        FeatureSet::new()
+    }
```
emlautarom1: Inline this, adds more noise than help the reader.

crates/core/src/tracker/inclusion.rs
```diff
@@ -611,12 +599,29 @@ mod tests {
...
+
     use super::*;
     use crate::types::SlotNumber;
 
     /// Shared recorder of duties passed to a callback.
     type Rec = Arc<Mutex<Vec<Duty>>>;
 
+    /// Feature set with `AttestationInclusion` toggled, so the real
+    /// `submitted()` gate accepts attester/aggregator duties in tests.
+    fn featureset(attestation_inclusion: bool) -> FeatureSet {
```
emlautarom1: nit: we could make this return an `Arc<FeatureSet>` directly to avoid it in all call-sites

emlautarom1 (approved):
LGTM

PR #502, opened 2026-06-26
feat(app): implement stacksnipe

### LINKED ISSUE #71: Implement app/stacksnipe

iamquang95:
fix https://github.com/NethermindEth/pluto/issues/71

### DIFF
--- a/crates/app/src/stacksnipe.rs
+++ b/crates/app/src/stacksnipe.rs
@@ -93,13 +93,13 @@ impl Instance {
     /// If the `/proc` path is empty, sniping is disabled and this returns
     /// immediately. All logs emitted while running carry the `stacksnipe`
     /// topic.
+    #[tracing::instrument(
+        name = "stacksnipe",
+        level = "debug",
+        skip_all,
+        fields(topic = "stacksnipe")
+    )]
     pub async fn run(self, ct: CancellationToken) {
-        let span = tracing::debug_span!("stacksnipe", topic = "stacksnipe");
-        // Full-path trait call so it doesn't shadow the inherent `instrument`.
-        tracing::Instrument::instrument(self.run_loop(ct), span).await;
-    }
-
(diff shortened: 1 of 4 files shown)


iamquang95:
@claude review this PR

crates/app/src/stacksnipe.rs
```diff
@@ -0,0 +1,436 @@
...
+    /// Polls the `/proc` path every `interval` and reports detected stack
+    /// components through the metrics callback, until `ct` is cancelled.
+    ///
+    /// If the `/proc` path is empty, sniping is disabled and this returns
+    /// immediately. All logs emitted while running carry the `stacksnipe`
+    /// topic.
+    pub async fn run(self, ct: CancellationToken) {
+        let span = tracing::debug_span!("stacksnipe", topic = "stacksnipe");
+        // Full-path trait call so it doesn't shadow the inherent `instrument`.
+        tracing::Instrument::instrument(self.run_loop(ct), span).await;
```
varex83: nit: Can we just use `#[instrument]` macro?

crates/app/src/stacksnipe.rs
```diff
@@ -0,0 +1,436 @@
...
+        }
+    }
+}
+
+/// Scans `proc_path` for processes that look like Ethereum validator stack
+/// components.
+///
+/// This is best effort and effectively infallible: every walk or read error is
+/// silently skipped.
+fn snipe(proc_path: &Path) -> Vec<StackComponent> {
```
varex83, line 144: Should we think about windows support here? I remember we used to think about windows support, not sure this is something we want in the current situation

varex83 (approved):

crates/app/src/stacksnipe.rs
iamquang95: I think for the current state, adding windows support requires more things to do (Also need to carefully handle windows support on other places as well).

PR #512, opened 2026-07-01
fix(cluster): fix charon parity in legacy lock hashing and deserialization

iamquang95:
## Summary
  Charon-parity fixes for the legacy (v1.0–v1.5) cluster lock/definition path (from `shared/audit/cluster.md`).

  ### `fix(cluster): use distributed_validators count for legacy lock hash mixin`
  `hash_lock_legacy` mixed in `validator_addresses.len()` (via `Deref`) instead of `distributed_validators.len()` → divergent legacy `lock_hash` / `CountGreaterThanLimit` when counts differ. Matches
  `ssz.go:714`.

  ### `fix(cluster): reject distributed validator fee recipient in v1.0-v1.5 locks`
  Reject non-empty `distributed_validators[*].fee_recipient_address` on deserialize; pluto silently dropped it. Matches `lock.go:365-369,387-391`.

  ### `fix(cluster): reject non-zero operator nonce in v1.0/v1.1 definitions`
  Reject `operator.nonce != 0` (new `DefinitionError::NonZeroOperatorNonce`); pluto silently dropped it. Matches `operator.go:51`.

  ### `fix(cluster): correct timestamp gating branch in legacy definition hash`
  Fix operator-precedence in the timestamp branch to mirror `ssz.go:146-155`. No behavior change (empty write is a no-op).

### DIFF
--- a/crates/cluster/src/definition.rs
+++ b/crates/cluster/src/definition.rs
@@ -272,6 +272,10 @@ pub enum DefinitionError {
     #[error("Invalid compounding: the version does not support compounding")]
     InvalidCompounding,
 
+    /// Non-zero operator nonce (v1.0/v1.1 operators must have a zero nonce)
+    #[error("non-zero operator nonce not supported")]
+    NonZeroOperatorNonce,
+
     /// Peer not found
     #[error("Peer not in definition: {peer_id}")]
     PeerNotFound {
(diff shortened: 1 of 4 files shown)


iamquang95:
@claude review the change

varex83 (approved):

PR #513, opened 2026-07-01
fix(ssz): harden bitfield/list decoding

iamquang95:
- **`d9b6ef9` — reject over-capacity bitlists in `BitList::from_ssz_bytes`**
Decode ignored its `MAX`; over-capacity `aggregation_bits` from an untrusted beacon node decoded silently and corrupted the tree-hash signing root. Made the decoder fallible + `MAX` cap;  propagated at the eth2api/core call sites.

  - **`6036aa0` — reject malformed bitlist encodings missing the sentinel bit**
Empty buffer / zero final byte were coerced to an empty bitlist instead of erroring (a valid empty `Bitlist` is `0x01`), so the root diverged from spec decoders.

  - **`b687a28` — reject non-zero padding bits in `BitVector` decode**
Padding bits above `SIZE` flowed into `tree_hash_root` but were invisible to `bit_at`, so equal values could hash differently. Applied to both decode paths.

  - **`2d16e4d` — guard `default_hash_fn`/`with_bits` inputs and make `SszList` `MAX` mandatory**
`default_hash_fn` now errors on non-64-aligned input instead of panicking; `with_bits` asserts + documents its bit-index precondition; `SszList` `MAX` is now mandatory (no `= 0` default) to  avoid the non-spec capacity-less root.

### DIFF
--- a/crates/core/src/tracker/inclusion.rs
+++ b/crates/core/src/tracker/inclusion.rs
@@ -83,6 +83,9 @@ pub enum InclusionError {
     /// A versioned attestation carried no payload.
     #[error("missing attestation payload")]
     MissingAttestation,
+    /// An aggregation-bits field could not be SSZ-decoded.
+    #[error("decode aggregation bits")]
+    DecodeAggregationBits,
     /// A bitfield comparison failed.
     #[error(transparent)]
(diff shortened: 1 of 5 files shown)


iamquang95:
@claude review the PR

varex83 (approved):

PR #515, opened 2026-07-02
fix(eth2util): align RLP, Network, ENR & keystore

iamquang95:
- **fix(eth2util): tighten RLP length bounds check** — `from_big_endian` used `>` where charon uses strict `>=`, so edge-length long-form blobs (e.g. `[0xb8, 0x00]`) decoded to empty instead of being rejected.

  - **fix(cli): gate createcluster testnet config on `Network::is_non_zero`** — `Network::is_non_zero` was `!= default()` (true if *any* field differs); now ANDs the four identifying fields (ignoring  `capella_hard_fork`) like charon's `IsNonZero`. `create_cluster` now gates on it instead of `!TestnetConfig::is_empty()`, so a partial testnet config is rejected like charon. 

  - **fix(eth2util): match charon ENR port encoding + keystore warnings** — encode TCP/UDP ports as minimal big-endian and decode length-tolerantly, so ports < 256 round-trip with charon peers instead of  being dropped / published non-canonically; and route keystore weak-param warnings through `tracing::warn!` instead of `eprintln!`.

### DIFF
--- a/crates/cli/src/commands/create_cluster.rs
+++ b/crates/cli/src/commands/create_cluster.rs
@@ -154,15 +154,6 @@ pub struct TestnetConfig {
     pub testnet_name: Option<String>,
 }
 
-impl TestnetConfig {
-    pub fn is_empty(&self) -> bool {
-        self.testnet_name.is_none()
-            && self.fork_version.is_none()
-            && self.chain_id.is_none()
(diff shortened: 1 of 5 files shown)


iamquang95:
@claude review the PR

varex83 (approved):

PR #516, opened 2026-07-02
fix(crypto/frost): BLS/FROST primitive parity and edge-case correctness

varex83agent:
## Summary

Brings the BLS and FROST primitives into functional parity with Charon
(Go, tag `v1.7.1`) and the ObolNetwork kryptology fork on a set of
edge/soundness cases. Implements the five tasks from the
`03-crypto-bls-frost-parity-and-correctness` plan. All findings were
re-verified against the Go source before implementing.

## Changes

**T01 — `k1util::recover` recovery-id domain** (`crates/k1util`)
Charon's `Recover` only accepts recovery bytes `{0, 1, 27, 28}`; the Rust
port deferred to `RecoveryId::from_byte`, which also accepts the x-reduced
ids `2` and `3` (recovering a *different* key). Now rejects everything
outside the Charon domain up front, and the error carries the original
byte. `verify_65` inherits the fix.

**T02 — FROST round2 bcast bounds + self-broadcast** (`crates/frost`)
Matches `ObolNetwork/kryptology@v0.1.0` `dkg_round2.go`: the bcast upper
bound is `max_signers` (so a Charon-shaped broadcast map that includes the
node's own Round1Bcast is accepted), the shares upper bound stays
`max_signers-1`, and a self entry in the bcast map is skipped rather than
rejected with `InvalidParticipantId`.

**T03 — FROST round2 zero-share guard** (`crates/frost`)
Adds an explicit zero-valued Shamir share guard mirroring kryptology's
`ShamirShare.Validate` (`id == 0` / zero scalar) before the Feldman VSS
equation.

**T04 — Go Herumi empty/single aggregate parity** (`crates/crypto`)
`aggregate(&[])` now returns the serialized G2 identity signature (never an
error), and a single input is deserialized + re-serialized (canonicalized)
instead of returned verbatim, matching `tbls/herumi.go`.

**T05 — `threshold_split` contract, error mapping, misspelling** (`crates/crypto`)
Maps the `usize::try_from` failure to a new dedicated `Error::ThresholdOverflow`,
documents the retained `threshold > total` hardening (a divergence from
Charon's `threshold <= 1`-only check), fixes the misspelled `threashlod`
local, and unifies `BlstSecretKey::from_bytes` error mapping to
`Error::InvalidSecretKey` across `secret_to_public_key`,
`threshold_split_insecure`, and `sign`.

## Testing

- `cargo test --workspace --all-features` — all pass
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` — clean
- `cargo +nightly fmt --all --check` — clean

New/updated tests: recovery-id domain table test; frost round2
self-broadcast acceptance, bcast/shares bound rejection, and zero-share /
zero-id rejection tests; empty-aggregate identity + Go-fixture parity,
single-input canonicalization, malformed-single-input rejection; threshold
variant assertions, `threshold == total` validity, and consistent
invalid-secret-key error across all four entry points.

### DIFF
--- a/crates/frost/src/kryptology.rs
+++ b/crates/frost/src/kryptology.rs
@@ -413,11 +413,11 @@ pub fn round2(
     received_bcasts: &BTreeMap<u32, Round1Bcast>,
     received_shares: &BTreeMap<u32, ShamirShare>,
 ) -> Result<(Round2Bcast, KeyPackage, PublicKeyPackage), KryptologyError> {
-    // Bounds mirror ObolNetwork/kryptology@v0.1.0 dkg_round2.go.
-    // feldman.Limit == max_signers.
-    //   bcast:   threshold-1 <= len <= max_signers      (may include this node's
-    // own bcast)   p2psend: threshold-1 <= len <= max_signers - 1  (never
-    // includes self)
+    // Bounds mirror ObolNetwork/kryptology@v0.1.0 dkg_round2.go, where
+    // `feldman.Limit == max_signers`:
+    // - bcast:   threshold-1 <= len <= max_signers      (may include this node's
+    //   own Round1Bcast)
+    // - p2psend: threshold-1 <= len <= max_signers - 1  (never includes self)
     let min_received = (secret.threshold - 1) as usize;
     let bcast_max = secret.max_signers as usize;
     let shares_max = (secret.max_signers - 1) as usize;
--- a/crates/frost/src/kryptology.rs
+++ b/crates/frost/src/kryptology.rs
@@ -413,12 +413,18 @@ pub fn round2(
     received_bcasts: &BTreeMap<u32, Round1Bcast>,
     received_shares: &BTreeMap<u32, ShamirShare>,
 ) -> Result<(Round2Bcast, KeyPackage, PublicKeyPackage), KryptologyError> {
+    // Bounds mirror ObolNetwork/kryptology@v0.1.0 dkg_round2.go.
+    // feldman.Limit == max_signers.
+    //   bcast:   threshold-1 <= len <= max_signers      (may include this node's
+    // own bcast)   p2psend: threshold-1 <= len <= max_signers - 1  (never
+    // includes self)
     let min_received = (secret.threshold - 1) as usize;
-    let max_received = (secret.max_signers - 1) as usize;
+    let bcast_max = secret.max_signers as usize;
+    let shares_max = (secret.max_signers - 1) as usize;
     if received_bcasts.len() < min_received
-        || received_bcasts.len() > max_received
+        || received_bcasts.len() > bcast_max
         || received_shares.len() < min_received
-        || received_shares.len() > max_received
+        || received_shares.len() > shares_max
     {
         return Err(KryptologyError::IncorrectPackageCount);
     }
@@ -432,8 +438,13 @@ pub fn round2(
     let mut share_sum = Scalar::ZERO;
 
     for (&sender_id, bcast) in received_bcasts {
+        // Charon's getRound2Inputs may include this node's own Round1Bcast in the
+        // broadcast map. Go's Round2 skips it (`if id == dp.Id { continue }`) rather
+        // than erroring. Self's commitment is added to peer_commitments separately
+        // below, and self's share contribution is the own_share_scalar term — so the
+        // self entry must be skipped here for both verification and share summation.
         if sender_id == secret.id {
-            return Err(KryptologyError::InvalidParticipantId(sender_id));
+            continue;
         }
 
         let sender_commitment =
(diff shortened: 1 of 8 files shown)


crates/k1util/src/k1util.rs
```diff
@@ -167,18 +167,32 @@ pub fn recover(hash: &[u8], sig: &[u8]) -> Result<PublicKey> {
...
 
+    // `recovery_byte` is guaranteed to be 0 or 1 here, so `from_byte` cannot
+    // fail; keep the fallible call to avoid an unchecked construction.
     let recovery_id =
         RecoveryId::from_byte(recovery_byte).ok_or(K1UtilError::InvalidSignatureRecoveryId {
-            invalid_recovery_byte: recovery_byte,
+            invalid_recovery_byte: original_recovery_byte,
         })?;
 
     let pubkey = ecdsa::VerifyingKey::recover_from_prehash(hash, &signature, recovery_id)
```
iamquang95, line 208: recover() still rejects high-S signatures that charon accepts.

**Go side** — charon `Recover` ([`app/k1util/k1util.go:120`](https://github.com/ObolNetwork/charon/blob/v1.7.1/app/k1util/k1util.go#L120)) delegates to decred's `ecdsa.RecoverCompact`  ([`dcrd/dcrec/secp256k1/v4@v4.4.0` `ecdsa/signature.go:796`](https://github.com/decred/dcrd/blob/dcrec/secp256k1/v4.4.0/dcrec/secp256k1/ecdsa/signature.go#L796)). Its only S validation is  [`signature.go:898-905`](https://github.com/decred/dcrd/blob/dcrec/secp256k1/v4.4.0/dcrec/secp256k1/ecdsa/signature.go#L898-L905): reject `S >= group order` and `S == 0`. No low-S rule — a high-S signature recovers fine.

  **Rust side** — this call, `ecdsa::VerifyingKey::recover_from_prehash`, ends with a self-verification ([`ecdsa-0.16.9`  `recovery.rs:313` (https://github.com/RustCrypto/signatures/blob/ecdsa/v0.16.9/ecdsa/src/recovery.rs#L313)) that hits k256's `verify_prehashed`, which rejects high-S outright ([`k256-0.13.4`  `src/ecdsa.rs:203-205`](https://github.com/RustCrypto/elliptic-curves/blob/k256/v0.13.4/k256/src/ecdsa.rs#L203-L205): `if sig.s().is_high() { return Err }`)

crates/k1util/src/k1util.rs
```diff
@@ -315,6 +329,96 @@ mod tests {
...
+        let digest = hex::decode(DIGEST_1).unwrap();
+        let base_sig = sign(&key, &digest).unwrap();
+        let natural_v = base_sig[K1_REC_IDX]; // 0 or 1
+
+        // (recovery_byte, expect_ok). 27/28 are the Bitcoin-compact equivalents
+        // of 0/1. For domain validation we only assert accept/reject.
+        struct Case {
+            v: u8,
+            accepted: bool,
+        }
```
emlautarom1: Too noisy in code, not explicit enough in actual test runs. Prefer to use `test_case` to parametrize the test.

crates/frost/src/kryptology.rs
```diff
@@ -413,12 +413,18 @@ pub fn round2(
     received_bcasts: &BTreeMap<u32, Round1Bcast>,
     received_shares: &BTreeMap<u32, ShamirShare>,
 ) -> Result<(Round2Bcast, KeyPackage, PublicKeyPackage), KryptologyError> {
+    // Bounds mirror ObolNetwork/kryptology@v0.1.0 dkg_round2.go.
+    // feldman.Limit == max_signers.
+    //   bcast:   threshold-1 <= len <= max_signers      (may include this node's
+    // own bcast)   p2psend: threshold-1 <= len <= max_signers - 1  (never
+    // includes self)
```
emlautarom1: The formatting here is broken

crates/k1util/src/k1util.rs
```diff
@@ -167,18 +167,32 @@ pub fn recover(hash: &[u8], sig: &[u8]) -> Result<PublicKey> {
...
     let signature =
         Signature::from_slice(&sig[..SIGNATURE_LEN - 1]).map_err(K1UtilError::InvalidSignature)?;
 
+    // `recovery_byte` is guaranteed to be 0 or 1 here, so `from_byte` cannot
+    // fail; keep the fallible call to avoid an unchecked construction.
     let recovery_id =
         RecoveryId::from_byte(recovery_byte).ok_or(K1UtilError::InvalidSignatureRecoveryId {
-            invalid_recovery_byte: recovery_byte,
+            invalid_recovery_byte: original_recovery_byte,
         })?;
```
emlautarom1: This infallible, so use `.expect` instead.

emlautarom1 (approved):
Some nits only

PR #522, opened 2026-07-03
fix(logging): redact sensitive credentials from log output

### LINKED ISSUE #504: Implement a redact logging
### Description
Some logs in the app run can print sensitive information (auth token, username, and password).

Find a library to redact these at the log level.

Patrick-Ehimen:
Closes #504

## Summary

- **`keymanager::Client`** (`eth2util`): stores bearer token as `Secret<String>` (secrecy crate). Derived `Debug` now prints `Secret([REDACTED])` instead of the raw value. `expose_secret()` is called only at the HTTP header construction site.
- **`KeymanagerConfig`** (`dkg`): replaces derived `Debug` with a manual impl that prints `<redacted>` for `auth_token`, following the existing `LokiConfig` pattern in the tracing crate. The field stays `String` so the `bon` builder API and CLI arg tests are unchanged.
- **`bootnode::resolve_relay`** (`p2p`): adds a `redact_url()` helper that strips embedded userinfo (basic-auth credentials) from relay URLs before they appear in tracing fields.

## Test plan

- [x] `cargo test -p pluto-eth2util -- keymanager` — `client_debug_redacts_auth_token` passes
- [x] `cargo test -p pluto-dkg -- keymanager` — `keymanager_config_debug_redacts_auth_token` passes
- [x] `cargo test -p pluto-p2p -- bootnode::tests` — all 3 `redact_url_*` tests pass

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -100,7 +100,7 @@ scrypt = "0.11.0"
 subtle = "2.6"
 unicode-normalization = "0.1.25"
 zeroize = "1.8.2"
-secrecy = "0.8"
+secrecy = "0.10"
 uuid = { version = "1.19", features = ["serde", "v4"] }
 unsigned-varint = { version = "0.8", features = ["futures"] }
 serde_with = { version = "3.16", features = ["hex", "base64"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -100,6 +100,7 @@ scrypt = "0.11.0"
 subtle = "2.6"
 unicode-normalization = "0.1.25"
 zeroize = "1.8.2"
+secrecy = "0.8"
 uuid = { version = "1.19", features = ["serde", "v4"] }
 unsigned-varint = { version = "0.8", features = ["futures"] }
 serde_with = { version = "3.16", features = ["hex", "base64"] }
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -84,6 +84,7 @@ scrypt = "0.11.0"
 subtle = "2.6"
 unicode-normalization = "0.1.25"
 zeroize = "1.8.2"
+secrecy = "0.8"
```
emlautarom1: Please bump to the latest version

```suggestion
secrecy = "0.10"
```

crates/p2p/src/bootnode.rs
```diff
@@ -147,6 +148,21 @@ pub async fn new_relays(
...
+        return raw.to_owned();
+    };
+    if url.username().is_empty() && url.password().is_none() {
+        return raw.to_owned();
+    }
+    if url.set_username("").is_err() || url.set_password(None).is_err() {
+        return "<redacted>".to_owned();
+    }
+    url.to_string()
+}
```
emlautarom1: Relays do not carry credentials and we do not intend to support such feature. For reference check Charon v1.7.1: https://github.com/ObolNetwork/charon/blob/749d2d7ab0b8ace34f2e686a5af5910c8adb4dc0/p2p/bootnode.go#L155.

emlautarom1 (changes requested):
LGTM, but we could use less code given the expected usages.

emlautarom1:
Thanks for your contribution! Please take a look at the provided comments so we can proceed with the PR.

Patrick-Ehimen:
@emlautarom1 i will get right on it

emlautarom1 (approved):
LGTM, thanks for your contribution.

PR #524, opened 2026-07-06
fix: reduce hot-path allocation, cloning, and lock contention

varex83agent:
## Summary

Implements the eight hot-path optimizations from the **hot-path allocation, cloning, and lock-contention reduction** plan. Every change is perf-only and preserves functional equivalence with Charon v1.7.1 — no wire format, validation, ordering, or metric semantics change.

## Tasks

| Task | Crate | Change |
|------|-------|--------|
| **T01** | `p2p` | Add alloc-free `PeerStore::has_connection`; migrate the 5 emptiness-only `connections_to_peer(..).is_empty()` callers (qbft, frostp2p, bcast, parsigex, monitoringapi). |
| **T02** | `consensus` (qbft) | Thread one `Arc<QbftConsensusMsg>` across all peers on broadcast instead of an N-1 deep clone of the multi-MB payload; move `msg`/`justification`/`values` out of `pb_msg` on receive; `values_by_hash` now consumes `Vec<Any>`. |
| **T03** | `eth2api` | `Arc`-wrap the cached `ActiveValidators`/`CompleteValidators` maps (cheap clones via `Deref`), add a read-lock fast path on a warm cache, and release the write lock across the beacon-node fetch (re-checking on re-acquire). |
| **T04** | `core` (scheduler) | Replace the O(N) `to_entries()` status-metric scan with a per-pubkey last-status map, zeroing only the previously-reported series — mirrors Charon's `statusGauge.Reset`. |
| **T05** | `core` (scheduler) | Cache `slots_per_epoch` on the actor at build time instead of a beacon-node round-trip on every duty lookup. |
| **T06** | `parsigex` | Acquire the peer-store guard once per broadcast (using T01's `has_connection`) and share the encoded payload via `Arc<[u8]>` instead of a per-target `Vec<u8>` copy. |
| **T07** | `core` (parsigdb) | Fold the threshold check into `store` under the lock and return only the matched set via a `StoreOutcome` enum, dropping the per-store full-accumulator deep clone. |
| **T08** | `ssz` | Pass the buffer by value into `merkleize_impl` (one copy instead of two); reuse a single `Sha256` in `default_hash_fn`; drop the `put_bitlist` scratch clone. |

## Tests

New unit/regression tests added for each task (peer-store connectivity, `Arc::ptr_eq` shared-buffer assertions for qbft and parsigex, concurrent valcache misses, per-pubkey status-metric transitions, cached-`slots_per_epoch` epoch math, parsigdb duplicate/mismatch paths, and ssz golden merkleization roots).

## Quality gates

- `cargo +nightly fmt --all --check` ✅
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` ✅
- `cargo test --workspace --all-features` ✅ (all suites pass)
- `cargo deny check` ✅

### DIFF
--- a/crates/consensus/src/qbft/component.rs
+++ b/crates/consensus/src/qbft/component.rs
@@ -186,9 +186,6 @@ pub enum Error {
 }
 
 /// Canonicalizes inbound `Any` values into the hash map used by QBFT messages.
-///
-/// Consumes `values` and moves each entry into the map instead of cloning,
-/// since the caller owns the values and discards the source vector afterwards.
 pub(crate) fn values_by_hash(values: Vec<Any>) -> Result<ValueMap> {
     let mut out = ValueMap::new();
 
@@ -349,7 +346,6 @@ impl Consensus {
         mut pb_msg: pbconsensus::QbftConsensusMsg,
         ct: &CancellationToken,
     ) -> Result<()> {
-        // Verify the inner message and derive its duty (borrow only).
         let inner = pb_msg.msg.as_ref().ok_or(Error::InvalidConsensusMessage)?;
 
         self.verify_msg(inner)?;
@@ -370,8 +366,6 @@ impl Consensus {
             }
         }
 
-        // Move the validated parts out of `pb_msg` (it is dropped after this),
-        // avoiding a clone of the message, justifications, and values.
         let msg = pb_msg.msg.take().ok_or(Error::InvalidConsensusMessage)?;
         let justification = std::mem::take(&mut pb_msg.justification);
         let values = values_by_hash(std::mem::take(&mut pb_msg.values))?;
--- a/crates/consensus/src/qbft/component.rs
+++ b/crates/consensus/src/qbft/component.rs
@@ -186,16 +186,19 @@ pub enum Error {
 }
 
 /// Canonicalizes inbound `Any` values into the hash map used by QBFT messages.
-pub(crate) fn values_by_hash(values: &[Any]) -> Result<ValueMap> {
+///
+/// Consumes `values` and moves each entry into the map instead of cloning,
+/// since the caller owns the values and discards the source vector afterwards.
+pub(crate) fn values_by_hash(values: Vec<Any>) -> Result<ValueMap> {
     let mut out = ValueMap::new();
 
     for value in values {
-        let decoded = decode_supported_any(value)?;
+        let decoded = decode_supported_any(&value)?;
         let hash = match decoded {
             DecodedValue::UnsignedDataSet(inner) => msg::hash_proto(&inner)?,
             DecodedValue::PriorityResult(inner) => msg::hash_proto(&inner)?,
         };
-        out.insert(hash, value.clone());
+        out.insert(hash, value);
     }
 
     Ok(out)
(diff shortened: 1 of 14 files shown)


crates/consensus/src/qbft/component.rs
```diff
@@ -366,8 +370,12 @@ impl Consensus {
             }
         }
 
-        let values = values_by_hash(&pb_msg.values)?;
-        let wrapped = msg::Msg::new(msg.clone(), pb_msg.justification.clone(), Arc::new(values))?;
+        // Move the validated parts out of `pb_msg` (it is dropped after this),
+        // avoiding a clone of the message, justifications, and values.
```
emlautarom1: In general, these AI comments do not provide much value. They explain what is immediately happening in the code. Unless the code is not clear then prefer to drop it. The comments will eventually get stale and add confusion later.

crates/consensus/src/qbft/p2p.rs
```diff
@@ -183,7 +183,13 @@ impl Handle {
     pub async fn broadcast(&self, msg: pbconsensus::QbftConsensusMsg) -> BroadcastResult {
         let request_id = self.next_request_id.fetch_add(1, Ordering::Relaxed);
         self.cmd_tx
-            .send(BroadcastCommand { request_id, msg })
+            .send(BroadcastCommand {
+                request_id,
+                // Wrap once so the fan-out shares a single allocation across all
+                // peers rather than deep-cloning the (potentially multi-MB)
+                // payload per target, matching Charon's shared-pointer broadcast.
```
emlautarom1: Example: here, the fact that we're using an Arc is required due to the `BroadcastCommand` signature. If you need to justify its usage do it on the type definition itself, or just drop it.

crates/core/src/scheduler.rs
```diff
@@ -662,6 +675,48 @@ struct Validator {
...
+    static MAP: LazyLock<Mutex<HashMap<types::PubKey, String>>> =
+        LazyLock::new(|| Mutex::new(HashMap::new()));
+    &MAP
+}
+
+/// Submits the `validator_status` gauge for `pubkey`, emulating Charon's
+/// `statusGauge.Reset(pubkey, ...)` + `Set(1)` in O(1) per validator: it zeroes
+/// only the previously-reported series for this pubkey (when the status
+/// changed) and sets the current one to 1.
+fn submit_validator_status_metric(pubkey: &types::PubKey, status: &str) {
```
emlautarom1, line 693: We should explore if we can make this a feature in Vise itself. We already are referencing it from Git.

crates/core/src/scheduler.rs
```diff
@@ -203,10 +206,16 @@ impl SchedulerBuilder {
             .await
             .ok_or(SchedulerError::Terminated)??;
 
+        // Fetch the chain constants once at build time (the node is synced at
+        // this point) and cache `slots_per_epoch` on the actor, mirroring
+        // Charon's memoized-spec behaviour.
+        let (_slot_duration, slots_per_epoch) = client.api().fetch_slots_config().await?;
```
emlautarom1: We should explore a more general solutions: there are beacon client calls that are expected to not change throughout the lifetime of the application (ex. `fetch_slots_config`), so we could cache them at the client level.

For now this is fine, but there are multiple places where this function is invoked and we'll end up adding caching code to each one of them.

crates/eth2api/src/valcache.rs
```diff
@@ -121,24 +129,38 @@ impl ValidatorCache {
...
+        } // read lock released here
+
+        // Slow path: cache miss. Snapshot the client + pubkeys we need, then
+        // release the lock so the beacon-node round-trip does not block
+        // concurrent readers. This improves on Charon, which holds the write
+        // lock across the fetch; observable behaviour is identical (a
+        // benign double-fetch race is collapsed by the re-check below).
+        let (eth2_cl, pubkeys) = {
+            let inner = self.0.read().await;
+            (inner.eth2_cl.clone(), inner.pubkeys.clone())
```
emlautarom1: We could refactor this and extract client and pubkeys from `inner`: these values do not change so no need to put them under the lock.

emlautarom1 (approved):
LGTM, some small nits only.

PR #525, opened 2026-07-06
feat(infra): porting prometheus alert

iamquang95:
This alert will be later use on TestSmoke

### DIFF
--- a/test-infra/rules.yml
+++ b/test-infra/rules.yml
@@ -1,0 +1,44 @@
+groups:
+- name: pluto
+  rules:
+  - alert: Pluto Down
+    expr: up == 0
+    for: 15s
+    annotations:
+      description: "Pluto {{ $labels.job }} is down"
+
+  - alert: Error Log Rate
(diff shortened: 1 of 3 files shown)


test-infra/rules.yml
```diff
@@ -0,0 +1,44 @@
...
+  rules:
+  - alert: Pluto Down
+    expr: up == 0
+    for: 15s
+    annotations:
+      description: "Pluto {{ $labels.job }} is down"
+
+  - alert: Error Log Rate
+    expr: app_log_error_total > 0
+    for: 15s
```
emlautarom1, line 12: Could you elaborate from where are we getting these constraints? Are they defined in the Charon repo or did you come up with them?

iamquang95: It froms here: https://github.com/ObolNetwork/charon/blob/v1.7.1/testutil/compose/static/prometheus/rules.yml

varex83 (approved):

PR #526, opened 2026-07-06
test: Go simnet harness for charon/pluto e2e duty tests

varex83agent:
## What

Adds `test-infra/harness/`, a Go test harness that runs simnet-style end-to-end **attester duty** tests over distributed-validator clusters mixing **charon nodes** (in-process, reusing charon's own `app.Run`) and **pluto nodes** (subprocesses of the pluto binary), plus a CI workflow.

Charon `v1.7.1` (our parity reference) is consumed as a regular Go module dependency — no fork. The harness reuses its `cluster.NewForT` deterministic fixtures, `testutil/beaconmock`, `testutil/validatormock` and `testutil/relay` directly.

## Design

```
                relay (in-process libp2p)
                beaconmock (shared chain state, deterministic duties)
                    ▲ Go              ▲ HTTP (per-node gateway + submission capture)
             charon app.Run      pluto run / charon run (subprocess)
             (in-process)             ▲ validator API
                                 validatormock (harness-driven over HTTP)
```

- **Per-node HTTP beacon gateways over one shared beaconmock.** Charon's beaconmock implements dynamic behaviour (duties, validators, attestation data) only on its Go client interface; its HTTP server serves static stubs. The gateway fronts the shared mock over real HTTP so external processes get a complete beacon API, and captures each node's submissions for assertions.
- **Real p2p partial-signature exchange.** Upstream simnet uses an in-memory ParSigEx transport; the harness omits it so out-of-process nodes participate in QBFT consensus and threshold signing.
- **Assertions at the beacon boundary**: for some slot, every node must submit an attestation, and all payloads must be identical after JSON normalization (group-signed threshold aggregates). The in-process baseline additionally uses upstream-style `BroadcastCallback` assertions.

## Scenarios

| Test | Cluster | Status |
|---|---|---|
| `TestSimnetAttesterCharonInProcess` | 3× charon, in-process bmock | ✅ passes (~5s) |
| `TestSimnetAttesterCharonViaGateway` | 3× charon via HTTP gateway | ✅ passes (~4s) |
| `TestSimnetAttesterPluto` | 3× pluto | ⏭ skips until `pluto run` lands |
| `TestSimnetAttesterMixed` | 2× charon + 2× pluto, threshold 3 | ⏭ skips until `pluto run` lands |

The pluto scenarios probe `$PLUTO_BIN run --help` and **activate automatically** once the `run` command exists. The subprocess path is already validated end-to-end today: pointing `PLUTO_BIN` at a charon binary (which accepts the same flags) passes the mixed scenario in ~5s, covering the on-disk fixture layout, run flags, readiness polling, HTTP-driven validator mock and capture assertions.

## CI

`.github/workflows/go-harness.yml`:
- **charon-simnet**: charon-only scenarios, no Rust build — fast signal.
- **pluto-simnet**: builds `pluto-cli` (same steps as dkg-runner) and runs the full suite with `PLUTO_BIN` set.

## Notes

- The two `replace` directives in `go.mod` are copied from charon's own `go.mod` (Go does not propagate a dependency's replaces); they must move in lockstep when bumping the pinned charon version.
- `pluto run` requirements for the scenarios to activate are documented in the README (charon-equivalent flags + validator API on the configured address).
- Follow-ups: proposer/sync-committee scenarios (direct translation of upstream simnet's beaconmock options) and below-threshold fault injection.


varex83:
@varex83agent /review-pr

crates/relay-server/src/p2p.rs
```diff
@@ -77,6 +77,15 @@ pub async fn run_relay_p2p_node(
     let mut external_addrs = external_tcp_multiaddrs(&config.p2p_config)?;
     external_addrs.extend(external_udp_multiaddrs(&config.p2p_config)?);
 
+    // Advertise the manually configured external addresses to libp2p so they
+    // are folded into the circuit reservations we hand out. Without at least
+    // one reachable address, rust-libp2p relay clients reject the reservation
+    // with `NoAddressesInReservation`; auto-detected listen addresses are
+    // added below as they are confirmed.
+    for addr in &external_addrs {
+        node.add_external_address(addr.clone());
```
varex83agent, line 86: Manually-configured external addresses bypass private-address filtering, and rust-libp2p has no go-libp2p reservation backstop.

This loop calls `add_external_address` for every entry in `external_addrs` (built from operator-supplied `--p2p-external-ip`/`--p2p-external-hostname`) with **no** `filter_private_addrs`/is-private gate. In rust-libp2p, `add_external_address` → `ExternalAddrConfirmed` feeds `libp2p-relay`'s `AcceptReservationReq { addrs }` verbatim; libp2p-relay does **not** filter non-public addresses out of circuit reservations. go-libp2p does: `makeReservationMsg` (circuitv2/relay/relay.go) unconditionally skips `!manet.IsPublicAddr(addr)`.

Net effect: a misconfigured private/loopback external addr is advertised inside circuit reservations under Pluto, whereas charon/go-libp2p would silently drop it. The auto-detected `AddrUpdate::Add` path (line 138) **is** correctly gated by `filter_private_addrs`, so only the manual path is affected. Impact is low (it adds addresses rather than breaking connectivity, and only under misconfiguration), but consider mirroring go-libp2p's is-public guard here or documenting the divergence.

crates/relay-server/src/p2p.rs
```diff
@@ -120,15 +129,23 @@ pub async fn run_relay_p2p_node(
...
 
-                // Update listener address list
+                // Keep the ENR listener list and the libp2p external addresses
+                // (which back the circuit reservations we issue) in sync, so
+                // reserving clients always receive at least one reachable
+                // address for the same set we advertise over `/enr`.
                 match address_update {
                     AddrUpdate::Add(address) => {
+                        node.add_external_address(address.clone());
                         listeners.write().await.push(address);
```
varex83agent, line 139: nit: `listeners.push(address)` on `AddrUpdate::Add` is unconditional. If libp2p ever re-emits `NewListenAddr` for the same address (interface flaps / re-listens), `listeners` accumulates duplicates. This is masked downstream — `web.rs`'s advertised-addrs dedups the external∪listeners union via a HashSet, and `swarm.add_external_address` dedups internally — so there's no user-visible drift. Correctness is fine; a dedup-on-push (or a set) would just be tidier for a long-lived relay.

crates/tracing/src/init.rs
```diff
@@ -125,10 +126,61 @@ fn default_env_filter() -> EnvFilter {
     EnvFilter::new("info")
 }
 
+/// Selects the tracing filter directive using this precedence:
+///
+/// 1. `RUST_LOG` from the environment — the standard runtime escape hatch for
```
varex83agent, line 131: `RUST_LOG` winning over an explicitly-passed `--log-level` inverts charon's precedence. Charon resolves as explicit `--log-level` flag > `CHARON_LOG_LEVEL` env > default (`cmd/cmd.go` `bindFlags` short-circuits on `f.Changed`), so an explicit flag always beats the env var there.

The change is defensible — `RUST_LOG` is the tracing-subscriber standard and supports per-target directives (`info,pluto_p2p=debug`) that charon's single-scalar level can't express, and charon has no `RUST_LOG`. But it's a genuine precedence divergence for that env var; the doc comment should state explicitly that this intentionally differs from charon (where the explicit flag wins over the env).

crates/tracing/src/init.rs
```diff
@@ -44,10 +42,13 @@ pub struct LokiInit {
...
-    let env_filter = if let Some(override_env_filter) = config.override_env_filter.as_ref() {
-        EnvFilter::from_str(override_env_filter).unwrap_or_else(|_| default_env_filter())
-    } else {
-        EnvFilter::try_from_env("RUST_LOG").unwrap_or_else(|_| default_env_filter())
+    let rust_log = std::env::var("RUST_LOG").ok();
+    let env_filter = match resolve_filter_directive(
+        rust_log.as_deref(),
+        config.override_env_filter.as_deref(),
+    ) {
+        Some(directive) => EnvFilter::try_new(&directive).unwrap_or_else(|_| default_env_filter()),
```
varex83agent, line 50: `EnvFilter::try_new(&directive).unwrap_or_else(|_| default_env_filter())` swallows a parse error and silently drops to `info`. This preserves the pre-existing behavior (not a regression), but it's poor observability for exactly the fine-grained escape-hatch path the new doc comment advertises: an operator who typos `RUST_LOG=infp,pluto_p2p=debug` gets plain `info` with zero signal. Since tracing isn't initialized yet here, consider an `eprintln!` to stderr on the error arm before falling back. Not blocking.

crates/tracing/src/init.rs
```diff
@@ -125,10 +126,61 @@ fn default_env_filter() -> EnvFilter {
...
+    rust_log: Option<&str>,
+    override_env_filter: Option<&str>,
+) -> Option<String> {
+    if let Some(rust_log) = rust_log
+        && !rust_log.trim().is_empty()
+    {
+        return Some(rust_log.to_string());
+    }
+
+    override_env_filter.map(str::to_string)
```
varex83agent, line 151: Asymmetric blank-string handling: `rust_log` is rejected when `trim().is_empty()` (so `Some("   ")` falls through to the override), but `override_env_filter` is returned verbatim with no equivalent guard. A blank `--log-level`/`CHARON_LOG_LEVEL` (`Some("   ")`) is therefore forwarded to `EnvFilter::try_new`, fails to parse, and silently falls back to `info` — instead of cleanly returning `None` and using the default. The override is CLI-sourced so blank is unlikely, but the asymmetry is surprising given both are `Option<&str>` directives. Either apply the same `trim().is_empty()` filter to the override, or document the difference.

crates/tracing/src/init.rs
```diff
@@ -125,10 +126,61 @@ fn default_env_filter() -> EnvFilter {
...
+    #[test]
+    fn rust_log_takes_precedence_over_override() {
+        assert_eq!(
+            resolve_filter_directive(Some("info,pluto_p2p=debug"), Some("warn")),
+            Some("info,pluto_p2p=debug".to_string())
+        );
+    }
+
+    #[test]
+    fn override_used_when_rust_log_absent_or_blank() {
```
varex83agent, line 167: nit: `resolve_filter_directive` tests cover blank/empty `rust_log` but not a blank *override* (e.g. `resolve_filter_directive(None, Some("   "))`), which currently returns `Some("   ")` rather than `None` per the asymmetry above. Adding that case would pin the intended contract and catch a future change to the override's empty-handling. The existing cases mirror the precedence rules clearly otherwise.

crates/tracing/src/init.rs
```diff
@@ -125,10 +126,61 @@ fn default_env_filter() -> EnvFilter {
...
+/// launched with. Takes the `RUST_LOG` value as an argument rather than reading
+/// the environment directly so the precedence logic is unit-testable.
+fn resolve_filter_directive(
+    rust_log: Option<&str>,
+    override_env_filter: Option<&str>,
+) -> Option<String> {
+    if let Some(rust_log) = rust_log
+        && !rust_log.trim().is_empty()
+    {
+        return Some(rust_log.to_string());
```
varex83agent, line 148: nit: both `Some(rust_log.to_string())` and `override_env_filter.map(str::to_string)` convert `&str -> String` via `ToString` (Display-based) rather than `ToOwned`. `to_owned()` / `str::to_owned` is marginally more idiomatic and is what clippy's `str_to_string` lint prefers. Purely stylistic; flag only if that lint is enabled.

test-infra/harness/simnet_test.go
```diff
@@ -0,0 +1,321 @@
...
+			t.Logf("broadcast from peer %d: %v", res.PeerIdx, res.Duty)
+		}
+	}
+}
+
+// assertCapturedAttestations polls gateway captures until, for some slot,
+// every node has submitted an attestation and all submitted payloads are
+// identical after JSON normalization (implementations may order keys
+// differently).
+func assertCapturedAttestations(t *testing.T, ctx context.Context, bnet *BeaconNet, n int) bool {
```
varex83agent, line 232: `assertCapturedAttestations` inspects only `pool/attestations` and asserts every node submitted an identical attestation for some slot. The in-process path (`assertBroadcasts`) is stronger — it requires each peer to complete all of `DutyPrepareAggregator`, `DutyAttester`, `DutyAggregator`. So for the headline cross-implementation scenario (`TestSimnetAttesterMixed`, gateway mode) the aggregation duties are effectively unverified: the gateway captures `aggregate_and_proofs` in `submissionPaths` but nothing asserts they arrived from every node. Consider also asserting on captured `aggregate_and_proofs` submissions per node to match the in-process coverage and back the README's 'both implementations participate in every duty' claim.

test-infra/harness/simnet_test.go
```diff
@@ -0,0 +1,321 @@
...
+		asserted = assertCapturedAttestations(t, ctx, bnet, n)
+		cancel()
+	} else {
+		asserted = assertBroadcasts(t, ctx, cancel, fixture, results, n)
+	}
+
+	// Surface node errors first: they are the root cause when assertions
+	// could not complete because a node exited early.
+	err := eg.Wait()
+	testutil.SkipIfBindErr(t, err)
```
varex83agent, line 148: `SkipIfBindErr(t, err)` is applied to `eg.Wait()`, which only carries errors from in-process charon nodes (`app.Run` on the errgroup). Pluto **subprocess** bind failures never flow into `eg` — they surface as a `WaitReady` timeout (`t.Fatalf`) or missing-submission assertion timeout, neither of which is skipped. So the address-in-use flakiness `SkipIfBindErr` is meant to absorb will hard-fail the pluto/mixed scenarios. Consider detecting bind failure from subprocess output/exit and skipping consistently.

test-infra/harness/simnet_test.go
```diff
@@ -0,0 +1,321 @@
...
+			}
+
+			if bySlot[slot] == nil {
+				bySlot[slot] = make(map[int]string)
+			}
+
+			bySlot[slot][sub.NodeIdx] = normalized
+		}
+
+		for slot, byNode := range bySlot {
```
varex83agent, line 256: nit: `assertCapturedAttestations` iterates `bySlot` (a map) and returns on the first fully-populated slot. Go randomizes map iteration order, so when multiple complete slots exist the asserted/logged slot varies run-to-run. Correctness is fine (any complete slot proves the invariant), but iterating slots in sorted order would make the chosen slot deterministic and failure logs reproducible across runs.

test-infra/harness/pluto.go
```diff
@@ -0,0 +1,187 @@
...
+	require.NoError(t, cmd.Start(), "start pluto relay")
+
+	t.Cleanup(func() {
+		_ = cmd.Process.Kill()
+		_ = cmd.Wait()
+	})
+
+	// Wait for the ENR HTTP endpoint to accept connections before returning.
+	deadline := time.Now().Add(30 * time.Second)
+	for time.Now().Before(deadline) && ctx.Err() == nil {
```
varex83agent, line 94: The readiness loop condition `time.Now().Before(deadline) && ctx.Err() == nil` falls through to `t.Fatalf("...not ready")` when the context is cancelled (a node exited early / teardown began), even though the relay itself may be healthy. The diagnostic is then misleading. Consider checking `ctx.Err()` explicitly and reporting cancellation distinctly from a genuine readiness timeout. (Same shape in `StartPlutoNode` around line 166.)

test-infra/harness/pluto.go
```diff
@@ -0,0 +1,187 @@
...
+
+	cmd := exec.CommandContext(ctx, bin, args...)
+	cmd.Dir = dir
+
+	stdout, err := cmd.StdoutPipe()
+	require.NoError(t, err)
+
+	cmd.Stderr = cmd.Stdout
+
+	go logLines(t, "pluto[relay]", stdout)
```
varex83agent, line 83: The `logLines` goroutine reads the subprocess stdout pipe and calls `t.Logf`. `t.Cleanup` does `Process.Kill()` + `Wait()`, which closes the pipe and lets the goroutine observe EOF and exit — so it doesn't leak. But there's a race between the goroutine's last `t.Logf` and the framework tearing down the test after Cleanup returns, which can trip `Log in goroutine after Test has completed` under `-race`. charon's own harness has the same pattern, so low-risk; a done-channel/WaitGroup that Cleanup waits on would make it deterministic. (Same at the node path, line 149.)

test-infra/harness/pluto.go
```diff
@@ -0,0 +1,187 @@
...
+//
+// `--p2p-advertise-private-addresses` is required: the relay binds loopback in
+// the harness, and without advertising private addresses the circuit
+// reservations it issues carry no address, which rust-libp2p relay clients
+// reject with `NoAddressesInReservation`.
+func StartPlutoRelay(t *testing.T, ctx context.Context, bin string) string {
+	t.Helper()
+
+	dir := t.TempDir()
+	httpAddr := testutil.AvailableAddr(t).String()
```
varex83agent, line 65: Port-allocation TOCTOU: `testutil.AvailableAddr` binds `localhost:0`, resolves the addr, and immediately closes the listener before the port is handed to the pluto subprocess. Between close and the subprocess's bind, another process/test can grab the port. A subprocess bind failure does **not** route through `SkipIfBindErr` — it crashes and surfaces as a `WaitReady` timeout, turning a flaky port race into a hard failure rather than a skip. Applies to the p2p/monitoring addrs here and in `StartPlutoNode` (lines 136-137). Consider retrying node startup on bind failure or passing the still-open listener fd.

test-infra/harness/gateway.go
```diff
@@ -0,0 +1,455 @@
...
+		writeJSON(w, map[string]any{
+			"data": map[string]string{
+				"disconnected":  "0",
+				"connecting":    "0",
+				"connected":     "8",
+				"disconnecting": "0",
+			},
+		})
+
+	default:
```
varex83agent, line 172: The `ServeHTTP` switch enumerates known duty/submission endpoints and falls through to the reverse proxy for everything else. A POST to an endpoint the underlying mock doesn't implement (e.g. a future block-proposal or sync-message submission) is silently proxied and returns whatever the mock returns (often 404/405), which a node may treat as a transient error and retry indefinitely rather than failing the test clearly. Acceptable for the curated attester-only flow today, but a `default` that logs unhandled POSTs would close a latent completeness gap.

test-infra/harness/gateway.go
```diff
@@ -0,0 +1,455 @@
...
+	case eth2spec.DataVersionCapella:
+		inner = att.Capella
+	case eth2spec.DataVersionDeneb:
+		inner = att.Deneb
+	case eth2spec.DataVersionElectra:
+		inner = att.Electra
+	case eth2spec.DataVersionFulu:
+		inner = att.Fulu
+	default:
+		inner = att.Electra
```
varex83agent, line 382: nit: the `default:` arm of `versionedAttestationJSON` sets `inner = att.Electra` and reports `att.Version.String()` as the version. For a genuinely unset/unknown `att.Version`, `att.Electra` may be nil, producing an inconsistent `{version, data}` pair (version says X, body is a nil Electra attestation) that the client could mis-decode. Prefer returning an error for unrecognized versions so a spec mismatch fails loudly in the test log.

test-infra/harness/charon.go
```diff
@@ -0,0 +1,104 @@
...
+	gatewayAddr string,
+	results chan<- SimResult,
+) {
+	t.Helper()
+
+	conf := app.Config{
+		Log:              log.DefaultConfig(),
+		Feature:          featureset.DefaultConfig(),
+		SimnetBMock:      mode == CharonInProcessBMock,
+		SimnetVMock:      true,
```
varex83agent, line 61: nit: `SimnetVMock: true` is set unconditionally for every in-process charon node, including the gateway/mixed scenarios. That's intentional (charon drives its own in-process VC while pluto nodes get the harness HTTP vmock in `vmock.go`), but it means charon and pluto nodes are exercised through *different* validator-client drivers in the same interop test — charon via its scheduler-wired validatormock, pluto via the harness HTTP vmock with a wall-clock ticker. The interop assertion therefore compares outputs of two different vmock drivers, not one shared driver. Worth a comment here or in the README so the assertion's scope isn't overstated.

.github/workflows/go-harness.yml
```diff
@@ -0,0 +1,88 @@
+name: Go simnet harness
+
+on:
+  push:
+    branches: ["main"]
+    paths:
```
varex83agent, line 6: nit: both jobs share this `paths` filter including `**/*.rs`, `**/Cargo.toml`, and `Cargo.lock`. The `charon-simnet` job builds no Rust and only exercises charon in-process, so it re-runs on every Rust-only change that can't affect it. Only `pluto-simnet` legitimately needs the Rust paths. Splitting the filters per job would avoid redundant runs; a unified filter is a reasonable simplicity tradeoff if you'd rather keep it.

varex83agent (commented):
## Summary

Adds `test-infra/harness/` — a Go simnet harness running end-to-end **attester-duty** tests over clusters mixing in-process charon nodes and subprocess pluto nodes (charon `v1.7.1` consumed as a plain module dependency) — plus four small, well-scoped production fixes: `#[serde(default)]` on `partial_deposit_data`, relay circuit-reservation address advertisement, `Node::remove_external_address`, and `RUST_LOG` > `--log-level` precedence.

Overall quality is high. The Rust fixes are correct and unit-tested where it matters; the `serde(default)` change is parity-safe (`partial_deposit_data` is part of the SSZ hash-tree-root covered by lock verification, so defaulting it to empty cannot bypass validation and matches charon's `omitempty` deserializer). The relay fix correctly gates auto-detected listen addresses behind `filter_private_addrs`.

### Bugs (must-fix)

None.

### Major

None.

### Minor / nit

A handful of minor findings, none blocking:

1. **Relay parity divergence** — the manually-configured `--p2p-external-ip` path adds addresses to libp2p without an is-public filter, whereas go-libp2p strips private addrs from circuit reservations. Low impact (only under misconfiguration, and it adds rather than removes reachable addrs), but worth a filter or a documented note.
2. **`RUST_LOG` precedence** intentionally inverts charon's flag-beats-env rule — defensible, but the doc comment should call out the divergence.
3. **Harness assertion coverage** — the gateway/mixed path only asserts attestation equality, not the aggregator duties the in-process baseline checks.
4. **Test-flakiness sources** — port-allocation TOCTOU + `SkipIfBindErr` not covering subprocess bind failures; non-deterministic slot selection; `t.Logf`-after-test race under `-race`.

Plus assorted nits. Leaving as `COMMENT` for author discretion — nothing here blocks merge.

🤖 Reviewed by the multi-agent review pipeline.

PR #540, opened 2026-07-10
feat(app): implement simnet integration

iamquang95:
Wires in-process simnet support into `pluto run` (`--simnet-beacon-mock` /
  `--simnet-validator-mock`), so a full all-pluto cluster runs the complete duty
  pipeline — schedule → QBFT consensus → threshold sign → broadcast — against mock
  beacon/validator clients, with no real beacon node or VC. Mirrors Charon's
  `app.Run` simnet wiring; this is what the compose smoke harness
  (`TestSmoke/all_pluto`) drives.

  ### Changes
  - Start the ported `BeaconMock` + `validatormock` inside `run`: point both beacon    clients at the in-process mock, load share keystores for the validator mock,    and drive it from the scheduler via a new abstract `SlotTickFn` seam (keeps    `wire.rs` decoupled from the mock type).
  - `--p2p-relays=""` now means "no relays" (Charon parity) — enables    isolated/offline runs.
  - Beacon-mock endpoint coverage for the whole flow: POST validators, block    produce/publish, sync-committee duties/contribution, and attestation /    sync-committee pool submits; the validator mock now sends wire-format    `SingleAttestation` and dials loopback (not the `0.0.0.0` bind address).
  - Log-level parity with Charon so the smoke alert gate can pass: quiet    simnet-only noise (libp2p relay churn, `force_direct`, weak-KDF keystores,    expired-duty QBFT drops).

  ### Testing
  - `cargo +nightly fmt --all --check` and    `clippy --workspace --all-targets --all-features -D warnings` clean; workspace    tests green.
  - New in-repo tests: single-node boot, validator resolution, produce-block,    attestation-shape, and the `--simnet-*` bridge.
  - End-to-end duty flow (4 nodes, threshold 3) exercised via the compose smoke    harness.


varex83:
Can we close it in favor of #553?

iamquang95:
Yup, let's close this, this is a POC

PR #541, opened 2026-07-10
test(app): add smoke test

iamquang95:
Add docker-compose smoke tests (charon parity + pluto/charon node mixing)

  Ports charon v1.7.1's testutil/compose smoke-test framework to test-infra/compose, adding a per-node implementation axis so clusters can mix charon and pluto nodes. Each scenario builds a real cluster  (keygen → run with simnet mocks → 60s under Prometheus alert gating); any firing alert fails the test.

  Scenarios cover charon's full matrix plus pluto variants (all_pluto, mixed_2_charon_2_pluto, pluto_dkg, gated behind -pluto-run). all_pluto passes against feat/simnet-run: 4 nodes reach QBFT consensus,  threshold-sign, and broadcast duties with zero errors/warnings.

  Deviations from charon (each commented):
  - Alert gate matches firing, not active — charon's never fires (upstream no-op bug)
  - Prometheus scrape config generated per cluster size (charon scrapes a non-existent bootnode)
  - Pluto keygen gets flags, not CHARON_* env vars (pluto CLI gap)
  - Path-less relay URL, dkg --shutdown-delay (distroless image), loki/otlp only when monitoring enabled, dropped CI-only sudo chown

  Run: PLUTO_REPO=<repo> go test ./smoke -v -integration [-pluto-run]


iamquang95:
@claude review this PR

iamquang95:
First, clone this branch and go to `compose` folder.
```
cd test-infra/compose
PLUTO_REPO=<your pluto repo want to run smoke test>  go test ./smoke -v -integration -timeout=35m
```

The 4 pluto scenarios vary two independent things — **who generates keys** vs **who runs the cluster** — plus DKG-vs-trusted-dealer:

  | Scenario | Keygen impl | Runtime impl | Unique coverage |
  |----------|-------------|--------------|-----------------|
  | `pluto_keygen_create` | **pluto** | charon | pluto artifacts → charon reads them |
  | `all_pluto` | pluto | **pluto** | pluto end-to-end, alone |
  | `mixed_2_charon_2_pluto` | charon | **charon + pluto** | live cross-impl duties + charon artifacts → pluto reads them |
  | `pluto_dkg` | **pluto DKG** | pluto | pluto's distributed keygen ceremony |


  ### What each one actually proves

  **`pluto_keygen_create` — artifact format, pluto→charon direction.** Pluto runs `create cluster` (writes `cluster-lock.json`, key shares, deposit data, ENR keys); then **charon** nodes load and run them.
  If pluto's lock format, share encoding, or lock signature deviated, charon rejects at startup or fails lock verification. Note the flip side: the runtime is *all charon*, so this exercises **zero pluto runtime code** — it's purely "can charon consume what pluto produced."

  **`all_pluto` — most pluto code, weakest proof.** Keygen, run, QBFT consensus, parsigex, sigagg, broadcast — all pluto, 4 nodes. It exercises the widest pluto surface, but because every peer is pluto, a  bug where **pluto disagrees with the spec but agrees with itself** passes silently (both sides compute the same wrong thing). Green here means "pluto is internally consistent and live," not "pluto is correct."

  **`mixed_2_charon_2_pluto` — the real interop test.** Threshold is 3-of-4, so **every** duty needs ≥3 of the 4 nodes to sign → both charon and pluto must participate in every duty for consensus to decide.
  This is the scenario that forced out every real interop bug:

  - signing-domain divergence (fork_schedule endpoint vs spec keys) — charon was rejecting 100% of pluto's partial signatures
  - mock head-root divergence (the Go `math/rand` port) — cross-impl sync-committee messages
  - QBFT reaching quorum across implementations, partial-sig exchange, aggregation across impls

  It also covers the **charon→pluto artifact direction** (charon does keygen here), so the two artifact directions together are covered by mixed + keygen_create. This is the only scenario that proves pluto and charon can actually run a validator *together*.

  **`pluto_dkg` — the distributed keygen ceremony.** The other three use `create cluster` (centralized/trusted-dealer keygen). This runs pluto's real DKG (frost) p2p ceremony across 4 nodes, then runs the cluster. Validates the ceremony end-to-end.

  ### Gaps worth knowing

  - **All simnet, mock beacon + mock VC.** No real beacon-node or real-VC path is gated (real VCs can't sync against the v1.7.1 mock).
  - **Liveness, not per-duty correctness.** "Green" = nodes broadcasting, no error/warn floods, cross-impl consensus decides — it won't catch a duty that produces a wrong-but-non-erroring value.
  - **Residual, currently masked in mixed:** infosync isn't wired into `pluto run` (#402B), so mixed excludes the `sched` warn topic; and there's a rare relay-circuit hang seen once in `all_pluto` that the deadline bound now surfaces as a `ConsensusTimeout` rather than hiding.

test-infra/compose/README.md
```diff
@@ -0,0 +1,154 @@
+# Pluto Compose
+
+> Run, test, and debug a developer-focussed insecure local pluto/charon cluster using docker compose
+
+This is a port of charon's `[testutil/compose](https://github.com/ObolNetwork/charon/tree/main/testutil/compose)`
```
varex83:
```suggestion
This is an extension of charon's `[testutil/compose](https://github.com/ObolNetwork/charon/tree/main/testutil/compose)`
```

varex83 (approved):
I approve, but I have one concert regarding maintainability, so the code is essentially copy of go implementation + support for pluto. So I see no reason of having this code in pluto repository, it would be much more reasonable to implement that part in Rust OR in charon repository. Wdyt? cc @emlautarom1

test-infra/compose/README.md
```diff
@@ -0,0 +1,154 @@
...
+Implementation names are validated when configs are written and loaded; anything
+other than `charon` or `pluto` is rejected.
+
+## Usage Examples
+
+Install the `compose` binary:
+
+```
+# From inside the pluto repo
+go install ./test-infra/compose/compose
```
emlautarom1: This does not quite work on my machine. I have to do the following (`cwd` is the root of the project):

```suggestion
go -C test-infra/compose install ./compose
```

test-infra/compose/README.md
```diff
@@ -0,0 +1,154 @@
...
+cd /tmp
+mkdir pluto-compose
+cd pluto-compose
+```
+
+Create the default cluster:
+
+```
+export PLUTO_REPO=/path/to/pluto
+compose clean && compose new --node-impls=pluto && compose define && compose lock && compose run
```
emlautarom1: `compose define` pulls the latest Charon docker image, not `v1.7.1`. Is this intended/correct?

emlautarom1 (commented):
First, @iamquang95 could you merge the latest changes in main into this branch? The `compose run` command fails because this branch is still missing the main workflow run: https://github.com/NethermindEth/pluto/blob/b42647e4849a7a2b32fb4b971a0ffce639fef3ba/crates/cli/src/commands/run.rs#L929

@varex83 As for your concern, I agree that it does not make sense to have this in Go: it introduces friction for developers (additional toolchain) and CI, with higher maintenance cost long term.

Nevertheless, since this code is for E2E testing we can keep it as is to not block progress and eventually we can port it. Also, since it's essentially the same Charon code we have high confidence that the tests are correct (if we were to port them we could introduce new bugs)

test-infra/compose/README.md
iamquang95: I removed the `compose` cli. This is not needed for smoke test

iamquang95:
Also, if we want, I can convert the smoke test to rust. Or we can delay the conversion later

test-infra/compose/define.go
```diff
@@ -0,0 +1,633 @@
...
+	}
+
+	return nil
+}
+
+// writePrometheusConfig writes prometheus scrape configs for the actual
+// cluster size, replacing the static 4-node default copied from static/.
+// Unlike charon's static config, this scrapes the relay (not a non-existent
+// "bootnode") and covers all NumNodes so the `up == 0` alert works.
+func writePrometheusConfig(dir string, conf Config) error {
```
emlautarom1, line 373: With this function, the `test-infra/compose/static/prometheus/prometheus.yml` file is stale (it's never used). Consider removing it.

test-infra/compose/README.md
```diff
@@ -0,0 +1,125 @@
...
+- Pluto nodes run `pluto:{tag}` (default `local`), built automatically from the repo
+root `Dockerfile` during the define step. This requires the `PLUTO_REPO` env var
+pointing at the pluto repo.
+- `KeyGenImpl` selects which implementation runs the single-container keygen steps
+(`create cluster` / `create dkg`); it defaults to node0's implementation.
+- The relay always runs the charon node-base image.
+
+Pluto accepts `CHARON_`* env vars and charon-compatible flags by design (CLI parity),
+so the generated docker-compose.yml services are identical for both implementations
+apart from the image. All node roles are supported for both implementations: keygen
```
emlautarom1: This is inaccurate: the `create` commands are currently missing `CHARON_*` env var support.

The test suite mentions this as a gap:
https://github.com/NethermindEth/pluto/blob/add04711d7ab97ea27f701aeb2c4bda974673087/test-infra/compose/config.go#L268-L271

I recommend adding support for them as part of this PR so we can drop this deviation.

test-infra/compose/auto.go
```diff
@@ -0,0 +1,303 @@
...
+
+	return nil
+}
+
+// execDown executes `docker compose down`.
+func execDown(ctx context.Context, dir string) error {
+	// Charon runs `sudo chown -R runner:docker` here to fix file ownership on
+	// its GitHub Actions runner. Dropped: `runner:docker` only exists in CI, so
+	// locally it just prompts for a sudo password and errors ("illegal group
+	// name"). Use -sudo-perms if a real permission fix is ever needed.
```
emlautarom1: We want to add a manually-triggered workflow for these tests, so we might want to keep the `chown` command.

test-infra/compose/lock.go
```diff
@@ -0,0 +1,188 @@
...
+		beaconMock = true
+		beaconNode = ""
+	}
+
+	lockFile := fmt.Sprintf("/compose/node%d/cluster-lock.json", index)
+
+	// The path-less URL form (multiaddrs response) instead of charon-compose's
+	// /enr path: pluto's relay parsing roundtrips URLs through a multiaddr,
+	// which cannot represent a URL path. Charon supports both forms.
+	//nolint:revive // tls not required for testing.
```
emlautarom1, line 111: This seems to be a feature gap in Pluto. Worth adding a ticket to address it so we can keep using the same test code as in Charon.

emlautarom1 (changes requested):
- Pluto should support `CHARON_` env vars instead of making a workaround. Since the change is small, let's include it in this PR.
- Add a Github workflow so we can run the tests manually (no automatic triggers for this due to how heavy they are).

crates/cli/src/cli.rs
```diff
@@ -186,3 +186,189 @@ pub enum CreateCommands {
...
+/// `[env: VAR]`. Charon does not expose env values in its help either (its
+/// viper env binding is invisible to cobra's help renderer), so this is also
+/// the parity behaviour.
+fn hide_env_values(cmd: clap::Command) -> clap::Command {
+    cmd.mut_args(|arg| arg.hide_env_values(true))
+        .mut_subcommands(hide_env_values)
+}
+
+#[cfg(test)]
+mod tests {
```
emlautarom1: This is too much, drop these tests and all associated ceremony.

emlautarom1 (approved):
Adding the attributes is enough, no need to add all of the extra test code around it.

PR #542, opened 2026-07-13
fix(peerinfo): serve inbound requests on Inactive connections

iamquang95:
## Summary

  peerinfo's `ConnectionHandler::poll` gated on connection state **first**, so an
  `Inactive` connection returned before handling inbound requests. `Inactive` only
  means the *remote* doesn't serve peerinfo (our outbound negotiation failed) — but
  the gate order meant we also stopped **answering that peer's inbound polls**.

  The charon relay is exactly this peer: it polls peerinfo as a client and never
  serves it, so our outbound to it fails → connection goes `Inactive` → we stop
  answering the relay.

  ## Change

  Reorder `poll` in `crates/peerinfo/src/handler.rs`: move the state gate **below**
  inbound handling. Inbound is served unconditionally; the gate now governs only
  outbound. Gate is kept (not removed) so `Inactive` still stops dialing and reports
  `Unsupported` once, avoiding a tight retry loop. One file: the reorder + a test.

  ## Impact if not fixed

  Every relay→node peerinfo poll times out: logs flood with `Peerinfo failed: read
  response` warnings (the smoke-test noise) and relay-side peer metadata
  (version/git-hash/RTT) never populates. Healthy node↔node exchanges unaffected.

### DIFF
--- a/crates/peerinfo/src/handler.rs
+++ b/crates/peerinfo/src/handler.rs
@@ -142,20 +142,11 @@ impl ConnectionHandler for Handler {
         cx: &mut Context<'_>,
     ) -> Poll<ConnectionHandlerEvent<ReadyUpgrade<StreamProtocol>, (), Result<Success, Failure>>>
     {
-        match self.state {
-            State::Inactive { reported: true } => {
-                return Poll::Pending; // Nothing to do on this connection
-            }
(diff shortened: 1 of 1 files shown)


iamquang95:
@claude review this PR

varex83 (approved):

PR #544, opened 2026-07-14
fix(scheduler): emit to subscribers asynchronously

### LINKED ISSUE #535: fix(core): scheduler subscribers permanently stop on broadcast Lagged
Part of #402 — hardening follow-up to the app-wiring PR. Pairs with the retry-wrapper issue.

## Scope

Scheduler duty/slot subscribers ride a bounded `tokio::sync::broadcast` channel (buffer 100) and **permanently terminate** on `RecvError::Lagged` (`crates/core/src/scheduler.rs:128-137,161-169`). A subscriber stuck for ~25 slots (e.g. hung beacon calls at 4 broadcasts/slot) silently kills all duty dispatch for that subscriber until the node restarts.

This is a Pluto-only failure mode: Charon dispatches per-duty goroutines and cannot lag. It is made reachable by the current serial subscriber processing (see the retry-wrapper issue, which restores async dispatch); fix both directions — a lagged subscriber should skip missed messages and continue with a warning (or subscribers should be made non-blocking per duty), never stop permanently.

## Dependencies

None (the retry-wrapper issue reduces the likelihood but does not remove the failure mode).

## Acceptance

- A lagged subscriber logs a warning and continues receiving subsequent duties/slots.
- Regression test: a slow subscriber that lags the channel resumes processing instead of terminating.

iamquang95:
Fix https://github.com/NethermindEth/pluto/issues/535

 ## Changes

  The slot and duty subscriber loops in `scheduler.rs` awaited each handler inline before receiving the next broadcast event. A slow or hung handler stalled the loop until it lagged, at which point the subscriber **terminated** (the loop breaks on `RecvError::Lagged`) — cascading into duty failures. Surfaced during simnet smoke debugging.

  This spawns the handler per event and loops straight back to `rx.recv()`, so a slow subscriber never blocks emission of later slots/duties.

  - `subscribe_slot` / `subscribe_duty`: spawn a detached task per event instead of awaiting inline.
  - `label` → `Arc<str>` so it clones cheaply into each task; duty error log gains `%duty`.
  - Regression tests: a later slot/duty is still delivered to a subscriber whose earlier handler hangs (fails on the inline-await version, passes after).

  ## Charon references

  Charon emits to subscribers asynchronously so a slow subscriber never blocks emission.

  - **Slot subscribers — goroutine per subscriber:** `core/scheduler/scheduler.go:157-167` (`go func(sub){ ... }(sub)`; matches the `"Emit scheduled slot event"` log).
  - **Duty trigger — goroutine per duty:** `core/scheduler/scheduler.go:240-265` (log at `:263`; matches the `"Trigger duty subscriber error"` log).

  Not a 1:1 port: Charon runs one goroutine per duty calling all subscribers sequentially, but each Charon subscriber is itself retry-wrapped fire-and-forget (`go retryer.DoAsync(...); return nil`, `core/retry.go`), so those calls return instantly. pluto has no such retry-wrapper, so it puts the detachment in the scheduler instead — one task per (subscriber, event). Different structure, same guarantee: a slow handler never blocks emission.

  ## Note

  Changes the concurrency contract — subscribers are now invoked concurrently, and tasks are detached fire-and-forget. Safe standalone in healthy operation (instances decide fast, tasks are short-lived),
  but **best landed with/after the consensus deadline-cancellation change**, which caps stuck-instance task lifetime.

### DIFF
--- a/crates/core/src/scheduler.rs
+++ b/crates/core/src/scheduler.rs
@@ -135,8 +135,17 @@ impl SchedulerBuilder {
                             }
                         });
                     }
-                    // NOTE: A lagging subscriber requires further analysis.
-                    // Log the error and terminate the subscription.
+                    // NOTE: Handlers are spawned per event above, so the
+                    // receive loop drains immediately. Lag therefore no longer
+                    // signals a slow handler — it can only fire if this loop
+                    // task itself is starved (runtime saturation). Log and
+                    // terminate the subscription.
+                    //
+                    // Trade-off of spawning: a permanently stuck handler now
+                    // leaks its detached task rather than eventually lagging and
+                    // terminating the subscription. Bounding that growth is a
+                    // follow-up, tracked with the consensus duty-deadline
+                    // cancellation that caps stuck-task lifetime.
                     Err(sync::broadcast::error::RecvError::Lagged(skipped)) => {
                         tracing::error!(
                             skipped,
--- a/crates/core/src/scheduler.rs
+++ b/crates/core/src/scheduler.rs
@@ -1,6 +1,6 @@
 use std::{
     collections::{HashMap, hash_map::Entry},
-    sync::{LazyLock, Mutex},
+    sync::{Arc, LazyLock, Mutex},
 };
 
 use backon::{BackoffBuilder, Retryable};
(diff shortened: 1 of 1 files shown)


iamquang95:
@claude review this PR

varex83 (approved):

PR #554, opened 2026-07-20
fix(p2p): recover relay boot races so cluster peers stay routed

iamquang95:
## Issue

  Pluto's relay manager (`crates/p2p/src/relay/manager.rs`) mishandles the circuit-reservation lifecycle at boot. Three bugs:

  1. **N listeners, 1 reservation.** One `ListenOn` was issued per relay transport address. libp2p's relay client holds a single reservation per relay connection and routes `ListenOn` by peer id, so each listener displaces the previous one — they race, sometimes settling with no live reservation. (`libp2p-relay-0.21.1/src/priv_client.rs`, `reservation_addresses`.)
  2. **Demote on any expiry.** `on_expired_listen_addr` demoted `Reserved → Established` on any circuit-addr expiry, including a never-confirmed sibling listener's, while another listener still held a live reservation.
  3. **Campaign dropped mid-flight.** `on_dial_failure` dropped a cluster-peer dial campaign on `DialPeerConditionFalse` assuming an existing connection — but that also fires while the campaign's own earlier dial is still negotiating. If that dial then failed, no `ConnectionClosed` ever fired (nothing established) and force-direct skips zero-connection peers, so the peer was never dialed again.

  ## Impact

  A node could sit partitioned for up to 60s after boot — demoted from `Reserved`, all peer routing dropped, cross-impl peers getting `NO_RESERVATION`. In the compose smoke test this surfaced as an intermittent alert-gate failure (~1/N runs) with no ERROR log. Pre-existing in `main`; unrelated to any single caller.

  ## Solution

  - Request exactly one circuit listener per relay (dialers build a target's circuit addrs from the relay's transport addrs regardless of which we advertise — same as Charon).
  - Track confirmed circuit addrs per relay (`reserved_addrs`); demote only when the last one expires.
  - On `DialPeerConditionFalse`, drop the campaign only if the peer store shows an active connection; otherwise keep it armed.
  - Add a zero-connection sweeper on the existing 15s watchdog tick to re-arm any known peer left with no connection and no campaign — the campaign-model equivalent of Charon's `NewRelayRouter` re-publish loop. Watchdog now runs before `process_relay_dials` in `poll` so a swept dial fires the same pass.

### DIFF
--- a/crates/p2p/src/relay/manager.rs
+++ b/crates/p2p/src/relay/manager.rs
@@ -1,7 +1,7 @@
 //! The [`RelayManager`] behaviour: reservation lifecycle and peer routing.
 
 use std::{
-    collections::{HashMap, VecDeque},
+    collections::{HashMap, HashSet, VecDeque},
     convert::Infallible,
     pin::Pin,
     task::{Context, Poll},
@@ -84,6 +84,18 @@ pub struct RelayManager {
     /// waiting for another `MutablePeer` update.
     relay_addrs: HashMap<PeerId, Vec<Multiaddr>>,
 
+    /// Circuit listen addresses the swarm has confirmed per relay.
+    ///
+    /// Each relay gets one `ListenOn` per transport address, and every
+    /// circuit listener confirms (`NewListenAddr`) or dies
+    /// (`ExpiredListenAddr`) independently — commonly a mix of both from the
+    /// same relay within milliseconds at boot, since concurrent listeners
+    /// race to reserve. A relay therefore only leaves `Reserved` when its
+    /// *last* confirmed address expires; reacting to any single expiry would
+    /// tear down peer routing while a sibling listener still holds a live
+    /// reservation.
+    reserved_addrs: HashMap<PeerId, HashSet<Multiaddr>>,
+
     /// Tracks when each relay last entered `Established` without having since
     /// reached `Reserved`. The watchdog uses this to identify relays whose
     /// reservation never confirmed (or whose refresh was denied so libp2p's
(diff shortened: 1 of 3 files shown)


iamquang95:
@claude review this PR

emlautarom1 (approved):
Can't really make a judgement here since this is extremely libp2p specific. Do you know more about this @varex83 ?

crates/p2p/src/relay/manager.rs
```diff
@@ -397,7 +451,21 @@ impl RelayManager {
...
+                // connection and routes `ListenOn` to the existing connection
+                // by peer id, so every additional listener displaces the
+                // previous listener's reservation: N listeners race each
+                // other at boot (confirm/expire churn) and can settle with no
+                // live reservation at all. One listener reserves once,
+                // cleanly. Dialers don't depend on which transport address is
+                // embedded in our advertised circuit address — both pluto and
+                // charon construct a target's circuit addresses from the
+                // relay's own transport addresses.
+                if let Some(circuit_addr) = Self::circuit_addrs(peer_id, &dial_state.addrs)
```
varex83, line 465: What if we want to have several protocols for relay connection (tcp/udp) or for example direct connection through IP vs DNS?

iamquang95: Multiple protocols do not require multiple ListenOn calls. TCP/QUIC and IP/DNS addresses are alternative dial paths. Once one relay connection succeeds, one reservation is created on it. The remaining addresses stay available for reconnect and failover.

varex83 (approved):

PR #555, opened 2026-07-21
cli: config --p2p-relay-loglevel into tracing filter

### LINKED ISSUE #483: relay: wire --p2p-relay-loglevel into the tracing filter (quiet benign libp2p relay handler warnings)
## Summary

The relay's `--p2p-relay-loglevel` / `PLUTO_P2P_RELAY_LOGLEVEL` flag is captured into config but **never applied** to the tracing subscriber. There is currently no way to control the verbosity of libp2p's own logs independently of Pluto's `--log-level`. This surfaces concretely as benign-but-noisy `WARN`s from the upstream relay handler that operators cannot quiet.

## Background / motivation

While investigating relay `WARN` logs we found this recurring entry:

```
WARN new_established_connection{...}: libp2p_relay::behaviour::handler: Dropping existing deny/accept future in favor of new one
```

Source: `libp2p-relay-0.21.1/src/behaviour/handler.rs:521,539`. It fires when a single connection sends overlapping reservation accept/deny requests; the handler keeps a single `reservation_request_future` slot and replaces the in-flight one. This is **benign** — the latest reservation future wins and the client retries/renews — but it is logged at `WARN` by the upstream crate and cannot be suppressed surgically without controlling the log filter.

## Root cause: dead config

- `crates/cli/src/commands/relay.rs:100` stores `p2p_relay_log_level` into `Config::libp2p_log_level`.
- `crates/relay-server/src/config.rs:44-46` holds the field.
- But the tracing filter is built **only** from `--log-level` via `build_console_tracing_config(self.log.level, ...)` (`crates/cli/src/commands/relay.rs:93-94`), which sets `override_env_filter` to that single level string (`crates/cli/src/commands/common.rs:49`), consumed in `crates/tracing/src/init.rs:47-51`.
- `libp2p_log_level` is therefore never read → dead config.

The flag mirrors Charon's `p2p-relay-loglevel`, so the intent was always to control libp2p verbosity.

## Suggested solution (preferred): wire the flag into the `EnvFilter`

When `--p2p-relay-loglevel` is set (non-empty), compose it into the env filter as per-target directives for the libp2p crates, e.g.:

```
<log-level>,libp2p=<p2p-relay-loglevel>
```

or target the relay crate specifically (`libp2p_relay=<level>`). The `EnvFilter` syntax already supports comma-separated per-target directives, so this can be threaded through `build_console_tracing_config` (e.g. add an optional `libp2p_level`/`extra_directives` parameter) or applied in the relay command before building the config. With the default empty value, behavior is unchanged.

This gives operators a real knob (parity with the flag's documented intent) and lets them quiet the libp2p relay handler noise (e.g. `--p2p-relay-loglevel=error`) without lowering Pluto's own log level.

## Workaround (if we don't wire the flag now)

Hardcode a scoped directive for just the noisy module in the relay command:

```rust
let log_config = build_console_tracing_config(
    format!("{},libp2p_relay::behaviour::handler=error", self.log.level),
    &self.log.color,
    loki_config,
);
```

Downside: blunt (hides all `WARN`-and-below from that module path) and not operator-configurable.

## Acceptance criteria

- `--p2p-relay-loglevel` / `PLUTO_P2P_RELAY_LOGLEVEL` measurably changes libp2p log verbosity at runtime.
- Empty/unset value preserves current behavior.
- The benign `Dropping existing deny/accept future...` warning can be suppressed via the flag.

## References

- `crates/cli/src/commands/relay.rs:93-94,100,145-151`
- `crates/relay-server/src/config.rs:44-46`
- `crates/cli/src/commands/common.rs:32-50`
- `crates/tracing/src/init.rs:46-51`
- Upstream log site: `libp2p-relay-0.21.1/src/behaviour/handler.rs:521,539`

PoulavBhowmick03:
Closes #483

### DIFF
--- a/crates/cli/src/commands/relay.rs
+++ b/crates/cli/src/commands/relay.rs
@@ -98,10 +98,8 @@ impl TryInto<pluto_relay_server::config::Config> for RelayArgs {
             }
         };
 
-        let p2p_relay = self.relay.p2p_relay_log_level;
-
         let log_config = build_console_tracing_config(
-            relay_filter(self.log.level.as_str(), &p2p_relay),
+            relay_filter(&self.log.level, &self.relay.p2p_relay_log_level),
             &self.log.color,
             loki_config,
         );
@@ -110,7 +108,7 @@ impl TryInto<pluto_relay_server::config::Config> for RelayArgs {
             .data_dir(self.data_dir.data_dir)
             .http_addr(self.relay.http_address)
             .auto_p2p_key(self.relay.auto_p2p_key)
-            .libp2p_log_level(p2p_relay)
+            .libp2p_log_level(self.relay.p2p_relay_log_level)
             .max_res_per_peer(self.relay.max_res_per_peer)
             .max_conns(self.relay.max_conns)
             // Invert p2p-advertise-private-addresses flag boolean:
--- a/crates/cli/src/commands/relay.rs
+++ b/crates/cli/src/commands/relay.rs
@@ -108,7 +108,6 @@ impl TryInto<pluto_relay_server::config::Config> for RelayArgs {
             .data_dir(self.data_dir.data_dir)
             .http_addr(self.relay.http_address)
             .auto_p2p_key(self.relay.auto_p2p_key)
-            .libp2p_log_level(self.relay.p2p_relay_log_level)
             .max_res_per_peer(self.relay.max_res_per_peer)
             .max_conns(self.relay.max_conns)
             // Invert p2p-advertise-private-addresses flag boolean:
--- a/crates/cli/src/commands/relay.rs
+++ b/crates/cli/src/commands/relay.rs
@@ -12,6 +12,14 @@ use tracing::{error, info};
 /// once `BackgroundTaskController::shutdown` has been signalled.
 const LOKI_FLUSH_TIMEOUT: Duration = Duration::from_secs(3);
 
+fn relay_filter(base: &str, relay_level: &str) -> String {
+    if relay_level.is_empty() {
+        base.to_string()
+    } else {
+        format!("{base},libp2p_relay={relay_level}")
+    }
+}
+
 /// Arguments for the relay command.
 #[derive(clap::Args, Clone)]
 pub struct RelayArgs {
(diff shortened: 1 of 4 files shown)


PoulavBhowmick03:
Hey @emlautarom1 , if there is anything in this PR that I need to change or work on, or address any other issues, do let me know

emlautarom1 (approved):
Verified that the wiring does suppress the warning from #483 properly when set to `error` while leaving our own logs at `info`.
Added the following:
- Dropped `Config::libp2p_log_level`: dead code.
- One-line doc on `relay_filter` recording the prefix-match contract.
- Replaced the tests with a check for correct behavior. Fails in more scenarios (in case of regressions).

For future work: invalid levels are swallowed silently, for example: `--p2p-relay-loglevel=fatal` resets the filter to `info` and discards `--log-level`, and a bad `--log-level` like `warning` silences all output. Pre-existing issue and applies to `run`/`dkg` too, so better fixed across all level flags in a follow-up PR.

varex83 (approved):

crates/cli/src/commands/relay.rs
```diff
@@ -11,6 +11,16 @@ use tracing::{error, info};
...
 const LOKI_FLUSH_TIMEOUT: Duration = Duration::from_secs(3);
 
+/// Adds a `libp2p_relay` directive to the `base` env filter, which `EnvFilter`
+/// prefix-matches against every `libp2p_relay::*` target.
+fn relay_filter(base: &str, relay_level: &str) -> String {
+    if relay_level.is_empty() {
+        base.to_string()
+    } else {
+        format!("{base},libp2p_relay={relay_level}")
+    }
```
iamquang95: Nit: Could we validate and normalize `relay_level` before interpolating it here?

Charon fails startup instead of continuing with a different filter: https://github.com/ObolNetwork/charon/blob/v1.7.1/cmd/relay/p2p.go#L35-L42.

iamquang95 (approved):

crates/cli/src/commands/relay.rs
emlautarom1: Changed to a closed enum with a hard failure on invalid filters. #671 expands on this idea.

PR #567, opened 2026-07-27
fix(core): cache beacon config for signing-domain resolution

### LINKED ISSUE #563: bug(core): signing-domain resolution performs live, uncached BN config fetches on every duty-critical operation
## Summary

Every signature verification resolves the eth2 signing domain by fetching BN config endpoints
**at call time, with no caching**: an attester partial-sig verification costs **4 sequential HTTP
calls** (spec ×2 + genesis + fork_schedule — `crates/eth2util/src/signing.rs:105-117`,
`crates/eth2api/src/extensions.rs:287,352,411`); the aggregator fetch path adds another spec lookup.

On a non-local BN link this eats the 4s duty budget, and a single stalled request fails the
operation. Worse, the failure is **misreported as a signature failure** — the VC gets HTTP 400
`partial signature verification failed` when no signature was ever checked.

Charon has the same call shape but its go-eth2-client fork caches
[`Spec`](https://github.com/ObolNetwork/go-eth2-client/blob/v0.27.1-obol.1/http/spec.go#L43) /
`Genesis` /
[`ForkSchedule`](https://github.com/ObolNetwork/go-eth2-client/blob/v0.27.1-obol.1/http/forkschedule.go#L39)
— on identical infra (BN behind a ~670ms tunnel), charon ran a clean 14-epoch hoodi window while
pluto missed 3/14.

## Evidence (hoodi, 2026-07-22/23, 2×pluto+2×charon, validator 1454246)

Transport failures confirmed live in three verification paths (sigagg 4×, parsigex external store
2×, vapi 400 1×); likely contributed to ≥1 attester miss and to aggregator starvation (2/22
aggregator duties succeeded; direct log: `fetch aggregator data → GetSpec(… operation timed out)`).

```
ERROR … pluto_core::sigagg: verify failed error=aggregate signature verification failed:
  Request error: error sending request for url (…/eth/v1/beacon/genesis)
lodestar VC: submitPoolAttestationsV2 failed with status 400: partial signature verification failed
```

iamquang95:
Fixes #563.

  Signing-domain resolution hit the BN live on every verification — 4 HTTP calls per attester partial-sig check — and a BN failure mid-verification was misreported to the VC as 400 "partial signature verification failed". Charon avoids both via caching in its go-eth2-client fork.

  - `BeaconNodeClient` (now a single `Arc<Inner>`) embeds a config cache for spec/genesis/fork_schedule: 5-min TTL, concurrent cold lookups coalesce into one fetch, errors and empty/malformed responses never cached.
  - Both scheduling and submission clients are warmed at startup, before duty scheduling; warm failure aborts startup.
  - All consumers read config through the cached client; raw `fetch_domain*` methods are deleted, leaving one domain derivation.
  - `SigningError` → HTTP by provenance: BN/config failure 502, invalid submitted signature 400, invalid trusted key material (e.g. bad pubshare) 500.


iamquang95:
@claude review this PR

emlautarom1 (commented):
Couple of notes:
- The endpoints suggested to be cached with a TTL never change in the most popular Beacon clients (Lighthouse, Prysm and Teku, representing ~90%). We might as well just cache for the lifetime of the application and avoid the TTL complexity.
- We should explore dropping the `eth2api` crate as is: instead of relying on code generation, we can use agents to have a tailor made crate that exposes the endpoints that we need precisely, with full control over the underlying HTTP client. This is a lot more work but simplifies the build process, total LOC, improves error reporting, etc.

With this, I'm not sure we want to go this route right now unless it's a hard requirement for the MVP (ie. testnet shows that without it Pluto misbehaves)

emlautarom1:
Since the fix is required (both Hoodi testnet and Kurtosis setups are showing problems), let's explore a more minimal set of changes, specially since - as mentioned - we'll most likely redesign the eth2api after the MVP release:

- Cache the three leaf fetches only: `fetch_spec_data`, `fetch_genesis_data`, `fetch_fork_schedule_data` in `crates/eth2api/src/extensions.rs`. Everything that uses it benefits automatically with no changes.
- Lifetime cache, no TTL: matches Charon's go-eth2-client and in fact these never change in Lighthouse/Prysm/Teku (~90% of beacon clients).
- Process-global keyed by `base_url` (`static LazyLock<RwLock<HashMap<Url, ...>>>`, like `eth2util/src/network.rs`). client.rs is codegen so we can't add a field, and a wrapper type results in massive changes.

Let's keep the original issue open until we have a good redesign for current codegen.

iamquang95:
@emlautarom1 I updated the PR with the minimal changes on `extension.rs` which uses a global cache for chain config.

varex83 (approved):

emlautarom1 (approved):
LGTM, with this the changes are a lot smaller. We'll revisit the whole `eth2api` crate so no need to have anything beyond the basic feature.

Issue #402, opened 2026-05-12
Implement `app/app.go` equivalent — node graph wiring & lifecycle


varex83:
## Scope

Port `charon/app/app.go` (~1256 LOC) — the **integration anchor** that instantiates every duty-pipeline component in the correct order with lifecycle hooks. This is the final assembly step that makes Pluto a runnable node.

## Dependencies (must land first)

This issue cannot start until every component below is implemented:

**Track A:**
- #393 `core/qbft` async
- `core/consensus` framework (controller/wrapper/debugger) — see Track A epic for issue number
- #157 `core/consensus/qbft`, #158 timer, #48 instance, #62 metrics, #39 mocks
- #176 `core/scheduler`, #171 `core/dutydb`, #172 `core/fetcher`, #170 `core/bcast`

**Track B:**
- #180 `core/validatorapi`, #179 `core/tracker`, #178 `core/sigagg`, #169 `core/aggsigdb`

**Track C:**
- #68 `app/health`, #183 `app/sse`, `app/monitoringapi`
- `core/gater`, #177 `core/priority`, #173 `core/infosync`

## Acceptance

- All components instantiated with proper lifecycle ordering (start/stop semantics match Go)
- Channel/event wiring matches the Go graph
- Unit-level integration test boots the graph against `beaconmock` (#37)
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean

emlautarom1:
Pending work will be tracked in #610

PR #573, opened 2026-07-30
feat(eth2api): add Charon-parity beacon node request metrics

varex83:
## Summary

Adds Charon `eth2wrap`-parity Prometheus metrics for beacon node requests, using Pluto's `vise` conventions (not the `prometheus` crate, which Pluto doesn't use):

- `app_eth2_requests_total{endpoint}`
- `app_eth2_errors_total{endpoint}`
- `app_eth2_latency_seconds{endpoint}` — with Charon's exact 17 latency buckets

Metric names match Charon's `app_eth2_*` series. They're defined in a new `crates/eth2api/src/metrics.rs` and auto-registered via `#[vise::register]`, so the existing `/metrics` endpoint surfaces them with no extra wiring.

A small `instrument(endpoint, fut)` helper reproduces Charon's per-call semantics — record a request + observe latency always, increment errors on `Err` — and slots in ahead of the existing `.map_err(...)` handling at each call site (the generated client methods return `anyhow::Result`, so caller types are preserved).

## Coverage

Instruments the endpoints Pluto actually calls today, using Charon's snake_case labels:

- **Inside existing `eth2api` wrappers (zero call-site churn):** `attester_duties`, all `submit_*` duties, `spec`, `genesis`, `fork_schedule`, `validators`.
- **Inline at `core`/`app` call sites:** `proposal`, `attestation_data`, `sync_committee_contribution`, `aggregate_attestation`, `proposer_duties`, `sync_committee_duties`, `node_syncing`, `node_version`, `node_peer_count`.

Endpoints Charon instruments but Pluto doesn't call yet (e.g. `beacon_block_root`, the `*_selections`, `prepare_sync_committee_subnets`) are intentionally skipped — each is now a one-line `instrument(...)` addition when the port reaches them.

## Deliberate deviations from Charon

1. **`using_fallback` gauge omitted** — Pluto has no multi/fallback beacon client yet (#402 part B). To be added when fallback lands.
2. **`errors_total` counts transport/decode `Err` only, not HTTP 4xx/5xx.** Charon's go-eth2-client returns an error on non-2xx; Pluto's generated client returns `Ok(Response::BadRequest(..))` for 4xx, which wrappers convert to `Err` *after* the instrumented call. Full error-count parity would require instrumenting around response classification — can follow up if exact parity is wanted.

## Testing

- `cargo +nightly fmt --all --check` ✅
- `cargo clippy -p pluto-eth2api -p pluto-core -p pluto-app --all-targets --all-features` ✅ (clean)
- `cargo test`: eth2api (119) ✅, core (600) ✅, app ✅ — includes new unit tests asserting the exposition text and all latency buckets.

### DIFF
--- a/crates/core/src/scheduler.rs
+++ b/crates/core/src/scheduler.rs
@@ -812,9 +812,12 @@ async fn wait_chain_start(client: &pluto_eth2api::BeaconNodeClient) -> Result<()
 /// Blocks until the beacon node is synced.
 async fn wait_beacon_sync(client: &pluto_eth2api::BeaconNodeClient) -> Result<()> {
     let fetch = || {
-        client
-            .api()
-            .get_syncing_status(pluto_eth2api::GetSyncingStatusRequest {})
+        pluto_eth2api::instrument(
+            "node_syncing",
+            client
+                .api()
+                .get_syncing_status(pluto_eth2api::GetSyncingStatusRequest {}),
+        )
     };
     let fetch_backoff = crate::expbackoff::fast();
 
@@ -963,9 +966,7 @@ async fn fetch_proposer_duties(
         .epoch(slot.epoch().to_string())
         .build()
         .map_err(pluto_eth2api::EthBeaconNodeApiClientError::RequestError)?;
-    let resp = client
-        .api()
-        .get_proposer_duties(req)
+    let resp = pluto_eth2api::instrument("proposer_duties", client.api().get_proposer_duties(req))
         .await
         .map_err(pluto_eth2api::EthBeaconNodeApiClientError::RequestError)?;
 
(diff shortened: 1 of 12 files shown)


crates/eth2api/src/metrics.rs
```diff
@@ -0,0 +1,142 @@
...
+    F: Future<Output = Result<T, E>>,
+{
+    let metrics = &*ETH2_METRICS;
+    metrics.requests_total[endpoint].inc();
+
+    let start = Instant::now();
+    let result = fut.await;
+    metrics.latency_seconds[endpoint].observe(start.elapsed().as_secs_f64());
+
+    if result.is_err() {
```
iamquang95: is_err() never fires for BN HTTP errors

crates/core/src/scheduler.rs
iamquang95: Missing instrument("attester_duties", ...)

crates/core/src/validatorapi/component.rs
iamquang95: Missing instrument("attester_duties", ...)

iamquang95 (approved):

PR #574, opened 2026-07-30
feat(app): wire priority + infosync into the node

varex83agent:
## What

Wires the `priority` protocol behaviour and `infosync` component into the running node, and triggers a cluster-wide priority exchange on the last slot of each epoch.

Previously both crates were implemented and tested but never mounted (`crates/app/src/node/mod.rs` carried a `TODO(#402 part B)`), so a pluto node neither advertised `charon/priority/2.0.0` nor proposed into the info_sync QBFT round. In a mixed cluster this made Charon peers log `protocols not supported: [charon/priority/2.0.0]` and `Priority protocol consensus: consensus timeout` every epoch.

## Scope — participation only ("A")

This is the participation half of #402 part B:

- `CoreBehaviour` gains a `priority` field, constructed via `pluto_priority::new_component` on the **shared** `p2p_context` and riding the **existing** `Arc<qbft::Consensus>` (priority uses raw QBFT, matching Charon's `wirePrioritise`).
- `infosync::Component` is built with `version::SUPPORTED`, the concatenated `protocols()` of consensus/parsigex/peerinfo/priority, and `[Builder?]+Full` proposal types.
- A per-epoch trigger is registered as a scheduler slot subscriber, guarded by `Slot::last_in_epoch()` (6s exchange timeout, `min_required` = cluster threshold — both matching Charon).

**Deferred (`TODO(#402 part B)`):** consuming the decided result — routing the duty path through `ConsensusController` and registering the protocol-switch subscriber (`set_current_consensus_for_protocol`). That is a functional no-op while QBFTv2 is the only consensus protocol, so it is intentionally left out. This PR does **not** wire `proposals()` into the fetcher, because Charon does not either (its `Proposals()`/`Protocols()` accessors are dead code; full-vs-blinded is decided by `--builder-api`).

## Verification

- `cargo +nightly fmt --all --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo test --workspace --all-features` all pass.
- End-to-end on a 2-charon + 2-pluto kurtosis cluster: at the epoch boundary all four nodes reach **byte-identical** info_sync results (e.g. `protocol: qbft=4000, parsigex=3996, peerinfo=3992, priority=3988`), proving pluto's `calculateResult` is deterministic-equivalent to Charon's. Charon's `protocols not supported` and `consensus timeout` warnings stop; regular duties are unaffected.


crates/app/src/node/mod.rs
```diff
@@ -378,12 +382,21 @@ async fn run(config: AppConfig, ct: CancellationToken) -> Result<(), AppError> {
...
             .await
             .map_err(AppError::Deadline)?,
     );
+    // Priority needs an owned `DutyDeadlineCalculator` (it is not `Clone`), so
+    // build a second instance from the same client. The calculator is stateless
+    // beacon-derived config, so this is identical to the one above.
+    let priority_deadline_calc =
+        pluto_core::deadline::DutyDeadlineCalculator::from_client(&eth2_cl)
+            .await
+            .map_err(AppError::Deadline)?;
```
emlautarom1: You can clone the existing dealine calculator and not create a new one:


```suggestion
    let priority_deadline_calc = Arc::clone(&deadline_calc);
```

crates/app/src/node/wire.rs
```diff
@@ -608,6 +613,28 @@ pub async fn wire_core_workflow(
...
+                let ct = ct.clone();
+                let slot = slot.clone();
+                async move {
+                    if slot.last_in_epoch()
+                        && let Err(err) = infosync.trigger(ct.child_token(), slot.slot).await
+                    {
+                        tracing::warn!(%err, slot = ?slot.slot, "infosync trigger failed");
+                    }
+                    Ok::<(), AppError>(())
+                }
```
emlautarom1, line 629: The `subscribe_slot` handles errors by logging already, so we can propagate errors:


```suggestion
                async move {
                    if slot.last_in_epoch() {
                        infosync.trigger(ct.child_token(), slot.slot).await?;
                    }
                    Ok::<(), AppError>(())
                }
```

crates/app/src/node/behaviour.rs
```diff
@@ -65,6 +67,14 @@ pub struct CoreHandles {
...
+    /// to the per-epoch infosync trigger.
+    pub priority: Arc<pluto_priority::Component>,
+    /// Expired-duty receiver paired with `priority`, drained by
+    /// `pluto_priority::Component::start`. Taken (moved) once at start.
+    pub priority_expired_rx: tokio::sync::mpsc::Receiver<pluto_core::types::Duty>,
+    /// Infosync component driving the per-epoch cluster-wide priority exchange.
+    pub infosync: Arc<pluto_infosync::Component>,
 }
 
 /// Composes the core behaviours and builds the libp2p [`Node`].
```
emlautarom1, line 97: Probably better to revisit this with the new set of arguments.

emlautarom1 (approved):
LGTM, some potential simplifications. See if we can reduce the overly-verbose comments and inline variables/avoid clones. Also, consider if we can add some tests to verify the wiring.

crates/app/src/node/mod.rs
varex83agent: Done in 669fe0f — dropped the second `DutyDeadlineCalculator` and now pass `Arc::clone(&deadline_calc)` (the shared `Arc<dyn DeadlineCalculator>`); `wire_p2p` takes the `Arc` and `new_component` accepts it via the blanket `impl DeadlineCalculator for Arc<T>`.

crates/app/src/node/wire.rs
varex83agent: Done in 669fe0f — the trigger error now propagates via `?` and `subscribe_slot` logs it; dropped the manual `warn!`.

crates/app/src/node/behaviour.rs
varex83agent: Done in 669fe0f — replaced the 14 positional args (and the `too_many_arguments` allow) with a `WireP2PParams` struct destructured at the top of `wire_p2p`.

PR #575, opened 2026-07-31
refactor(app): route the duty path through ConsensusController

iamquang95:
## Problem

  Charon constructs QBFT through ConsensusController, but Pluto wired QBFT directly.

  ## Fix

  Match Charon by constructing consensus through ConsensusController, wiring its default QBFT instance to P2P and its swappable current-consensus wrapper to core.


emlautarom1:
@claude

emlautarom1 (approved):
LGTM, the wrapper adds some indirection that makes it harder to understand errors.

I'd also be OK with removing the wrapper entirely now that Pedersen is out of scope: the code should be a lot shorter with no indirections.

iamquang95:
For the wrapper, I will keep the wrapper since the Pedersen is planned as a second protocol.

PR #576, opened 2026-07-31
fix(p2p): populate relay_connections, gate ping metrics to cluster peers

iamquang95:
Two `p2p` metric fixes for the Grafana dashboard:

  - **Extra entry on the peer panels**: `p2p_ping_latency_secs` / `ping_success` / `ping_error_total` were emitted for every pinged peer, so the relay server appeared as a phantom cluster peer on `max(p2p_ping_success) by (peer)`. All three are now gated to known cluster peers (Charon's `ClusterPeerIDs` allowlist).
  - **Blank "Connected Relays" panel**: `p2p_relay_connections` had no writer. It is now set from `RelayManager` — `1` while a reservation is held, `0` otherwise.

### DIFF
--- a/crates/core/src/tracker/metrics.rs
+++ b/crates/core/src/tracker/metrics.rs
@@ -15,6 +15,19 @@ pub struct TrackerMetrics {
     #[metrics(labels = ["duty", "peer"])]
     pub participation_success_total: LabeledFamily<(String, String), Counter, 2>,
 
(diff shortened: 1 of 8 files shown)


emlautarom1:
@claude

emlautarom1 (approved):
I would look to reduce the ammount of comments: they introduce more confusion than clarifications. If the metrics are used then keep them (no need to mention that they're deprecated).

PR #581, opened 2026-08-03
fix(cluster): accept charon's null/omitempty cluster-file shapes

varex83agent:
## Problem

pluto's versioned cluster-definition and distributed-validator structs are missing serde defaults for several fields that charon legitimately emits as JSON `null` or omits entirely. As a result, `serde` rejects cluster files that charon writes.

Concretely, charon (v1.7.1 defaults to writing cluster-file **v1.10**, and also reads **v1.8/v1.9**) can produce:

- `"operators": null` and `"validators": null` — Go marshals a nil slice as `null`, and these fields have **no** `omitempty`, so the key is always present but may be `null`.
- an absent `timestamp` and absent `name` — both `omitempty`, so the key is dropped when empty.
- an absent `public_shares` and absent `partial_deposit_data` on distributed validators — both `omitempty`.

`#[serde(default)]` alone handles *absent* keys but **not** an explicit `null`, so the null-marshaled slices additionally need `DefaultOnNull`.

## Fix

- **`operators`, `validators`** (definition v1.8 / v1.9 / v1.10): `#[serde(default)]` + `#[serde_as(as = "DefaultOnNull")]` — accepts both `null` and an absent key, deserializing to an empty vec. Mirrors the existing `deposit_amounts` handling.
- **`timestamp`, `name`** (v1.8 / v1.9 / v1.10): `#[serde(default)]`. (v1.10 already had it for `name`; v1.8/v1.9 did not.)
- **`public_shares`, `partial_deposit_data`** (`DistValidatorV1x8orLater`): `#[serde(default)]`.

All three definition versions are fixed because charon reads v1.8/v1.9/v1.10 (`cluster/version.go`: `supportedVersions`).

## Notes

- The existing `Definition` deserializer enforces `num_validators == validators.len()`, so `validators: null` is only valid alongside `num_validators: 0` — which is exactly what charon emits. The test reflects this.
- Serialization is unchanged: `DefaultOnNull`'s `SerializeAs` forwards to the normal serializer, so non-empty slices still serialize as arrays.

## Tests

- `definition_accepts_null_slices_and_absent_omitempty_fields` drives the v1.8/v1.9/v1.10 fixtures with `operators`/`validators` set to `null` and `name`/`timestamp` removed, asserting they parse to empty/default.
- `dist_validator_v1x8_accepts_absent_omitempty_fields` strips `public_shares` and `partial_deposit_data` from a real lock fixture's validator and asserts it parses.

`cargo test -p pluto-cluster` (98 tests), clippy, and fmt all pass.


emlautarom1 (changes requested):
The underlying problem is not fully solved:

Checked the corpus at `crates/cluster/src/examples/` using  this branch:

```
                              main             this PR
cluster-definition-000..006   PARSES (all 7)   PARSES (all 7)
cluster-lock-000              FAILS -> invalid type: null, expected a string
cluster-lock-001              FAILS -> invalid type: null, expected a string
cluster-lock-002              FAILS -> missing field `name`
cluster-lock-003              FAILS -> invalid value: string "", expected hex bytes convertible to target type
```

All seven definitions passed before this PR too, so the corpus result is unchanged end to end. Charon's own `TestExamples` round-trips and signature-verifies every one of these files.

The default-written v1.10 shape also still fails, with this branch applied:

```
baseline (unmodified)                            PARSES
distributed_validators: null                     FAILS -> invalid type: null, expected a sequence
builder_registration.message.fee_recipient = ""  FAILS -> invalid value: string ""...
builder_registration.message.pubkey = ""         FAILS -> ...
builder_registration.signature = ""              FAILS -> ...
deposit_data[0].pubkey = ""                      FAILS -> ...
```

This is due to the following:

1. **`lock.rs` isn't touched.** [`LockV1x8orLater.distributed_validators`](https://github.com/NethermindEth/pluto/blob/98d606581e542837fd794bb06f207765d03e6c4f/crates/cluster/src/lock.rs#L675) has no default, and Go marshals the nil slice as `null` with no `omitempty`, so charon's own empty-cluster lock is rejected. Same fix as the rest of this PR.

2. **A second root cause the PR doesn't cover:** `[u8; N]` hex fields reject `""` while `Vec<u8>` hex fields accept it, and charon's [`to0xHex(nil)`](https://github.com/ObolNetwork/charon/blob/v1.7.1/cluster/helpers.go#L254) returns exactly `""`. That accounts for `cluster-lock-003` and all four builder-registration/deposit failures above.

3. **The pre-v1.8 structs are in scope in practice.** The description says charon reads v1.8/v1.9/v1.10, but [`supportedVersions`](https://github.com/ObolNetwork/charon/blob/v1.7.1/cluster/version.go#L31) lists `v1_0` through `v1_10`. `cluster-lock-000/001` (v1.1.0, `config_signature: null`) and `-002` (v1.2.0, absent `name`) are real files that fail today.

I suggest to mirror Charon's `TestExamples` criteria: parse every file in `crates/cluster/src/examples/` raw, **locks included** (only definitions are processed today).

Feel free to fold the changes onto this PR or create follow ups.

varex83agent:
Thanks @emlautarom1 — addressed all four points; the whole `examples/` corpus (locks included) now parses.

1. **`lock.rs` untouched → fixed.** `distributed_validators` on every `Lock*` version (v1x0or1 … v1x8orLater) now has `#[serde(default)]` + `DefaultOnNull`, so charon's nil-marshaled `null` slice (empty-cluster lock) parses.
2. **`[u8; N]` hex fields reject `""` → fixed.** `HexBytes` now deserializes `""` to a zero-filled `[u8; N]`, matching charon's `to0xHex(nil) == ""` (an SSZ `Bytes48`/`Bytes32`/… of a nil slice is exactly `N` zero bytes). The blanket `TryFrom<Vec<u8>>` impl is split into concrete `Vec<u8>` / `[u8; N]` impls; `Vec<u8>` still decodes `""` to an empty vec. This covers `cluster-lock-003` and the builder-registration/deposit-data empty-hex cases.
3. **Pre-v1.8 in scope → fixed.** `name`/`timestamp`/`fee_recipient_address`/`withdrawal_address` (charon `omitempty`) get `#[serde(default)]`, `operators`/`validators` get `DefaultOnNull`, and `OperatorV1x1`/`OperatorV1x2orLater` `config_signature`/`enr_signature` accept `null`. That fixes `cluster-lock-000/001` (v1.1.0, `config_signature: null`) and `-002` (v1.2.0, absent `name`).
4. **Mirrored `TestExamples`.** New `parses_every_example_file` test reads every file in `crates/cluster/src/examples/` and parses definitions as `Definition` and **locks as `Lock`**. Also dropped the manual empty-hex rewrite in `parse_example_lock` now that the parser handles `""` natively.

`cargo test -p pluto-cluster` (all example files parse), clippy, and fmt pass.

crates/cluster/src/lock.rs
```diff
@@ -728,44 +734,54 @@ mod tests {
...
-                    }
-                }
-            }
-        }
-
-        serde_json::from_value(value).unwrap()
+        // Empty hex strings (charon's `to0xHex(nil)`) for fixed-size deposit
+        // fields are handled by the `HexBytes` deserializer, so parse directly.
+        serde_json::from_str(json).unwrap()
     }
```
emlautarom1: nit: inline this function

emlautarom1 (approved):
Before merging, ensure:
- Merge the latest changes on main
- On the `parses_every_example_file`, after each parse run `verify_hashes` and `verify_signatures` to match the Charon test.

varex83agent:
Addressed the pre-merge points:

1. **Merged latest `main`** into the branch.
2. **`parses_every_example_file` now mirrors charon's `TestExamples`** — after parsing each example file (definitions and locks), it runs `verify_hashes()` and `verify_signatures()`. The signature check uses a noop `EthClient` (`EthClient::new("")`), the equivalent of charon's `VerifySignatures(nil)` — none of the fixtures carry ERC-1271 smart-contract signatures.
3. Inlined the now-trivial `parse_example_lock` helper (nit).

`cargo test -p pluto-cluster` (104 tests), clippy, and fmt all pass.

PR #537, opened 2026-07-08
fix(docker): improve Dockerfile for cache efficiency and image size

Patrick-Ehimen:
Closes #351

## Changes
- Add `cargo-chef` planner/builder stages to cache dependency compilation separately from source
- Pin base images by digest (`rust:slim-bookworm`, `distroless/cc-debian13`)
- Use `rust-toolchain.toml` as single source of truth for the Rust toolchain version
- Pin `protobuf-compiler` to exact version and clean apt lists to reduce image size
- Add `--no-install-recommends` to avoid pulling unnecessary packages

PR #582, opened 2026-08-03
feat: align dockerfile with charons

OisinKyne:
I tried to swap directly to pluto with a charon docker-compose setup and ran into an issue that pluto doesn't default to `run`, fable also added some other adjustments too.


emlautarom1 (commented):
Thanks for your contribution!
We have folded the changes into #537, specifically
- `CMD ["run"]` added so Pluto defaults to the `run` command.
- `LABEL` metadata is already handled by our CI, so it's not required in the Dockerfile
- Permissions, users and mounts require a further work to avoid breaking existing infrastructure (CI, smoke tests) so we'll work on this on a separate PR.

PR #591, opened 2026-08-06
fix(relay): remove the bind race in the relay test fixture

### LINKED ISSUE #590: test(relay): Flaky test: `commands::relay::tests::serve_addr_enr_ext_ip` — bind race in the relay test fixture
## Symptom

  Intermittent CI failure: `unwrap()` on `reqwest::Error { Connect, ConnectionRefused }` for `http://127.0.0.1:<port>/enr`. The relay never came up, but the panic points at the HTTP request, so the real bind error only shows up in a log line (`Failed to bind HTTP listener: 127.0.0.1:<port>`).

  ## Cause

  `with_relay_server` (`crates/cli/src/commands/relay.rs:544` on `main`) allocates ports by **bind → read port → drop → let the relay rebind** (`relay.rs:555-574`). Under concurrency another process can claim the port in that window. The relay task then fails its bind and **exits permanently**, while the fixture goes straight to the request — so `retry_get` (`relay.rs:633`) can only ever produce `ConnectionRefused`, and the relay handle is awaited *after* the callback (`relay.rs:621`), so the request's `unwrap()` masks the real error.

  ## Evidence

  8 concurrent test processes × 500 suite runs × `--test-threads=32` on `main` (macOS/arm64): **6 failures spread over 5 different tests** — `serve_addr_metrics` ×2 (`/metrics`), `serve_addr_enr_ext_ip`, `serve_addr_enr`, `serve_addr_multiaddrs` (`ConnectionRefused`), `run_bootnode` (`FailedToListen(… AddrInUse)`). The flake is in the shared fixture, not in `serve_addr_enr_ext_ip`. It also reproduces deterministically by holding `http_address` with a squatter socket for ~1 s.

  ## Production bugs found alongside it

  - **A monitoring bind failure is only `warn!`-ed** (`web.rs:180-187`), leaving the relay running with an unserved port. Charon terminates the relay on it (`cmd/relay/relay.go` → `serverErr`). Most frequent failure above.
  - **Leaked listener on failed startup**: `run_relay_p2p_node` spawns `enr_server` (`p2p.rs:80`) before parsing the monitoring address (`p2p.rs:97-100`), and returns past the running ENR server without cancelling.
  - **`FailedToBindHttpListener(String)`** (`error.rs:22`) stringifies the `io::Error`, discarding `io::ErrorKind` — callers are pushed into string matching.

iamquang95:
Fix https://github.com/NethermindEth/pluto/issues/590

Relay startup is split into bind then serve:

  - `bind_relay()` binds every listener (monitoring, ENR HTTP, libp2p) and waits for libp2p to report the addresses it bound, before anything is served. `BoundRelay::serve()` then runs them as `select!` arms, so exiting drops the listeners.
  - Bind failures are now typed errors that fail the relay instead of a warning.
  - Tests use port 0 everywhere and read back what was bound, so the race is unrepresentable. Retry/backoff polling is gone (`backon` dropped from cli dev-deps).

  Two follow-ons from kernel-assigned ports: external addrs are now derived from the bound addresses (`external_multiaddrs(cfg, listen_addrs)` + `Node::set_advertised_addrs`) rather than the configured ones, which would advertise an undialable port 0; and private-address filtering moved from ingest to render time, since startup waits on every listener and a loopback-only node must still start.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5181,7 +5181,6 @@ dependencies = [
 name = "pluto-cli"
 version = "1.7.1"
 dependencies = [
- "backon",
  "bytes",
  "chrono",
  "clap",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -69,14 +69,15 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "alloy"
-version = "1.8.3"
+version = "2.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "50ab0cd8afe573d1f7dc2353698a51b1f93aec362c8211e28cfd3948c6adba39"
+checksum = "754c72f6ae867220976be4693d870f8f9866437f37fdeb5640aa24991e5ff97f"
 dependencies = [
  "alloy-consensus",
  "alloy-contract",
  "alloy-core",
  "alloy-eips",
+ "alloy-ens",
  "alloy-genesis",
  "alloy-network",
  "alloy-provider",
@@ -103,9 +104,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-consensus"
-version = "1.8.3"
+version = "2.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f16daaf7e1f95f62c6c3bf8a3fc3d78b08ae9777810c0bb5e94966c7cd57ef0"
+checksum = "99bdf6a3e99089d97e137009956928b41d645035b220cb303eae88182882e9ca"
 dependencies = [
  "alloy-eips",
  "alloy-primitives",
@@ -130,9 +131,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-consensus-any"
-version = "1.8.3"
+version = "2.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "118998d9015332ab1b4720ae1f1e3009491966a0349938a1f43ff45a8a4c6299"
+checksum = "18d98872f189ea87063912de45e7e0d6ae44700a8836f8bfa79b3265f0c38bfe"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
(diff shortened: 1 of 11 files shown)


iamquang95:
@claude review this PR

emlautarom1 (changes requested):
As part of the refinement phase, we should try to get a proper solution in place, and retrying here seems like a hack. The original issue describes the problem as something that happens on test-code only code, in particular in CI.

Ideally, we should have something like `wiremock`, where we can start configure the relay server using a builder, then start it and let it live until the end of the scope (the test). The `with_relay_server` abstraction is introducing more complications that benefits at this point.

emlautarom1:
For example, tests should ideally look something like:


```rust
#[tokio::test]
async fn serve_addr_enr_ext_ip() {
    let relay = test_relay_server_with(|args| args.p2p.external_ip = Some("222.222.222.222".into())).await.unwrap();
    let addr = relay.http_addr.unwrap();

    let response = http_get(&format!("http://{addr}/enr")).await.unwrap();
    let enr = Record::try_from(response.text().await.unwrap().as_str()).unwrap();

    assert_eq!(enr.ip(), Some(Ipv4Addr::new(222, 222, 222, 222)));
}

#[tokio::test]
async fn run_bootnode_auto_p2p() {
    let missing_key = test_relay_server_with(|args| args.relay.auto_p2p_key = false).await;
    assert!(matches!(
        missing_key,
        Err(super::CliError::RelayP2PError(
            pluto_relay_server::RelayP2PError::FailedToLoadPrivateKey(..)
        ))
    ));

    let _relay = test_relay_server_with(|args| { }).await; // starts with an auto-generated key
}

#[tokio::test]
async fn taken_monitoring_port_fails_the_relay() {
    let taken = net::TcpListener::bind(ANY_ADDR).await.unwrap();
    let addr = taken.local_addr().unwrap().to_string();

    let err = test_relay_server(|args| args.debug_monitoring.monitor_addr = Some(addr))
        .await
        .expect_err("relay must not start while its monitoring port is taken");

    assert!(matches!(
        err,
        super::CliError::RelayP2PError(
            pluto_relay_server::RelayP2PError::FailedToBindMonitoringListener { .. }
        )
    ));
}
```

The tricky part is how we can retrieve the actually bounded addresses. I think this requires some refactoring to the internals so we can split binding from serving: I explored the work a bit in `fix/relay-bind-before-serve` if you want to take a look.

emlautarom1 (approved):
Test look a lot better without the retry dance. There are a few things to improve but overall LGTM, the PR solves the original problem. The PR body is stale so it should be updated with the final design.

PR #642, opened 2026-08-13
fix(cluster): hash unsigned v1.3/v1.4 operator signatures as raw bytes

### LINKED ISSUE #633: Cluster `definition_hash` mismatch for unsigned v1.3/v1.4 definitions
## Summary

[`hash_definition_v1x3or4`](https://github.com/NethermindEth/pluto/blob/462cdf67f4f99b124885ab07451695655cca4fd7/crates/cluster/src/ssz.rs#L326-L329) writes each operator's `config_signature`/`enr_signature` via `put_bytes_n(.., SSZ_LEN_K1_SIG)`, left-padding to 65 bytes. Charon's `hashDefinitionV1x3or4` uses raw [`hh.PutBytes`](https://github.com/ObolNetwork/charon/blob/749d2d7ab0b8ace34f2e686a5af5910c8adb4dc0/cluster/ssz.go#L243-L247) for exactly these two fields; only the later [`hashDefinitionV1x5to9` uses `putBytesN`](https://github.com/ObolNetwork/charon/blob/749d2d7ab0b8ace34f2e686a5af5910c8adb4dc0/cluster/ssz.go#L348-L355), which Pluto [matches correctly](https://github.com/NethermindEth/pluto/blob/462cdf67f4f99b124885ab07451695655cca4fd7/crates/cluster/src/ssz.rs#L439-L442). The [creator `config_signature` in the same function](https://github.com/NethermindEth/pluto/blob/462cdf67f4f99b124885ab07451695655cca4fd7/crates/cluster/src/ssz.rs#L350) already uses raw `put_bytes`, matching Go — only the two operator fields diverge.

For a fully-signed definition (65-byte signatures) the two encodings coincide, so the fixture suite passes. For an unsigned or partially-signed v1.3/v1.4 definition — accepted on both sides: ["Completely unsigned operators are also fine"](https://github.com/ObolNetwork/charon/blob/749d2d7ab0b8ace34f2e686a5af5910c8adb4dc0/cluster/definition.go#L244), and Pluto carries [the same comment](https://github.com/NethermindEth/pluto/blob/462cdf67f4f99b124885ab07451695655cca4fd7/crates/cluster/src/definition.rs#L567) plus `verify_signatures_*_unsigned_operators` tests — Go hashes zero bytes (one empty chunk) while Pluto hashes 65 zero bytes (three chunks). Different `definition_hash`, so `verify_hashes` rejects a Charon-produced v1.3/v1.4 artifact and vice versa.

## Also: `unreachable!` on a JSON-supplied version string

[`get_operator_eip712_type`](https://github.com/NethermindEth/pluto/blob/462cdf67f4f99b124885ab07451695655cca4fd7/crates/cluster/src/eip712sigs.rs#L119-L122) panics via `unreachable!("invalid eip712 signature version")` when `support_eip712_sigs(version)` is false. The version flows from definition JSON; the panic is unreachable today only because [`verify_signatures` pre-checks](https://github.com/NethermindEth/pluto/blob/462cdf67f4f99b124885ab07451695655cca4fd7/crates/cluster/src/definition.rs#L550) — an invariant enforced purely by caller discipline. Make it return a typed `UnsupportedVersion`-style error and propagate.

## Proposed change

- Switch the two operator-signature writes in `hash_definition_v1x3or4` to raw `put_bytes`; leave `v1x5to9` untouched.
- Regression test against a Charon-produced fixture: an unsigned (or partially-signed) v1.3 definition whose expected hashes are captured from Go — not from Pluto's current output. The golden-test harness from #603 is the natural home.
- Make `get_operator_eip712_type` fallible; add a test that an unsupported version yields an error, not a panic.

## Acceptance

An unsigned v1.3 definition produced by Charon passes `verify_hashes()` in Pluto; no `unreachable!` reachable from definition JSON.

iamquang95:
Fix #633 

  `hash_definition_v1x3or4` left-padded operator `config_signature`/`enr_signature`  to 65 bytes; charon (ssz.go:244-247) writes the raw bytes. Signed definitions coincide (already 65 bytes), but unsigned/partially-signed v1.3/v1.4 definitions hashed differently, so `verify_hashes()` rejected charon-produced artifacts.

  - Switch the two operator-signature writes to raw `put_bytes` (v1x5to9 and the creator field were already correct).
  - `get_operator_eip712_type` now returns `Err(UnsupportedVersion)` instead of `unreachable!()` on a version string coming from definition JSON.
  - Regression fixtures (unsigned + partially-signed v1.3) captured from charon v1.7.1's own `SetDefinitionHashes()`, not from pluto's output; verified the  test fails against the old code. All existing signed-path fixtures unchanged.

### DIFF
--- a/crates/cluster/src/ssz.rs
+++ b/crates/cluster/src/ssz.rs
@@ -323,10 +323,12 @@ pub(crate) fn hash_definition_v1x3or4<H: HashWalker>(
                 put_byte_list(hh, o.enr.as_bytes(), SSZ_MAX_ENR, "enr")?;
 
                 // Field (2) 'ConfigSignature' Bytes65
-                put_bytes_n(hh, &o.config_signature, SSZ_LEN_K1_SIG)?;
+                hh.put_bytes(&o.config_signature)
+                    .map_err(SSZError::<H>::HashWalkerError)?;
 
                 // Field (3) 'ENRSignature' Bytes65
-                put_bytes_n(hh, &o.enr_signature, SSZ_LEN_K1_SIG)?;
+                hh.put_bytes(&o.enr_signature)
+                    .map_err(SSZError::<H>::HashWalkerError)?;
             }
 
             hh.merkleize(op_sub_idx)
--- a/crates/cluster/src/definition.rs
+++ b/crates/cluster/src/definition.rs
@@ -556,7 +556,7 @@ impl Definition {
         }
 
         let operator_config_hash_digest = digest_eip712(
-            &get_operator_eip712_type(self.version.as_str()),
+            &get_operator_eip712_type(self.version.as_str())?,
             self,
             &Operator::default(),
         )?;
@@ -1906,6 +1906,18 @@ mod tests {
         let definition = serde_json::from_str::<Definition>(json_str).unwrap();
 
         assert!(definition.verify_hashes().is_ok());
+    }
+
+    #[test]
+    fn cluster_definition_v1_3_0_unsigned_operators() {
+        for fixture in [
+            include_str!("testdata/cluster_definition_v1_3_0_unsigned.json"),
+            include_str!("testdata/cluster_definition_v1_3_0_partial_sigs.json"),
+        ] {
+            let definition = serde_json::from_str::<Definition>(fixture).unwrap();
+
+            assert!(definition.verify_hashes().is_ok());
+        }
     }
 
     #[test]
--- a/crates/cluster/src/eip712sigs.rs
+++ b/crates/cluster/src/eip712sigs.rs
@@ -26,6 +26,10 @@ pub enum EIP712Error {
     /// Failed to sign EIP-712.
     #[error("Failed to sign EIP-712: {0}")]
     FailedToSign(pluto_k1util::K1UtilError),
+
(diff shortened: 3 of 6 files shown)


crates/cluster/src/ssz.rs
```diff
@@ -323,10 +323,12 @@ pub(crate) fn hash_definition_v1x3or4<H: HashWalker>(
                 put_byte_list(hh, o.enr.as_bytes(), SSZ_MAX_ENR, "enr")?;
 
                 // Field (2) 'ConfigSignature' Bytes65
-                put_bytes_n(hh, &o.config_signature, SSZ_LEN_K1_SIG)?;
+                hh.put_bytes(&o.config_signature)
```
iamquang95, line 326: https://github.com/ObolNetwork/charon/blob/v1.7.1/cluster/ssz.go#L243-L247

```
				// Field (2) 'ConfigSignature' Bytes65
				hh.PutBytes(o.ConfigSignature)

				// Field (3) 'ENRSignature' Bytes65
				hh.PutBytes(o.ENRSignature)
```

emlautarom1 (approved):
LGTM

